PR #5, opened 2022-06-25
Disable telemetry for sumneko lua lsp

skovati:
Based on existing recommended config in [lsp-config](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#sumneko_lua), this disables the telemetry for sumneko lua lsp.


tjdevries:
Shouldn't they get merged or no?

skovati:
@tjdevries sorry, what do you mean? I know kickstart.nvim is supposed to act as the reference configuration for lspconfig, but this seemed like something that should be "backported"

PR #12, opened 2022-08-02
feat: enable `auto_install` for `nvim-treesitter`

numToStr:
I would say, why not? :)


Jarmos-san:
I don't know if an arbitrary parser should  be installed automatically without my manual intervention. I often use Neovim as an alternative to manpages or `cat` and if for whatever reason, if Neovin decided to download a parser for it, I would feel iffy about it. 

`ensure_installed` is more than enough I think.

numToStr:
@Jarmos-san IMO kickstart.nvim meant to be a starting guide with nice and sane defaults. You are free to change whatever you want and I highly doubt that people are using kickstart without any modifications.

tjdevries:
For now, let's just leave the way it is to not auto install everything -- just in case some parsers  currently have some difficulties or anything. Later as things settle even more, we can move to just install everything.

PR #16, opened 2022-08-24
Don't pass args to vim.lsp.buf.format/formatting in "Format"

Vzaa:
Call `vim.lsp.buf.format/formatting` without any arguments in `Format`.

With rust-analyzer I was getting the error below in 0.7.x with `vim.lsp.buf.formatting`. Incorrect parameters were being used as `FormattingOptions` before.

```
rust_analyzer: -32602: Failed to deserialize textDocument/formatting: data did not match any variant of untagged enum FormattingProperty; {"textDocument":{"uri":"file:///tmp/hello/src/main.rs"},"options":{"bang":false,"fargs":[],"args":"","line1":1,"line2":1,"reg":"","mods":"","tabSize":4,"insertSpaces":true,"range":0,"count":-1}}
```

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -289,7 +289,13 @@ local on_attach = function(_, bufnr)
   end, '[W]orkspace [L]ist Folders')
 
   -- Create a command `:Format` local to the LSP buffer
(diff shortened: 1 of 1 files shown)


tjdevries:
Thanks!

PR #18, opened 2022-09-01
Added cpp to treesitter and ordered TS imports

Ttibsi:
We're adding the clangd language server on line 299 so it seemed sensible to add cpp to the treesitter languages. (I could also have added C as well/instead - please advise if you'd prefer C there instead or added as well. I'm guessing C would make more sense as that's the main language nvim is written in?)

I've also alphabetised that list of treesitter languages while I was at it just to make it look better.


tjdevries:
yeah, go ahead and add C as well, seems like a fine idea.

Ttibsi:
@tjdevries Updated. :)

PR #20, opened 2022-09-21
Add optional packer plugins hook

elijahmanor:
As an experiment, I've gone "neovim config bankruptcy" and have been using `kickstart.nvim` for the last two weeks. I added a few custom plugins to packer and several custom options and keymaps.

Then I noticed how active this repo is (which is awesome) and I wanted to upgrade my local version, but my additions were intermingled in the base `init.lua` file. 

So, I pulled out my custom options and keymaps into a `nvim/after/plugin/defaults.lua` file (which worked out great), but the issue I was encountering was figuring out how to pull out the definition of my custom plugins, hence this PR. 

With the two new lines included in this PR it enables the user to safely reset their version of `init.lua` and let any of their customizations be in a `nvim/after/plugin/defaults.lua` file and optionally a `nvim/lua/custom/plugins.lua` module.

The optional plugins module could look like...

```lua
-- lua/custom/plugins.lua
local plugins = {}

function plugins.setup(use)
  use { "nvim-telescope/telescope-file-browser.nvim" }

  use({
    "folke/which-key.nvim",
    config = function()
      require("which-key").setup({})
    end,
  })
end

return plugins
```

> NOTE: As an aside, it is great that you all have added the `desc` to the mappings you've included, which works great with `which-key.nvim` making it easier for a new user to get their bearings on the possible unfamiliar keymaps.

You all have done a great job of keeping things simple, fresh, and updated as neovim and plugin APIs shift. For that reason, it seems like a nice feature to let the user update their version (using something like `wget https://github.com/nvim-lua/kickstart.nvim/blob/master/init.lua -O ~/.config/nvim/init.lua`) without losing their custom configurations.

Thanks for your consideration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,10 +12,42 @@ Kickstart targets *only* the latest stable neovim release (0.7) and the nightly.
 This repo is meant to be used as a starting point for a user's own configuration; remove the things you don't use and add what you miss. This configuration serves as the reference configuration for the [lspconfig wiki](https://github.com/neovim/nvim-lspconfig/wiki).
 
 ### Installation
+
 * Backup your previous configuration
 * Copy and paste the kickstart.nvim `init.lua` into `$HOME/.config/nvim/init.lua`
 * start neovim (`nvim`) and run `:PackerInstall`, ignore any error message about missing plugins, `:PackerInstall` will fix that shortly.
 * restart neovim
+
+### Configuration
+
+You could directly modify the `init.lua` file with your personal customizations. This option is the most straightforward, but if you update your config from this repo, you may need to reapply your changes.
+
+An alternative approach is to create a separate `custom.plugins` module to register your own plugins. In addition, you can handle further customizations in a `after/plugin/defaults.lua` file. See the following examples for more information. Leveraging these files should make upgrading to a newer version of this repo easier. 
+
+#### Example `plugins.lua`
+
+The following is an example of a `plugins.lua` file (located at `$HOME/.config/nvim/lua/custom/plugins.lua`) where you can register you own plugins. 
+
+```lua
+return function(use)
+  use({
+    "folke/which-key.nvim",
+      config = function()
+        require("which-key").setup({})
+      end
+  })
+end
+```
+
+#### Example `defaults.lua`
+
+The following is an example `defaults.lua` file (localed at `$HOME/.config/nvim/after/plugin/defaults.lua`) where you can define your own options, keymaps, autogroups, and more.
+
+```lua
+vim.opt.relativenumber = true
+
+vim.keymap.set('n', '<leader>sr', require('telescope.builtin').resume, { desc = '[S]earch [R]esume' })
+```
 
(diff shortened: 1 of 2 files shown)


tjdevries:
Ah, this is a good idea! Could you add an example in the README about how to use, then I'll merge. This is a great idea

elijahmanor:
Thanks @tjdevries I'll work on that. Good idea to include a sample in the README

tjdevries:
maybe also have custom.plugins just return a function, nothing else. So it's just basically a `return function(use) ... end` as a module.

elijahmanor:
Okay @tjdevries, I added some info to the README and simplified the module. Let me know if you'd like any more tweaks. Thank you for your feedback.

README.md
```diff
@@ -12,11 +12,43 @@ Kickstart targets *only* the latest stable neovim release (0.7) and the nightly.
...
 
+### Configuration
+
+You could directly modify the `init.lua` file with your personal customizations. This option is the most straightforward, but if you update your config from this repo, you may need to reapply your changes.
+
+An alternative approach is to create a separate `custom.plugins` module to register your own plugins. In addition, you can handle further customizations in a `after/plugin/defaults.lua` file. See the following examples for more information. Leveraging these files should make upgrading to a newer version of this repo easier. 
+
+#### Example `plugins.lua`
+
+The following is an example of a `plugins.lua` file (located at `$HOME/.config/nvim/lua/custom/plugins.lua`) where you can register you own plugins.
```
tjdevries, line 29: "you own plugins" -> "your own plugins"

README.md
```diff
@@ -12,11 +12,43 @@ Kickstart targets *only* the latest stable neovim release (0.7) and the nightly.
...
 * Backup your previous configuration
 * Copy and paste the kickstart.nvim `init.lua` into `$HOME/.config/nvim/init.lua`
 * start neovim (`nvim`) and run `:PackerInstall`, ignore any error message about missing plugins, `:PackerInstall` will fix that shortly.
 * restart neovim
 
+### Configuration
+
+You could directly modify the `init.lua` file with your personal customizations. This option is the most straightforward, but if you update your config from this repo, you may need to reapply your changes.
+
+An alternative approach is to create a separate `custom.plugins` module to register your own plugins. In addition, you can handle further customizations in a `after/plugin/defaults.lua` file. See the following examples for more information. Leveraging these files should make upgrading to a newer version of this repo easier.
```
tjdevries, line 25: can you add a link to the help for this (since it can be any name, not just `after/plugin/defaults.lua`

README.md
elijahmanor: done

README.md
elijahmanor: Okay, I tried to tweak the language to accommodate that. Good point! I'm not sure where I should link, so I added `:help load-plugins`. Is that what you meant?

tjdevries:
Perfect! Thanks :)

PR #24, opened 2022-10-19
fix: syntax highlights tree-sitter

zirbest:
- `navarasu/onedark.nvim` support both API TS* and the new one @*
- `mjlbach/onedark.nvim` is archived and not support the new syntax


tjdevries:
Ah, good point. Thanks

PR #30, opened 2022-11-09
Update README.md

craigmac:
Use links to tagged releases rather than hardcoding 0.x and having to change it every few months. Fix markdownlint errors and some spelling, like using the official 'Neovim', not 'neovim'.


tjdevries:
Thanks!

PR #33, opened 2022-11-17
lsp.update_ is deprecated

nikolasbertrand:
the lsp.update_ capabilities is deprecated , default_capabilities should be used instead.


nikolasbertrand:
Maybe had to before doing a pull request, but the error message   was already in an earlier pull request. But nothing changed since than. Unless you want to keep it in there as teachable moment?

tjdevries:
Thanks :)

PR #35, opened 2022-11-22
Update the amount of SLOC in README

fgrehm:
Looks like the project has grown a tad bit 😅 

Screenshot_20221122-192521~2


tjdevries:
haha thanks :) Some of the lines are primarily just due to adding some docs and a few other things, so I'm not worried about it. Thanks for updating though.

PR #36, opened 2022-11-23
Enable parameter text objects

c4rlo:
This seems like a natural addition. In classic Vim, I used to use https://github.com/b4winckler/vim-angry for this.


tjdevries:
makes sense to me! thanks :)

PR #54, opened 2022-12-18
Fix repeated initialization of lsp

j2501987804:
sumneko_lua will repeat setup, which may affect performance


tjdevries:
Oh, I like this idea. I might switch slightly the format in the next day or so. I will tag you to see if you like the change.

tjdevries:
Fixed now in #71. Thanks for the tip, made me dig into some new updates from mason and found a much more elegant way of config for the lsp's

Issue #45, opened 2022-12-17
On new installation ask for working environment on every startup


santisanchez:
To reproduce, after automatic install of plugins and TS i save the init.lua or just open the init.lua and ask everytime if i want to 

"Do you need to configure your work environment as `luassert`?:"

I don't know if this is intended or needed for every change, but found in this issue on nvim-lspconfig that disabling checkThirdParty = false in the settings.workspace stops asking

Link to the issue that solved my problem:
https://github.com/neovim/nvim-lspconfig/issues/1700

PR #57, opened 2022-12-18
fix: don't check third party for Lua LSP

ngryman:
Every time I open a Lua file to configure `nvim`, I have this annoying prompt:

```
Do you need to configure your work environment as `luassert`?

Request Actions:
1. Apply and modify settings
2. Apply but do not modify settings
3. Don't show again
Type number and <Enter> or click with the mouse (q or empty cancels):
```

I'm not familiar with `sumneko_lua` but [a similar issue was reported here](https://github.com/neovim/nvim-lspconfig/issues/1700) and [this fix was suggested](https://github.com/neovim/nvim-lspconfig/issues/1700#issuecomment-1033127328).

It works and makes the prompt disappear. I'm proposing this fix to you in case you're interested.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -381,7 +381,7 @@ require('lspconfig').sumneko_lua.setup {
       },
       workspace = {
         library = vim.api.nvim_get_runtime_file('', true),
-        checkcheckThirdParty = false,
+        checkThirdParty = false,
       },
       -- Do not send telemetry data containing a randomized but unique identifier
       telemetry = { enable = false },
--- a/init.lua
+++ b/init.lua
@@ -379,7 +379,10 @@ require('lspconfig').sumneko_lua.setup {
       diagnostics = {
         globals = { 'vim' },
       },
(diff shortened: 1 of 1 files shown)


init.lua
```diff
@@ -379,7 +379,10 @@ require('lspconfig').sumneko_lua.setup {
       diagnostics = {
         globals = { 'vim' },
       },
-      workspace = { library = vim.api.nvim_get_runtime_file('', true) },
+      workspace = {
+        library = vim.api.nvim_get_runtime_file('', true),
+        checkcheckThirdParty = false,
```
seantwie03, line 384: Looks like there is a typo

```suggestion
        checkThirdParty = false,
```

errantepiphany: Hah, I was just creating a PR for `checkThirdParty = false`, but then saw this PR. @ngryman Can you please fix the typo per @seantwie03 so this PR can get merged? I'm sure this fix will make a lot of people happier - thanks for submitting it! :)

tjdevries:
Closes #60
Closes #45

PR #62, opened 2022-12-19
Do not override treesitter lua highlighting with sumneko lua highlighting

errantepiphany:
I noticed that when opening a lua file, the highlighting initially looks really nice. (Thanks treesitter!) But then once sumneko_lua is done initializing, the highlighting changes slightly (specifically, parent fields turn back to white). This is because sumneko overrides treesitter's highlighting with its own. So this PR disables the sumneko highlighting so we can keep the pretty treesitter one.

This is the field:
https://github.com/sumneko/lua-language-server/wiki/Settings#semanticenable

Signed-off-by: David Ward <dward@redhat.com>

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -382,6 +382,8 @@ require('lspconfig').sumneko_lua.setup {
       workspace = { library = vim.api.nvim_get_runtime_file('', true) },
       -- Do not send telemetry data containing a randomized but unique identifier
       telemetry = { enable = false },
+      -- Do not override treesitter lua highlighting with sumneko lua highlighting
+      semantic = { enable = false },
     },
   },
 }


seantwie03:
I'm still a bit of a noob when it comes to LSP, with this configuration could there ever be a conflict between treesitter highlighting and LSP error reporting? For example, if treesitter and sumneko parse a file differently and disagree on what a symbol is, could it lead to problems?

errantepiphany:
According to the sumneko documentation, the entire [semantic](https://github.com/sumneko/lua-language-server/wiki/Settings#semantic) group of configuration properties only has to do with highlighting. I guess it could be possible that treesitter's grammar could mistakenly consider a token to be something different than what sumneko thinks it is, however the only repercussion of that would be a mis-coloring. But the LSP behavior would not be affected. But again, this is according to the docs. If the docs are wrong or misleading, then I could be wrong.

seantwie03:
Thanks for the links to the docs. I agree with you. Any error scenario like this is probably going to be very infrequent.

tjdevries:
I wonder if it's just because this colorscheme hasn't been updated yet to have the semantic highlights be good colors? I'll have to follow up on that to find that out.

errantepiphany:
Hi @tjdevries, thanks for looking at this. I will describe exact examples of the behavior I'm seeing with two different color schemes, the moment right after **sumneko_lua** finishes initializing:

`navarasu/onedark.nvim`
- semantic enabled: the `use` keyword changes from blue to red (bad), but the `vim.fn` field stays blue (good)
- semantic disabled: the `use` keyword stays blue (good), and the `vim.fn` field stays blue (good)

`Mofiqul/dracula.nvim`
- semantic enabled: the `use` keyword changes from green to orange (bad), and the `vim.fn` field changes from orange to white (bad)
- semantic disabled: the `use` keyword stays green (good), and the `vim.fn` field stays orange (good)

There could be other color issues, but I just wanted to give a couple examples that stood out to me. This might be a problem in sunneko_lua, however doesn't that become moot if we can just stop it from clobbering tree-sitter's colors with this PR?

tjdevries:
`use` is not a keyword -- it's a variable (actually, it's a parameter passed to the function). My guess is it's highlighting parameters as different things than regular variables. I think that's personal preference on what you'd want to do.

It'd be helpful to see screenshots, but I'm unlikely to turn it off by default I think. I'll try and do some testing this week though and maybe I'll end up agreeing with you haha

errantepiphany:
Oops, yeah, you're right: `use` is a parameter. But `vim.fn` is a field, and _both_ are getting clobbered. So it's not just a parameter-thing. Thanks again; I appreciate your time on this. I know it's small, but it's annoying to see colors unintentionally change (for the worse) a couple seconds after you open the file.

CharlesChiuGit:
@errantepiphany Maybe u can try `:TSHighlightCapturesUnderCursor` to see which hilight group is taking charge, a command from [ts-playground](https://github.com/nvim-treesitter/playground)

craigmac:
If you've got a [very recent](https://github.com/neovim/neovim/commit/ef91146efcece1b6d97152251e7137d301146189) Neovim you can also now do: `:Inspect` and `:Inspect!` (for pretty-print)

tjdevries:
Yeah, let's close for now (and fix any upstream issues if they exist). Thanks for doing all this tracking down and no worries at all :) Thanks for the detailed write ups!

PR #64, opened 2022-12-19
Fix error on Windows when space in install path

### LINKED ISSUE #55: Troubles installing it on Windows (E5113 & E919)
I get this error while trying to install it on windows, what is the issue?

_E5113: Error while calling lua chunk: C:\Users\MIO PC\AppData\Local\nvim\init.lua:7: Vim(packadd):E919: Directory not fo                                                              und in 'packpath': "pack/*/opt/packer.nvim"
stack traceback:
        [C]: in function 'cmd'
        C:\Users\MIO PC\AppData\Local\nvim\init.lua:7: in main chunk_

debashisbiswas:
Fixes #55. When there is a space in the install path, Packer is not cloned properly on startup. This change resolves the issue. Tested on Windows 11 with Neovim v0.8.1

`--depth 1` is borrowed from the original Packer documentation at https://github.com/wbthomason/packer.nvim#bootstrapping. This option only clones the most recent commit, which is all that is necessary for the initial install.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -3,7 +3,7 @@ local install_path = vim.fn.stdpath 'data' .. '/site/pack/packer/start/packer.nv
 local is_bootstrap = false
 if vim.fn.empty(vim.fn.glob(install_path)) > 0 then
   is_bootstrap = true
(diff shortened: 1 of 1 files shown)


tjdevries:
Nice, good catch! Thanks :)

PR #65, opened 2022-12-19
Update intstructions for existing nvim setups

matthiasdebernardini:
If you already had a nvim setup, then this makes it so that the kickstarter can work without a problem. Having that file exist and loaded with plugins gives the script many issues.

Also better to have it be a shell command rather than have the script delete a bunch of files on your computer.


tjdevries:
Can you make a FAQ section at the bottom and put this in there with some question like "What should I do if I alrady have an existing nvim setup" or similar?

matthiasdebernardini:
should probably squash these two commits into one? @tjdevries

tjdevries:
Thanks <3

PR #67, opened 2022-12-19
Ubuntu Dockerfile with working LSP builds

OhMyMndy:
Hi!

Thanks for the wonderful work, all!

I created a _Dockerfile_ for _Ubuntu_ to show how you can install _Neovim_ + all the dependencies necessary to run Packer + Mason successfully with k_ickstart.nvim_ 😊

### DIFF
--- a/Dockerfile.ubuntu
+++ b/Dockerfile.ubuntu
@@ -1,0 +1,21 @@
+FROM ubuntu:22.04
+
+ARG VERSION=stable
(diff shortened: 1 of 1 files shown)


tjdevries:
Thanks for sending this, but in the end I think i'm just going to remove the dockerfile -- I don't think it's incredibly helpful and ends up confusing people more often than it was worth. Sorry for the delay!

PR #71, opened 2022-12-20
move server config to easy to extend style

tjdevries:
Alternative to #54 

@j2501987804 what do you think of this style instead?

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -365,26 +365,30 @@ local servers = {
 
 -- Setup neovim lua configuration
 require('neodev').setup()
+--
+-- nvim-cmp supports additional completion capabilities, so broadcast that to servers
+local capabilities = vim.lsp.protocol.make_client_capabilities()
+capabilities = require('cmp_nvim_lsp').default_capabilities(capabilities)
 
 -- Setup mason so it can manage external tooling
 require('mason').setup()
 
 -- Ensure the servers above are installed
-require('mason-lspconfig').setup {
+local mason_lspconfig = require('mason-lspconfig')
+
+mason_lspconfig.setup {
   ensure_installed = vim.tbl_keys(servers),
 }
 
--- nvim-cmp supports additional completion capabilities, so broadcast that to servers
-local capabilities = vim.lsp.protocol.make_client_capabilities()
-capabilities = require('cmp_nvim_lsp').default_capabilities(capabilities)
-
-for lsp, settings in ipairs(servers) do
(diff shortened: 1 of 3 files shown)


craigmac:
Easier to follow, I'd say, and easier to extract should you not want to set them up via mason lsp-config.

init.lua
```diff
@@ -332,63 +332,57 @@ local on_attach = function(_, bufnr)
...
+  ensure_installed = vim.tbl_keys(servers),
 }
 
--- nvim-cmp supports additional completion capabilities
+-- nvim-cmp supports additional completion capabilities, so broadcast that to servers
 local capabilities = vim.lsp.protocol.make_client_capabilities()
 capabilities = require('cmp_nvim_lsp').default_capabilities(capabilities)
 
-for _, lsp in ipairs(servers) do
+for lsp, settings in ipairs(servers) do
```
MaxVerevkin, line 375: This should be `pairs`, I think.

tjdevries:
@craigmac @MaxVerevkin thoughts on new style? found a new setup that's pretty nice for mason-lspconfig that I wasn't aware of from #74

MaxVerevkin:
I like it. It also allows setting up new servers without editing the config (i.e. I can install a server and it will just work).

craigmac:
Ditto.

zdzolton:
(Apologies for commenting on a closed PR; I will gladly move elsewhere!)

Does this change allow for users to extend the language server list without directly modifying the init.lua file? I've been looking at using Kickstart by symlinking into my config directory [as mentioned here](https://github.com/nvim-lua/kickstart.nvim/#configuration) but haven't found a way to do what I want... Any ideas?

MaxVerevkin:
> extend the language server list without directly modifying the init.lua file?

Yes, just install them using `:Mason`.

zdzolton:
>Yes, just install them using :Mason.

Got it... Thank you, @MaxVerevkin!

I was worried that servers not added to the the local `servers` variable would miss out the completion capabilities configured here: 
https://github.com/nvim-lua/kickstart.nvim/blob/521940693e10eee57f4e67dce960d2229c1441d3/init.lua#L357-L379

tjdevries:
Nope, pretty sure they get picked up. But you should read Mason documentation to understand that better (and I should probably too haha)

j2501987804:
image
It seems that not all languages are settings, such as gopls, which provides {cmd,filetypes,filetypes} attributes

PR #74, opened 2022-12-20
feat: More user friendly language server installation

aidenlangley:
This change will automatically install language servers for users, and set up a default handler for them. Thought it was a little more user friendly & robust than the current solution.


tjdevries:
what do you think of this: https://github.com/nvim-lua/kickstart.nvim/pull/71

aidenlangley:
Eh, have just noticed I need to put the `ensure_installed` list back. `automatic_install` isn't what it seems.

tjdevries:
Ended up going with the style in #71. Thanks for the suggestion, didn't know about the setup_handlers stuff.

PR #77, opened 2022-12-20
Use the os clipboard

godalming123:
I feel like this is just a senseable default.


tjdevries:
This requires additional things installed -- I think it's better if people just do this if they want to (I know a lot of people do not like this setting). So I think I'll close for now.

(even though this is the setting I have)

PR #86, opened 2022-12-22
Link to the video introducing Kickstart.nvim

vipulmathur:
The video by @tjdevries is a good kickstart for folks coming to Kickstart.nvim.


tjdevries:
haha that's nice of you :) I will add this in the updated version in #178 (and then make a new video with even more updates... :) )

PR #101, opened 2022-12-28
chore(lsp): use builtin keymap

zirbest:
use builtin keymap

```
- gd      <-> c-]
- :Format <-> gq
 ```
more info
https://neovim.io/doc/user/lsp.html#lsp-config

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -306,7 +306,6 @@ local on_attach = function(_, bufnr)
   nmap('<leader>rn', vim.lsp.buf.rename, '[R]e[n]ame')
   nmap('<leader>ca', vim.lsp.buf.code_action, '[C]ode [A]ction')
 
-  nmap('gd', vim.lsp.buf.definition, '[G]oto [D]efinition')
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by tjdevries** [comment]
```
gd			Goto local Declaration.  When the cursor is on a local
			variable, this command will jump to its declaration.
			First Vim searches for the start of the current
			function, just like "[[".  If it is not found the
			search stops in line 1.  If it is found, Vim goes back
			until a blank line is found.  From this position Vim
			searches for the keyword under the cursor, like with
			"*", but lines that look like a comment are ignored
			(see 'comments' option).
			Note that this is not guaranteed to work, Vim does not
			really check the syntax, it only searches for a match
			with the keyword.  If included files also need to be
			searched use the commands listed in |include-search|.
			After this command |n| searches forward for the next
			match (not backward).
```

I like overriding `gd` for this case. I think it is idiomatic (and reasonable for new users as well).

Thanks for the suggestion though.

Issue #104, opened 2022-12-29
Cannot set lsp servers root_dir with root_pattern


guigui64:
**Describe the bug**
I am trying to override the root_dir of one of my LSP servers by using root_pattern but get the following error:
```
Error executing vim.schedule lua callback: /usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:289: Cannot serialise function: type not supported
stack traceback:
        [C]: in function 'encode'
        /usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:289: in function 'notify'
        /usr/share/nvim/runtime/lua/vim/lsp.lua:1326: in function ''
        vim/_editor.lua: in function <vim/_editor.lua:0>
```

**To Reproduce**
```lua
local servers = {
  tsserver = {
    root_dir = require("lspconfig").util.root_pattern("package.json")
  },
  sumneko_lua = {
    Lua = {
      workspace = { checkThirdParty = false },
      telemetry = { enable = false },
    },
  },
}
```

**Desktop:**
- OS: Ubuntu 22.04
- Terminal: Alacritty

**Neovim Version**
```
NVIM v0.8.1
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by runner@fv-az178-366

Features: +acl +iconv +tui
```

guigui64:
Nevermind I figured it out. `root_dir` needs to be given to the lspconfig setup method as a parameter and not in the settings table.
To keep my `servers` table as it is, I modified the `setup_handlers` part like so:

```lua
mason_lspconfig.setup_handlers {
  function(server_name)
    local root_dir = servers[server_name]['root_dir'] -- get root_dir from servers table
    servers[server_name]['root_dir'] = nil -- remove root_dir from table
    require('lspconfig')[server_name].setup {
      capabilities = capabilities,
      on_attach = on_attach,
      settings = servers[server_name], -- settings will not contain root_dir
      root_dir = root_dir, -- use root_dir on its own
    }
  end,
}
```

PR #107, opened 2022-12-30
Extract root_dir from server's settings table

guigui64:
See #104 for details

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -370,10 +370,14 @@ mason_lspconfig.setup {
 
 mason_lspconfig.setup_handlers {
   function(server_name)
+    -- In case root_dir is set in the servers table, we need to extract it before calling setup
+    local root_dir = servers[server_name]['root_dir']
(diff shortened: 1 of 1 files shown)


tjdevries:
You can customize individual servers using mason lspconfig like this:

```lua
    require("mason").setup()
    require("mason-lspconfig").setup()

    require("mason-lspconfig").setup_handlers {
        -- The first entry (without a key) will be the default handler
        -- and will be called for each installed server that doesn't have
        -- a dedicated handler.
        function (server_name) -- default handler (optional)
            require("lspconfig")[server_name].setup {}
        end,
        -- Next, you can provide a dedicated handler for specific servers.
        -- For example, a handler override for the `rust_analyzer`:
        ["rust_analyzer"] = function ()
            require("rust-tools").setup {}
        end
    }
```

This would be the recommended way you could override things for tsserver individually

tjdevries:
Thanks for the suggestion tho!

PR #110, opened 2023-01-01
added vim to tree-sitter's ensured installed list

nesper8:
please refer to https://github.com/nvim-treesitter/nvim-treesitter/issues/3092

> Since 0.8.0, Neovim bundles parsers and queries for c, lua, vim, and help. If you use nvim-treesitter, you must make sure these parsers are installed via nvim-treesitter so that both parser and queries (which are always installed) are taken from nvim-treesitter. (It's important for this that your nvim-treesitter plugin directory comes before both /usr/local/share/nvim/* and /usr/lib/nvim in your runtimepath.)


tjdevries:
Thanks!

PR #115, opened 2023-01-02
perf: replace Packer.nvim with Lazy.nvim

zirbest:


init.lua
```diff
@@ -1,95 +1,69 @@
...
-  }
+    event = 'InsertEnter',
+    dependencies = { 'hrsh7th/cmp-nvim-lsp', 'L3MON4D3/LuaSnip', 'saadparwaiz1/cmp_luasnip' },
+  },
 
-  use { -- Highlight, edit, and navigate code
+  { -- Highlight, edit, and navigate code
     'nvim-treesitter/nvim-treesitter',
-    run = function()
+    config = function()
```
mathjiajia, line 43: build?

zirbest: yep

ellisonleao:
for this case it's just better to use https://github.com/folke/LazyVim

szechp:
> for this case it's just better to use https://github.com/folke/LazyVim

Hm, that's what he did, he replaced packer with lazy. Did you mean using the preconfigured lazyvim? That's a good idea, but the preconfigured lazyvim has a quite different keymap Konfiguration. Since I'm used to kick-starts keymappings, I think it would be really useful to port kickstart from packer to lazyvim.

mortezadadgar:
But why?

ubmit:
That would be a good enhancement

szechp:
> But why?

Lazyvim is great, but it's a little bit more bloated than kickstart. Also if you are used to the few keymappings kickstart has, lazyvim is quite different. That's why it would be cool to have kickstart with lazy.nvim as a plugin manager.

tjdevries:
Doing this in #178 thanks for the PR though! :)

Issue #112, opened 2023-01-02
Having trouble installing on Windows 11


bfnoling:
I followed the instructions on the README as best I could for windows. I am running this on a Development VM where the user account is named "User"

1. Created 

**To Reproduce**
1. Installed the Windows prerequisites by installing the "Desktop development with C++" workload in Visual Studio
2. Created  "C:\Users\User\.config\nvim"
3. Pasted the contents of init.lua from the repo to C:\Users\User\.config\nvim\init.lua"
4. Updated the line about fuzzy finder with the Windows specific line in the repo
5. Launched nvim.exe and nothing has changed

I grabbed Neovim from Winget and am using v0.8.0-1210-gd367ed9b2

Windows Version 21H2

I am using neovim from Windows Terminal and every time I start nvim.exe, a line like the one below is added to C:\Users\User\AppData\Local\nvim-data\log

"ERR 2023-01-01T20:35:11.773 nvim.2828.0 terminfo_start:359: uv_tty_set_mode failed: invalid argument"

I really apologize if I missed something in the documentation. I got this running in Linux with no problem and appreciate your help.

seantwie03:
Hello,

Let's start by correcting the location of your init.lua file. It should be in `~/AppData/Local/nvim/init.lua`

Fore more information please see [:h config](https://neovim.io/doc/user/starting.html#config)

bfnoling:
That did the trick. Thank you!

PR #121, opened 2023-01-04
Add windows init lua location

seantwie03:
Specify location of init.lua on Windows.

https://github.com/nvim-lua/kickstart.nvim/issues/112


tjdevries:
Thanks, that's a great call out :) I appreciate that.

PR #124, opened 2023-01-09
Enabling Emmet Support

nesper8:
This PR enables Emmet support for website building
This PR will also only enable emmet for html, css, sass, scss, xml and jsx files

What is emmet?
check out https://emmet.io/

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -127,7 +127,7 @@ vim.cmd [[colorscheme onedark]]
 
 -- Enable Emmet only for html and css
 vim.g.user_emmet_install_global = 0
-vim.cmd('autocmd FileType html,css EmmetInstall')
+vim.cmd('autocmd FileType html,css,scss,sass,xml,jsx EmmetInstall')
 
 -- Set completeopt to have a better completion experience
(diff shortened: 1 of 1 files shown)


nesper8:
The last commit changes the Default Emmet leader key from <C-Y> to <Tab> to better imitate VSCode although you still need to do <Tab>, to expand the Emmet abbreviations

ubmit:
What do you think about using the LSP to handle it? I had been using `emmet_ls` and didn't face issues with it: https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#emmet_ls

We have only one language server setup by default though, which is the Sumneko and maybe we should keep as it's. So people can add emmet support or any other feature when they need them.

nesper8:
Wait so emmet_ls from Mason also serve the same purpose?
I haven't checked it out though

ubmit:
> Wait so emmet_ls from Mason also serve the same purpose?
> I haven't checked it out though 

I guess it does not sure sorry

nesper8:
just tried out emmet-ls 
feels like it is more easier than mattn's implementation 
so ill close this PR

PR #130, opened 2023-01-12
Update readme installation for mac

xuwhite:
added intallation path for mac


tjdevries:
Thanks, will update in #178

PR #131, opened 2023-01-13
Add hint to ripgrep dependency to readme

Dennis-Rall:
Thanks for the amazing project to get the nvim configuration going. I had a miner problem with telescope. You need to install a separate dependency to use the grep functions. I added a note to the installation details in the readme.


tjdevries:
Good point, I'll add this slightly modified in #178

PR #138, opened 2023-01-18
Configure packer_compiled.lua to be stored on XDG_DATA_DIRS

robsonpeixoto:
The `packer_compiled.lua` use a lot of values that only works on the host
where the file was compiled.

Put the compiled file on XDG_DATA_DIRS make easier use ~/.config/nvim to both
on Linux and Macos.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -12,15 +12,15 @@ local status_ok, packer = pcall(require, "packer")
 if not status_ok then
   return
 end
-local status_util_ok, util = pcall(require, "packer.util")
+local status_util_ok, packer_util = pcall(require, "packer.util")
 if not status_util_ok then
   return
 end
 
 -- Have packer use a popup window
 packer.init({
-  package_root = util.join_paths(vim.fn.stdpath('data'), 'site', 'pack'),
-  compile_path = util.join_paths(vim.fn.stdpath('data'), 'site', 'plugin', 'packer_compiled.lua'),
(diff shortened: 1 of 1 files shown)


init.lua
```diff
@@ -7,7 +7,23 @@ if vim.fn.empty(vim.fn.glob(install_path)) > 0 then
   vim.cmd [[packadd packer.nvim]]
 end
 
-require('packer').startup(function(use)
+-- Use a protected call so we don't error out on first use
+local status_ok, packer = pcall(require, "packer")
+if not status_ok then
+  return
+end
+local status_util_ok, util = pcall(require, "packer.util")
```
ubmit, line 15: Really like the `pcall` addition! Had an error when I first ran it because of the issue you're solving here 🙌 

What do you think about using `packer_util` instead of just `util`? I think it would be more semantic that way.

robsonpeixoto: Is is a shameless copy from https://github.com/LunarVim/Neovim-from-scratch/blob/2c7a64973602e8d9d18f6de75022014e96fa576a/lua/user/plugins.lua#L26-L30 =D

About the `packer_util`, done in 7881ea08439227fbda12d07c1fb3e6b8dbace53e

ubmit: Great 🙌

tjdevries:
Thanks, but i'm switching to lazy for the project, you can follow that in #178 .

I appreciate the PR

PR #143, opened 2023-01-21
Added friendly snippets plugin to LuaSnip

DanielBrittainJr:
The friendly snippets plugin is super helpful and does not bloat the config very much, since we are adding only two lines.

As a beginner, it took me some time to figure out how to add this to the Kickstart config. So, I think this will be helpful for other users like myself who start with ThePrimegeans config video and then convert to Kickstart. I like this config a lot and plan on sticking with it, but this was something I immediately missed, and think that due to the audience overlap others may run into this too. Granted it is very simple, but I hate to admit it took me a long time to figure out how to add it, so maybe other beginners may struggle as well.

Feel free to disregard this if you don't want it! This is my first-ever PR so I apologize if I have done something incorrectly.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -30,6 +30,11 @@ require('packer').startup(function(use)
     'hrsh7th/nvim-cmp',
     requires = { 'hrsh7th/cmp-nvim-lsp', 'L3MON4D3/LuaSnip', 'saadparwaiz1/cmp_luasnip' },
   }
+  
+  -- add friendly snippets to LuaSnip
+  use 'rafamadriz/friendly-snippets'
+  require("luasnip.loaders.from_vscode").lazy_load()
+
 
   use { -- Highlight, edit, and navigate code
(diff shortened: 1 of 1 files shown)


feoh:
Hiya! Thanks for your contribution. I struggled with this too. 

This project walks a difficult to manage line between providing a good starting point but not making too many assumptions around what people want.

There are conflicts to resolve between this PR and the current master, let's think on this and figure out what makes sense.

SamPosh:
@feoh I think this is a good idea to add this. But seems like he wrote it when it was packer based. I tried it lazy but it keeps failing. If it is added here it will be helpful. Seems many people loved this feature

SamPosh:
@DanielBrittainJr have you had a chance to rewrite this in lazyvim .seems you wrote that during kickstart was using packer as package manager.

feoh:
Closing this in favor of #296

PR #144, opened 2023-01-21
feat: Add treesitter auto_install

albertocerrone:
Easily allow user to decide if to auto_install parsers in case are missing.


tjdevries:
This is a good idea, I'll add it in my refresh of #178 . Thanks

albertocerrone:
You are welcome @tjdevries and thank you for the kickstarter! ❤️

PR #147, opened 2023-01-23
Adding which key

rjshrjndrn:
This will help freshers get used to keyboard shortcuts easily.


tytrdev:
Huge fan of which key. With a delay I think it's good even for experienced users. +1 for this!

tjdevries:
Adding in #178

PR #156, opened 2023-01-28
feat: Add auto save on InsertLeave event

aidenlangley:
Put together a simple auto save so thought I'd see if it would be desirable here too. If it doesn't make this config too feature rich I could also put together a simple auto format that would prefer null-ls over LSP if present.


tjdevries:
Thanks for the suggestion, but this is much too opinionated for what I'd want in kickstart. Thanks for sending the PR though!

PR #166, opened 2023-02-06
Add clipboard sync

tudorjnu:
I think it might be useful to have this out of the box

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -104,6 +104,11 @@ vim.wo.number = true
 -- Enable mouse mode
 vim.o.mouse = 'a'
 
+-- Sync clipboard between OS and Neovim.
+--  Remove this option if you want your OS clipboard to remain independent.
+--  See `:help 'clipboard'`
+vim.o.clipboard = 'unnamedplus'
+
 -- Enable break indent
 vim.o.breakindent = true
 


mkaito:
@feoh `unnamedplus` is the system clipboard.

tudorjnu:
Hi, sorry for the late reply! Yes, `unnamedplus` is the system clipboard and as such, it allows for nice integration with the system (copying and pasting config files most often). I know some more versed people in VIM do not like it but for me (as a beginner) it helped a lot

tjdevries:
Thanks, I do think this is a pretty reasonable default to have for people.

stefanbildl:
Unnamedplus is a good thing for complete beginners, I get it. 
But most neovim users are not new to vim's default behavior.

If you make the switch to neovim, you most likely already work with registers. Most neovim users will come from vscode or intellij vim plugins. There, the system clipboard is not linked to the default yank.

I think unnamedplus should be disabled again. But keep it as a comment for interested users!

feoh:
I disagree. I think this project has got to maintain clarity of purpose in order to be useful, and my impression of that purpose is to provide an accessible on-ramp for new users to get productive with Neovim in a very short time.

Experienced users can twiddle and tweak as they like, but I can tell you that as a recent-ish new user myself, having the system clipboard Just Work is a really desirable default from where I sit :)

PR #169, opened 2023-02-10
changed from packer to lazy

szechp:
i changed the plugin manager from packer to lazy. 
nvim/lua/custom/plugins.lua
nvim/after/plugin/defaults.lua
are not loading atm, im still trying to wrap my head around how to load this with lazy. suggestions welcome


tjdevries:
Thanks, I'm working on this in #178

Issue #171, opened 2023-02-12
sumneko_lua is deprecated, use lua_ls instead


jschuur:
After updating plugins today (`PackerSync`) I get the following error when I launch neovim:

```
sumneko_lua is deprecated, use lua_ls instead. See :h deprecated
This function will be removed in lspconfig version 0.2.0
stack traceback:
        .../site/pack/packer/start/nvim-lspconfig/lua/lspconfig.lua:41: in function '__index'
```

However, a simple renaming of `sumneko_lua` to `lua_ls` in my `init.lua` did not do the trick

```
lua_ls = {
  Lua = {
    workspace = { checkThirdParty = false },
    telemetry = { enable = false },
  },
},
```

This now results in 

```
[mason-lspconfig.nvim] Server "lua_ls" is not a valid entry in ensure_installed. Make sure to only provide lspconfig server names.
```

Mason lists `lua-language-server` as installed. 

This would seem to be related to [this renaming](https://github.com/neovim/nvim-lspconfig/pull/2439). 

The renamed repo links [here](https://github.com/LuaLS/lua-language-server/wiki/Getting-Started#command-line) for install instructions, but I'm not sure why I'm supposed to do a full brew install here, since I didn't use brew to set up the LSP under the old name.

Is this an issue with things being slightly out of sync, since this rename just popped up in the nightlies?

At this point, I've uninstalled lua-language-server in Mason and commented out the above block in `init.lua` to make the error go away.

I'm on macOS, NVIM v0.8.3.

jschuur:
Maybe I need to wait for [this mason-lspconfig PR](https://github.com/williamboman/mason-lspconfig.nvim/pull/148)?

jinzhongjia:
> I think I just need to wait for [this mason-lspconfig PR](https://github.com/williamboman/mason-lspconfig.nvim/pull/148)?

Yes, now we just need to wait it

tomsiwik:
My biggest pet peeve with nvim configs currently is that I don't know how to pin versions (upgrade/rollback command) and avoid these deprecation warnings (ignore command). Is there a way to manage this comfortably or are you stopping your day to day work to fix these warnings yourself, as a vim user? This would be a deal-breaker for me if I constantly get disturbed by randomly appearing warnings/deprecations and I have to fiddle around my config.

jinzhongjia:
[...]

PR #173, opened 2023-02-13
Use lua_ls as Lua Language Server

Labnann:
sumneko_lua is deprecated. Causing warning like:
```
mason-lspconfig.nvim] Server "sumneko_lua" is not a valid entry in ensure_installed. 
Make sure to only provide lspconfig server names
```
While the language server is not loaded. 
Fix this warning by using lua_ls instead of sumneko_lua.
#171

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -343,7 +343,7 @@ local servers = {
   -- rust_analyzer = {},
   -- tsserver = {},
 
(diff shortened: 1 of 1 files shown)


tjdevries:
Thanks, closing in favor of the work in #178

Issue #174, opened 2023-02-13
Sumneko_lua is not   recognized


kaboussi:
**Describe the bug**

**To Reproduce**
1. ...

**Desktop (please complete the following information):**
- OS: archlinux
- Terminal: alacritty

** Neovim Version **
 - Output of running `:version 0.8.2` from inside of neovim:
```
[mason-lspconfig.nvim] Server "sumneko_lua" is not a valid entry in ensure_installed. Make sure to only pro
vide lspconfig server names.
```

apmwebdev:
I was also running into this, but it looks like you can change `sumneko_lua` to `lua_ls` [here](https://github.com/nvim-lua/kickstart.nvim/blob/master/init.lua#L346) to solve the problem. Someone pointed me to https://github.com/williamboman/mason-lspconfig.nvim/blob/main/doc/server-mapping.md, which has the different LS name mappings. By my understanding `lua_ls` and `sumneko_lua` are both referring to `lua-language-server`.

Ninja-Yubaraj:
I was having the same issue since morning, and only after I did some digging, found out that `sumneko_lua` was missing from `Mason` project. As it turns out it might have gone deprecated (actually been renamed to `lua-ls`), and `lua-ls` is the new norm. 
 [Read More](https://www.reddit.com/r/neovim/comments/1107xiz/how_to_fix_sumneko_lua_is_deprecated_use_lua_ls/)

### Fix 1 (removing lua support alltogether):
To fix the issue, you just need to remove  few lines from the `init.lua` file:

```lua
sumneko_lua = {
    Lua = {
      workspace = { checkThirdParty = false },
      telemetry = { enable = false },
    },
  },
  ```
  
### Alt Fix (Rename the package):
  And by any chance if you need lsp support for lua, just rename `sumneko_lua` to `lua_ls` in `init.lua`.
  
  For those who did'nt customized `init.lua` according to their needs:
However, lemme be clear, this is kind of a "hotfix" as Imm sure the someone soon will do a PR and fix this issue, so just check back for updates in some time.

 For those who did:
 You guys know what you're doing, so chill...

**Comment by ykhan21** [comment]
I can fixed it by renaming `sumneko_lua` to `lua_ls`.

```lua
lua_ls = {
    Lua = {
      workspace = { checkThirdParty = false },
      telemetry = { enable = false },
    },
  },
```

monkeyribbon:
[...]

PR #175, opened 2023-02-13
switching from sumneko_lua to luau

kaboussi:
#174


tjdevries:
Thanks, going to update in separate PR

Issue #176, opened 2023-02-13
No plugins from custom.plugins is loaded


jby:
I have the same problem as described in #34.
I've commented in that issue quite a bit.
**To Reproduce**
1. Add plugins to `~/.config/nvim/lua/custom/plugins.lua`
2. Re-start vim or run `PackerSync`
3. No plugins from custom.plugins is installed

**Desktop (please complete the following information):**
- OS: macOS 13.2
- Terminal: iTerm2

** Neovim Version **
 - Output of running `:version` from inside of neovim:

```
NVIM v0.8.3
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by brew@Ventura-arm64.local

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/opt/homebrew/Cellar/neovim/0.8.3/share/nvim"
```
My current `~/.config/nvim/lua/custom/plugins.lua`: https://github.com/jby/nvim/blob/nvks/lua/custom/plugins.lua
```lua
return function(use)
  -- Configure nvim-tree
  use({
    {'nvim-tree/nvim-tree.lua', requires = 'nvim-tree/nvim-web-devicons'},
    config = function()
      require("nvim-tree").setup()
      'nvim-tree/nvim-web-devicons'
    end
  })
  -- Add friendly snippets
  use({
    'rafamadriz/friendly-snippets',
    config = function ()
      require('luasnip.loader.from_vscode').lazy_load()
    end
  })
end
```

BKDevelop:
ok I believe the problem is that you are using invalid Packer syntax (see [packer readme](https://github.com/wbthomason/packer.nvim#quickstart) for reference)

This should work

```
return function(use)
  -- Configure nvim-tree
  use {
    'nvim-tree/nvim-tree.lua', requires = 'nvim-tree/nvim-web-devicons',
    config = function()
      require("nvim-tree").setup()
      'nvim-tree/nvim-web-devicons'
    end
  }
  -- Add friendly snippets
  use {
    'rafamadriz/friendly-snippets',
    config = function ()
      require('luasnip.loader.from_vscode').lazy_load()
    end
  }

end
```

@tjdevries  I guess the readme should be updated? I won't submit a PR because I migrated to my own config a while ago and just got curios because @jby  resurrected my old thread. 

@jby maybe you wan't to contribute here?

Have a nice day all around!

[...]

PR #177, opened 2023-02-15
Corrected packer syntax according to comment in #176

jby:
This will solve most of the problem reported in #176


tjdevries:
Thanks, closing cause i'm removing packer in current PR

PR #182, opened 2023-02-17
feat: Enable relative line number by default

yosuanicolaus:
Enable relative line number by default

I think it might be very useful to have this turned on by default. 
It's one of the greatest feature from vim IMO- (as a beginner) 
being able to jump to whatever line by just looking at the relative line numbers, and then "7j" / "12k" / etc.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -198,6 +198,9 @@ vim.o.hlsearch = false
 -- Make line numbers default
 vim.wo.number = true
 
+-- Enable relative line number
+vim.o.relativenumber = true
+
 -- Enable mouse mode
 vim.o.mouse = 'a'
 


szechp:
Please leave it as it is, I hate this feature and I think it's really annoying for pair programming or pointing something out to a colleague in a zoom meeting. If you need it it's easy to enable. 

It would be cool tough to have both 😅

stefanbildl:
I think it shouldn't be the default, but maybe include it as a comment.

szechp:
I just found out there is a hybrid option in vim, where the current one displays the actual line number instead of zero. That would be a nice compromise

szechp:
This is really cool aswell: 
Relative in normal mode, absolute in insert mode:
```
autocmd InsertEnter * :set norelativenumber
autocmd InsertLeave * :set relativenumber 

```

But I really have to say, by doing things like this we get away from what the purpose of this repo is: a minimal starting point for people to get into neovim. The more stuff we put in there, the closer this gets to a full blown preconfig like astro or lunar.

yosuanicolaus:
I didn't realize that relativenumber was a pretty opiniated/debated topic in vim. I thought it was the standard to use them haha. Well, it's completely fine then. Should I close this PR?

PR #189, opened 2023-02-20
docs: add comment about WSL clipboarding

todpunk:
Through much troubleshooting, this line was the source of my hanging issues on WSL, which makes sense since there's more layers interacting across threads. I've been testing with this disabled for 2 days now and it's never hung since, after a day of hanging pretty often on `p`

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -203,6 +203,7 @@ vim.o.mouse = 'a'
 
 -- Sync clipboard between OS and Neovim.
 --  Remove this option if you want your OS clipboard to remain independent.
+--  If you are on WSL2, you should remove this line or nvim may hang on paste
 --  See `:help 'clipboard'`
 vim.o.clipboard = 'unnamedplus'
 


codeuntilcold:
Upvote! Another problem with this line on WSL2 is that it makes startup time for nvim really slow! On my machine using [vim-startuptime](https://github.com/dstein64/vim-startuptime) it recorded that `clipboard.nvim` took >600ms to load.

flamendless:
Same here

feoh (commented):
Hi guys!

First off thanks for your contribution!

While I love the idea of helping everyone no matter what platform they're using fall into
the "pit of success" I'm not sure kickstart.nvim is the right place to solve this problem.

You can see from this StackOverflow [question](https://stackoverflow.com/questions/61550552/how-to-copy-paste-from-vim-in-wsl) that there are many moving parts and a lot of complexity here.

There are a ton of issues, of which [this](https://github.com/neovim/neovim/issues/20396) is just one example on the main Neovim repository around this issue, please consider pursuing a solution over there so everyone can benefit.

feoh:
Closing for now. Please see above comment and feel free to re-open if you disagree.

PR #191, opened 2023-02-21
added descriptions to Diagnostic keymaps

szechp:
since kickstart is using whichkey by default now, i thought it would be a good idea to get the keymaps in line with all the others.


tjdevries:
Thanks!

PR #194, opened 2023-02-22
add options table for Lazy package manager

TomDeneire:
It took me some time to figure out where to add configuration for the Lazy package manager. By adding an empty table accompanied with a note, this could help others find their way more quickly.

Btw: thanks, @tjdevries, for your superb work!

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -35,7 +35,6 @@ I hope you enjoy your Neovim journey,
 
 P.S. You can delete this when you're done too. It's your config now :)
 --]]
-
 -- Set <space> as the leader key
 -- See `:help mapleader`
 --  NOTE: Must happen before plugins are required (otherwise wrong leader will be used)
@@ -57,6 +56,10 @@ if not vim.loop.fs_stat(lazypath) then
   }
 end
 vim.opt.rtp:prepend(lazypath)
+
+-- NOTE: Here is where you configure the Lazy package manager
+-- See `:help lazy.nvim-lazy.nvim-configuration` for more info
+local lazyopts = {}
 
 -- NOTE: Here is where you install your plugins.
 --  You can configure plugins using the `config` key.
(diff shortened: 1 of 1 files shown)


feoh:
Hiya! I'm confused as to what you're trying to do with this PR.

When I look at that lazy.nvim documentation, I mostly see options being specified on a per plugin basis as they are currently in kickstart.nvim

Can you please help me understand the advantages to merging this PR?

feoh:
Sorry to belabor your point. I do get that, but what I'm trying to understand is, OK. Your change lays the groundwork for adding addition lazy.nvim options that are plugin inspecific. Great. Why would most users want or need to utilize those options? What critical lack does adding this framework address?

I'm just trying to balance making this repo as useful as we can to everyone against not letting it become so huge and complex that it's harder to maintain. We're already getting feedback about that from some prior implementors.

Thanks for working with me on this.

TomDeneire:
No problem. My reason for this PR was that it took me (switching from vim to Lua config) quite some time to figure out how to configure lazy.nvim when I first installed kickstart. I figured that since the idea of kickstart is also to offer a scaffold for future user config, you might as well include this small addition which might save others some time. After all, the plugin manager is no small part of a good Neovim setup and lazy.nvim config contains some very important stuff which many kickstart users may want to change (like default lazy loading, install missing, cache settings,...) But if you think it's unnecessary, I'm not going to die on this hill 😉

feoh:
Pretty sure we have a project goal of "No dying on hills" :)

I tried staring at the lazy.nvim docs again and I can't see how new users would ever want to tweak anything in the option, but I'm SURE I'm missing something.

@tjdevries care to take a gander at this one when you can and weigh in please?

feoh:
Thanks for your contribution, but I still can't figure out what benefits the end user derives from this, so I'm closing it for now.

PR #210, opened 2023-03-06
fix: move the Mason setup time forward from before

KudoLayton:
This commit fix the Mason [#1045](https://github.com/williamboman/mason.nvim/issues/1045) issue. Quickly set up Mason to avoid DAP-related startup error messages.


cramsted:
Can confirm, fixes the issue.

tjdevries:
Thanks!

PR #211, opened 2023-03-08
Fix comment about loading kickstart plugins

mkochenough:


feoh:
Thanks for your contribution! I think the confusion here is that you can't just copy init.lua into your config. You must follow the install instructions as specified in [README.md](https://github.com/nvim-lua/kickstart.nvim#readme).

Closing for now.

PR #178, opened 2023-02-15
feat: move to lazy and fix many issues

tjdevries:
Closes #175
Closes #177
Closes #173
Closes #169
Closes #161
Closes #144
Closes #138
Fixes #136
Closes #137
Closes #131
Closes #117
Closes #130
Closes #115
Closes #86
Closes #105
Closes #70
Fixes #176
Fixes #174
Fixes #160
Fixes #158

PR #225, opened 2023-03-20
prevent error if you dont have a custom plugin

DexMech:
comment this line to prevent errors when we dont have custom plugins yet


amtoine:
@DexMech 
i think this is already being taken care of by the [`lua/custom/plugins/init.lua`](https://github.com/nvim-lua/kickstart.nvim/blob/master/lua/custom/plugins/init.lua) files introduced in #178 :yum:

Issue #257, opened 2023-04-09
enable kickstart.plugins.debug attemp to call field 'setup_handlers' (a nil value)


Drakx:
**Describe the bug**
Enabling kickstart.plugins.debug causes error on initial startup with ('a nil value')

Failed to run `config` for nvim-dap
```
~/.config/nvim/lua/kickstart/plugins/debug.lua:45: attempt to call field 'setup_handlers' (a nil value)
# stacktrace:
  - nvim/lua/kickstart/plugins/debug.lua:45 _in_ **config**
  - nvim/init.lua:66
```

**To Reproduce**
1. rm -rf ~/.config/nvim ~/.local/share/nvim ~/.cache/nvim
2. git clone git@github.com:nvim-lua/kickstart.nvim.git
3. nvim (allow for initial plugins to install)
4. in init.lua find 'require kickstart.plugins.debug' and change to `{ import = 'kickstart.plugins.debug' },`
5. close nvim and reopen

**Desktop (please complete the following information):**
- OS: macOS
- Terminal: alacritty

** Neovim Version **
 - Output of running `:version` from inside of neovim:

```
:version
NVIM v0.9.0
Build type: Release
LuaJIT 2.1.0-beta3
```

PR #258, opened 2023-04-09
mason nvim dap 2.0 removed setup_handlers()

Drakx:
mason released 2.0 which removed `setup_handlers()` and broke the `kickstart.plugins.debug` when enabled


Drakx:
initial issue https://github.com/nvim-lua/kickstart.nvim/issues/257

tjdevries:
Thanks!

Issue #271, opened 2023-04-19
module 'nvim-treesitter-textobjects' not found:


staycreativedesign:
Ran the init and 
```
Last login: Wed Apr 19 11:37:48 on ttys000
developer@Gustavos-MBP ~ % source ~/.zshrc
developer@Gustavos-MBP ~ % brew install neovim
==> Downloading https://formulae.brew.sh/api/formula.jws.json
######################################################################## 100.0%
==> Downloading https://formulae.brew.sh/api/cask.jws.json
######################################################################## 100.0%
==> Fetching dependencies for neovim: unibilium, libtermkey, libuv, libvterm, luajit, luv, msgpack and tree-sitter
==> Fetching unibilium
==> Downloading https://ghcr.io/v2/homebrew/core/unibilium/manifests/2.1.1-1
######################################################################## 100.0%
==> Downloading https://ghcr.io/v2/homebrew/core/unibilium/blobs/sha256:d437072bceb93b39d6231dc1132f10284c7033690e2e8fe85193670157c680a2
==> Downloading from https://pkg-containers.githubusercontent.com/ghcr1/blobs/sha256:d437072bceb93b39d6231dc1132f10284c7033690e2e8fe85193670157c680a2?se=2023-04-19T15%3A55%3A00Z&sig=HxdDyphfMOtIAEYeC8gDVs5ATm%2FHQhZl%2BCR2PBIf%2Fy0%3D&sp=r&spr=https&sr=b&sv=2019-12-12
######################################################################## 100.0%
==> Fetching libtermkey
==> Downloading https://ghcr.io/v2/homebrew/core/libtermkey/manifests/0.22-1
######################################################################## 100.0%
==> Downloading https://ghcr.io/v2/homebrew/core/libtermkey/blobs/sha256:4a463c5f31b1748ce885716a2f709f3ff1791725bb67e71bd9b44080148d6ff2
==> Downloading from https://pkg-containers.githubusercontent.com/ghcr1/blobs/sha256:4a463c5f31b1748ce885716a2f709f3ff1791725bb67e71bd9b44080148d6ff2?se=2023-04-19T15%3A55%3A00Z&sig=QB1cgBy5p%2BKuMFdByAZKJWVazi2AhapyRae6Xw0hI6M%3D&sp=r&spr=https&sr=b&sv=2019-12-12
######################################################################## 100.0%
==> Fetching libuv
[...]

PR #272, opened 2023-04-19
#271 - Remove treesitter.module.textobjects explicit dependancy

feoh:
Explicitly having this dependency in our init.lua is causing errors on startup for some people.

Since near as I can tell it's not needed I'm removing it.

Please test this on your system and approve this PR if it works for you, and we'll get it merged.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -35,7 +35,6 @@ I hope you enjoy your Neovim journey,
 
 P.S. You can delete this when you're done too. It's your config now :)
 --]]
-
 -- Set <space> as the leader key
 -- See `:help mapleader`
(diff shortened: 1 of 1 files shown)


Drakx:
It looks as though its been removed already https://github.com/nvim-lua/kickstart.nvim/blob/master/init.lua#L96

Drakx:
Seems to work fine with the little testing I've done.

WindSoilder:
Hi, I don't think we need to remove it, the issue is hightly due to the version of neovim.  I'd suggest to ask user to update neovim version and see how is going

feoh:
Almost certainly not necessary. Closing.

PR #273, opened 2023-04-20
Python debug configuration is added

SamPosh:
Python debug DAP configuration is added for handlers.

This sample can be helpful for beginners who wanted to debug python.

I spent 2 weeks to figure it out. Thought of adding the sample here so that others need not to spend the same amount of time. 

Thank you so much for this kickstart.nvim .

### DIFF
--- a/lua/kickstart/plugins/dap/handler/python.lua
+++ b/lua/kickstart/plugins/dap/handler/python.lua
@@ -1,8 +1,9 @@
 local dap = require('dap')
 return function()
+    local python_install_path = vim.fn.exepath('python')
     dap.adapters.python = {
       type = "executable",
-      command = "/usr/local/bin/python", -- use "which python" command and provide the python path
+      command = python_install_path, -- use "which python" command and provide the python path
       args = {
         "-m",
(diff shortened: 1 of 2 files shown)


feoh:
Hi, thanks for the contribution but I have very mixed feelings about this one.

This isn't how I configure nvim-dap-python, and I think a single example will suffice just fine.

I'm tempted to close this unless you can make a super compelling argument for your change :)

SamPosh:
I have added pytest debug configuration and profiler debug configuration as well.
For a beginner there will be confusion where to configure what , so I added it.

If you think these additional configurations can be added in much simpler way then it would be good if you can add those examples (I also will learn from it) :)

feoh:
I can appreciate that. The problem is that there are a million different languages and everyone has different favorites. If we add them all kickstart will become ENORMOUS and impossible to maintain.

Here's a [link](https://github.com/feoh/git_dotfiles/blob/master/.config/nvim/lua/kickstart/plugins/debug.lua) to my debug configuration that might give you some ideas, but I'm closing this for now. Thanks again for the contribution!

PR #280, opened 2023-04-24
Try kickstart in container

SamPosh:
To try kickstart.nvim in docker container.

If anybody wants to try kickstart.nvim in docker container ,they can try this.


feoh:
Hi. Thanks for your contribution but I need to close this one. Please take a look over our Issues list. So many people have all kinds of issues with the repository as it stands. I do not think we should add support for Docker into the mix. The number of potential support queries and issues will skyrocket.

PR #281, opened 2023-04-24
Dapui.toggle keymap is added

SamPosh:
DapUI toggle example is added.

Without toggle function, after the debugging is over we can't see the last session result.
Most of the time it will be required especially when unhandled exception closes the dapui, this toggle will be helpful to see the last time result.

### DIFF
--- a/lua/kickstart/plugins/debug.lua
+++ b/lua/kickstart/plugins/debug.lua
@@ -72,7 +72,9 @@ return {
         },
       },
     }
-
+    -- toggle to see last session result. Without this ,you can't see session output in case of unhandled exception.
+    vim.keymap.set("n", "<F7>", dapui.toggle)
+    
     dap.listeners.after.event_initialized['dapui_config'] = dapui.open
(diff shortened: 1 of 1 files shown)


feoh:
I personally like this idea and have something [similar](https://github.com/feoh/git_dotfiles/blob/master/.config/nvim/lua/kickstart/plugins/debug.lua#L57) in my configuration.

However we are forever riding the line with this project in providing a starting point versus dictating our preferences to others, so I'd love it if other folks would chime in on this one.

SamPosh:
@feoh when I play with debugger ,whenever debug session is closed I could not see the last session result. In my personal configuration I have added multiple other plugins but I believe this example should be added here so that others can also be benefited.

feoh:
@SamPosh I agree! :) I'm just looking for someone else's opinion as well :)

feoh:
The more I think about this the more I think there's no harm at all in it and it could help a bunch of people. Merging.

PR #283, opened 2023-04-24
A bit of clarification

nPHYN1T3:
Due to the previous experience I'm not 100% if the archive instructions are factual as I didn't nuke my settings to retry. I am also not 100% on the MacOS/Win directories.

I'm too old for this sh1t / did I do this right?

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Distribution Alternatives:
 * Extract the archive to:
   `~/.config/nvim` (Linux)
   `~/.config/nvim` (MacOS)
-  `C:\User\YourName\AppData\Local\nvim\` (Windows)
+  `%userprofile%\User\YourName\AppData\Local\nvim\` (Windows)
 * Ensure your extraction method did not extract with a parent folder. For example in ~/.config/nvim you should have init.lua not another folder called kickstart.nvim.
 
 ### GIT Clone Installation
@@ -40,7 +40,7 @@ Distribution Alternatives:
 * **You're ready to go!**
 
 * (Recommended/Optional) Fork this repo (so that you have your own copy that you can modify).
-* Clone the kickstart repo into `$HOME/.config/nvim/` (Linux/Mac) or `C:\User\YourName\AppData\Local\nvim\` (Windows)
+* Clone the kickstart repo into `$HOME/.config/nvim/` (Linux/Mac) or `%userprofile%\User\YourName\AppData\Local\nvim\` (Windows)
   * If you don't want to include it as a git repo, you can just clone it and then move the files to this location
 
 Additional system requirements:
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Distribution Alternatives:
 * Extract the archive to:
   `~/.config/nvim` (Linux)
   `~/.config/nvim` (MacOS)
-  `%userprofile%\User\YourName\AppData\Local\nvim-data\` (Windows)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -112,7 +129,7 @@ Each PR, especially those which increase the line count, should have a descripti
   * See [lazy.nvim uninstall](https://github.com/folke/lazy.nvim#-uninstalling) information
 * Are there any cool videos about this plugin?
   * Current iteration of kickstart (coming soon)
-  * Here is one about the previous iteration of kickstart: [video introduction to Kickstart.nvim](https://youtu.be/stqUbv-5u2s).
+  * Here is one about the previous iteration of kickstart: [video introduction to Kickstart.nvim](https://youtu.be/stqUbv-5u2s). Note the install via init.lua no longer works.
```
feoh, line 132: I think signposting this video as obsolete is a really good call, but might it be better to just say something like "Note the install instructions here no longer work. Just follow the instructions above instead" or similar?

feoh (changes requested):
You did GREAT and you are most certainly NOT too old for this :) We're not too old until we're DEAD I'll bet you'd be shocked to learn how old I am. Let's put it this way, I'd be 1 year away from retirement a generation ago :)

Just a couple suggestions for improvement inline.

nPHYN1T3:
Good catch, no the Win bit won't work but not because of the Tilde...I mean that's an issue too but the whole /here/there/everywhere when Windows should be DRIVE:\here\there\everywhere. I just stole that bit from a section further down in the original readme heh.

So how the hell do I make the revision now in a way that keeps this as a pull?

Dude I'm very too old for this shit. I've been coding since the 1980's and while I'm not 100% inline with Stallman I'm close enough to make dealing with all this "new stuff" a problem. I distrust if not hate all this new fangled crap like npm, composer, git...it feels like automatic bank withdrawls...silently slipping malware into everything while everyone sleeps on the EZ automation.

*SHAKES CANE!*
*YELLS AT CLOUD!*

feoh:
> Good catch, no the Win bit won't work but not because of the Tilde...I mean that's an issue too but the whole /here/there/everywhere when Windows should be DRIVE:\here\there\everywhere. I just stole that bit from a section further down in the original readme heh.

That's why I suggested $HOME. So something like:

`$HOME\AppData\Local\nvim\` 

That way, it'll work whether the user has their home set to C, D, or wherever. Windows will totally support this in Powershell and most other apps.

[8< Old person yells at cloud snip :) 8<]

> OK I made a change on mine to fix the Win paths.
> 
> As a side note I "dun exploded" my git clone install the other day and ended up testing my Archive install for others who hate git cloning and it works so weee...
> 
> Please let me know if I need to do another pull to make the update sync.

You're good! No need to re-pull. If you commit and push to your branch, those changes are reflected in the current pull.

And seriously thanks a LOT for contributing back to the project!

nPHYN1T3:
On Windows iirc $HOME will be %HOME%, I can change to that.

README.md
feoh: The rest looks good. I'm really concerned that this particular turn of phrase will confuse newcomers.  Please consider changing this to just ask folks to follow these instructions instead and I'll merge this in a jiffy :)

feoh: OK, this is excellent work. I think I'm going to merge and then make that one last wording tweak around the out of date install information in the video myself. Thanks a bunch!

PR #296, opened 2023-05-02
Added friendly snippet support

SamPosh:
Added friendly snippet support. 

Previously some one tried to add it when Kickstart was based on Packer. 

But after lazy vim it is not added. So added that support in this.

Think this will help a bunch

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -93,7 +93,7 @@ require('lazy').setup({
 
   { -- Autocompletion
     'hrsh7th/nvim-cmp',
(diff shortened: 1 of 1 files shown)


mech-a:
@feoh I think this is a sound addition. There's no change to start-up time (0.24ms) and it adds reasonable snippets that are useful out of the box and make the experience more smooth. This implementation is implemented as recommended from LuaSnip. 

On the side, I do think cmp in general warrants a bit more documentation in the init.lua, which I'll make an issue for.

feoh:
thanks so much for weighing in! I agree!

PR #302, opened 2023-05-07
change luasnip tab jumping to locally jumpable

mech-a:
**Issue**: When using `luasnip.expand_or_jumpable`, if our window includes the snippet but our cursor is not within it, hitting tab can bring us back into the snippet, an annoying and unexpected behavior for new users. 

This is especially annoying when a user manually edits a snippet (e.g. by not using `Tab`, scrolling over to it); LuaSnip or cmp will not register that the snippet is done, and hitting tab takes us back to the snippet unexpectedly.

**Proposed Solution**: Changing this to `luasnip.expand_or_locally_jumpable` would make it so that we'd only jump if our cursor is inside of the snippet, which I believe more users are used to having from other development environments. Carry this forward to `S-Tab` behavior by switching `jumpable(-1)` to `locally_jumpable(-1)`.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -478,7 +478,7 @@ cmp.setup {
     ['<S-Tab>'] = cmp.mapping(function(fallback)
       if cmp.visible() then
         cmp.select_prev_item()
-      elseif luasnip.jumpable(-1) then
+      elseif luasnip.locally_jumpable(-1) then
         luasnip.jump(-1)
       else
         fallback()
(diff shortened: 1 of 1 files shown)


mech-a:
Here's an example of the behavior that this avoids:

https://github.com/nvim-lua/kickstart.nvim/assets/11798509/35f5c0eb-88e6-405e-988b-f3d73f6289fb

feoh:
Thanks for this! To be honest I'm struggling a bit understanding the LuaSnip stuff. I feel like the way I have it configured in my own repo isn't great, and the more docs I read the more confused I get :)

Not to punish a good deed, but would you please consider commenting on the two FriendlySnips PRs:

https://github.com/nvim-lua/kickstart.nvim/pull/296
https://github.com/nvim-lua/kickstart.nvim/pull/143

Issue #277, opened 2023-04-23
`vim.treesitter.get_query() is deprecated` warning when fuzzy searching


krlenell:
## Describe the bug

When attempting to fuzzy search by typing the spacebar followed by '/' this warning appears. This happened after running `brew update` and `brew upgrade` earlier. Fuzzy searching can still be done, but the alert appears anyways.

Here is the full warning:

```
vim.treesitter.get_query() is deprecated, use vim.treesitter.query.g
et_query() instead. :help deprecated
stack traceback:
        ...r/neovim/0.9.0/share/nvim/runtime/lua/vim/treesitter.lua:
28: in function '__index'
        ...im/lazy/telescope.nvim/lua/telescope/builtin/__files.lua:
435: in function 'v'
        ...im/lazy/telescope.nvim/lua/telescope/builtin/__files.lua:
570: in function 'v'
        .../nvim/lazy/telescope.nvim/lua/telescope/builtin/init.lua:
511: in function 'current_buffer_fuzzy_find'
        /Users/name/.config/nvim/init.lua:279: in function </Users/be
vim.treesitter.query.get_query() is deprecated, use vim.treesitter.q
uery.get() instead. :help deprecated
stack traceback:
        ...im/0.9.0/share/nvim/runtime/lua/vim/treesitter/query.lua:
200: in function <...im/0.9.0/share/nvim/runtime/lua/vim/treesitter/
query.lua:199>
        [C]: in function 'pcall'
        ...im/lazy/telescope.nvim/lua/telescope/builtin/__files.lua:
435: in function 'v'
        ...im/lazy/telescope.nvim/lua/telescope/builtin/__files.lua:
570: in function 'v'
        .../nvim/lazy/telescope.nvim/lua/telescope/builtin/init.lua:
511: in function 'current_buffer_fuzzy_find'
        /Users/name/.config/nvim/init.lua:279: in function </Users/be
p/.config/nvim/init.lua:277>
Press ENTER or type command to continue
```

Specified lines in init.lua:
```
vim.keymap.set('n', '<leader>/', function()
  -- You can pass additional configuration to telescope to change theme, layout, etc.
  require('telescope.builtin').current_buffer_fuzzy_find(require('telescope.themes').get_dropdown {
    winblend = 10,
    previewer = false,
  })
end, { desc = '[/] Fuzzily search in current buffer' })
```

## To Reproduce

1. Likely need to update treesitter and nvim
2. Press space
3. press `/`

[...]

PR #303, opened 2023-05-08
Telescope to use latest release branch

declan-whiting:
Fix for #277 

From Telescopes ReadMe:

> It is suggested to either use the latest release [tag](https://github.com/nvim-telescope/telescope.nvim/tags) or our release branch (which will get consistent updates) [0.1.x](https://github.com/nvim-telescope/telescope.nvim/tree/0.1.x).

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -35,7 +35,6 @@ I hope you enjoy your Neovim journey,
 
 P.S. You can delete this when you're done too. It's your config now :)
 --]]
-
 -- Set <space> as the leader key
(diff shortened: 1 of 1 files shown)


mech-a:
@feoh I don't believe we'd have to worry - the change in this PR is to track the release branch `0.1.x`, which should be receiving consistent updates as per the readme. New users can use Lazy to update the plugin as needed, and the kickstart config should be up to date as long as Telescope uses branches 0.1.x, which seems to be for the long run.

PR #307, opened 2023-05-11
implemented autopairs default plugin

vkuppoor:
Hey guys, I'm taking this 1-credit college class on github and project management tools. For the final project, I have to make an open source contribution. So, for my first opensource contribution, I thought I'd add an autopairs plugin as a sensible default plugin--I've made it optional to enable in the main init.lua file. Feel free to reject this PR--I'm just doing it to learn. But I'd be very happy if it were accepted as there is extra credit if its accepted!

Thanks for reviewing!

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -175,6 +175,7 @@ require('lazy').setup({
   --       Uncomment any of the lines below to enable them.
   -- require 'kickstart.plugins.autoformat',
   -- require 'kickstart.plugins.debug',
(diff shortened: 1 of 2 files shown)


feoh:
Thanks very much for your contribution! However the point of kickstart.nvim is to provide a framework to allow people to pick and choose their own plugins that they can add.

I'm closing this for now.

PR #203, opened 2023-02-27
Features to move to diffs and show the preview values.

luizcalliari:
Adding `[+c` and` ]+c` to move to diffs and `leader+ph` to show a pop up with the preview diff value bellow the cursor.

PR #313, opened 2023-05-18
Fix init.lua conflicting (with vimdiff) keybinding

componhead:
`[c` and `]c` conflict with vimdiff and is not overwritable

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -112,8 +112,6 @@ require('lazy').setup({
         changedelete = { text = '~' },
       },
       on_attach = function(bufnr)
-        vim.keymap.set('n', '[c', require('gitsigns').prev_hunk, { buffer = bufnr, desc = 'Go to Previous Hunk' })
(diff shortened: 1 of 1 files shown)


feoh:
Hi, could you please work with @luizcalliari who introduced this change in https://github.com/nvim-lua/kickstart.nvim/pull/203 ?

@componhead - If you can come to agreement with @luizcalliari on a solution and they agree to revert their change, please do this by publishing a revert of their commit instead of just papering over the file again.

Thanks all.

componhead:
Change #203 is overwritten by merge commit 66dc5b8

luizcalliari:
Hello. Sorry, I couldn't answer yesterday. I believe just changing the keys would be a solution.
Maybe use: 
```
vim.keymap.set('n', '<leader>gp', require('gitsigns').prev_hunk, { buffer = bufnr, desc = '[G]o to [P]revious Hunk' })
vim.keymap.set('n', '<leader>gn', require('gitsigns').next_hunk, { buffer = bufnr, desc = '[G]o to [N]ext Hunk' })
```

componhead:
Definitely. Thank you.

PR #315, opened 2023-05-20
Template for optionally using system installed language servers on a per-server basis

cxxsucks:
Added the option to use locally installed language server executables rather than having to use those installed by `mason`.

This option is helpful in that some language servers are installed along with language compiler by package manager, like `clangd` is usually bundled with `clang/LLVM`. Having the underlying system's package manager would update language servers along with the rest of system as well as reducing duplicate downloading and installation introduced by `mason`, saving some disk spaces.

The option is disabled by default and *no specific language server config changes are made*, so as to be compatible with previous versions. Existing user-specific configurations can merge this change seamlessly. 

Choice of using system servers is done on a per server basis, meaning that a user can, say, opt for locally installed `/usr/bin/clangd` while also have `lua_ls` managed by `mason`.  
Detailed comments on instructions of configuring locally installed servers is provided, along with explanations of logic code.

I am well aware of that commits on custom language server configuration is forbidden to be merged. While this commit is indeed about language server configuration, it is *not customized* in any sense, since what this commit offers is a template for using locally installed servers, but no specific change is written into that template.  
In other words, this commit induces no imminent changes to users since no customization is made. **The updated version is still an empty language server configuration as before**, but subsequent users now have one more option when doing customization: an option to use some locally installed servers. And they can safely ignore that and configure like before since the option is, by default, disabled. 

#### Demo of using a system language server:

Screenshot_2023-05-20_10-34-11  
(Enabled `clangd` for the screenshot, which is **not** committed. The committed `init.lua` has default settings as before.)

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -416,11 +416,17 @@ end
 --
 --  Add any additional override configuration in the following tables. They will be passed to
 --  the `settings` field of the server config. You must look up that documentation yourself.
+--
+--  All servers, downloaded or installed locally, MUST be added and have their `settings`
+--  field configured here, or they will not be set up.
+--  Servers meant to be automatically downloaded and installed should only appear here,
+--  but not in the `system_server_cmds` table right after this. Servers already installed
+--  on your system should appear both here AND the `system_server_cmds`.
 local servers = {
   -- clangd = {},
   -- gopls = {},
-  --  pyright = {},
-  --  rust_analyzer = {},
+  -- pyright = {},
+  -- rust_analyzer = {},
   -- tsserver = {},
 
   lua_ls = {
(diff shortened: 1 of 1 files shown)


feoh:
Hi! Thanks for your contribution.

While I can understand your desire to do this, I'm extremely reluctant to merge it and increase the already signint complexity of the kickstart.

Please remember, this is just meant as a starting point for novice users, and you're more than welcome to customize to your heart's content locally.

@tjdevries please lmk if you feel differently (and anyone else from the community please chime in as well!) otherwise I'm tempted to close this.

cxxsucks:
some guys around me, which are not novices, wanted this feature
to debloat their Arch system. Therefore I thought this might be
something many desire.

Turns out for most newcomers complication of config does not worth
the few hundred megabytes of disk space saved :)

Told these Arch guys to use symlinks instead. Closing it for now.

PR #316, opened 2023-05-21
Adding autocompletion to brackets and quotes.

jsphwllng:
Adding a small autocomplete plugin to the plugin list. The functionality is described succinctly [here](https://github.com/cohama/lexima.vim#basic-rules).

This input rule is a common feature in modern IDEs and I think it would make the transition from other editors to NVIM more smooth.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -71,6 +71,9 @@ require('lazy').setup({
 
   -- Detect tabstop and shiftwidth automatically
   'tpope/vim-sleuth',
+    
+  -- Automatically append closing quotes and brackets when typing
+  'cohama/lexima.vim'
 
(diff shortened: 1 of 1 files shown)


mech-a:
I believe most autocomplete bracket suggestions are declined, like #307. The cmp configuration is already quite bloated.

feoh:
I agree. Thank you for your contribution @jsphwllng but I'm afraid we need to close this PR for now. users are free to add this plugin themselves, that's the point of the kickstart! :)

jsphwllng:
Thanks @feoh and @mech-a - I saw some discussions online where people were asking why this feature wasn't included by default in NVIM and thought this would be a good fix for it. WDYT about including it in the commented out plugins list?

PR #323, opened 2023-05-26
FIX: Conflict with vimdiff keybinding

luizcalliari:
I am changing keybinding to solve the conflict with vimdiff.


feoh (approved):
This looks good. Thanks for your contribution!

Issue #314, opened 2023-05-18
Windows path in README wrong?


ckaag:
## Describe the bug
`%userprofile%\AppData\Local\nvim-data\` does not seem to be the correct path for neovim configs under windows (it's used for data as the name implies). The path should be `%userprofile%\AppData\Local\nvim\` (corresponding to ~/.config/nvim on Unix).

It would also probably help to add in a line that a C compiler is required for the plugins to run (e.g. `scoop install gcc`). But that's pretty self-explainable by error messages alone.

## To Reproduce
1. install neovim on Windows (I was using scoop)
2. use the `nvim-data` path 
3. nvim will not load the init.lua

## Desktop
- OS:  Windows 10

## Neovim Version
v0.9.0

feoh:
Please feel free to submit a pull request. You can even do this from the Github web UI without using the command line as well :)

feoh:
I don't currently have a working Windows install to test on. I should revert my laptop to Windws I guess but it's running Manjaro KDE and I've been pretty happy with it :)

PR #326, opened 2023-05-29
fix(docs): update windows path to a correct folder

kaminskypavel:
the path `nvim-data` on windows is not picked up by nvim by default
this PR changes it to the correct folder.

see also https://github.com/nvim-lua/kickstart.nvim/issues/314


kaminskypavel:
tested on Win11 and Nvim 0.9.0

PR #330, opened 2023-06-02
Refactor the theme & status line configurations into their own lua files

feoh:
This change refactors the theme & status line configuration into their own lua files so they can be more readily customized by the user. As kickstart has grown, it's becoming harder and harder to keep up with changes to the main init.lua file if you're maintaining local customizations. We should still encourage users to modify init.lua to suit their needs, but also make it possible to leave it unmodified if that's their choice.

I realize theme related change is a bottomless pit so I'm not suggesting ANY change to the actual theme choices.

However as a partially blind user I simply can't use the onedark scheme chosen, and over time I've struggled with maintaining and integrating my alternate theme choice in amongst the other upstream changes to init.lua, so I think we could save everyone some pain and complexity despite the addition of two more files.

I know that's a bit of a sacred cow to some but I hope you'll at least appreciate the spirit of my suggestion if not the change itself :)

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -129,28 +129,11 @@ require('lazy').setup({
     },
   },
 
-  {
-    -- Theme inspired by Atom
-    'navarasu/onedark.nvim',
-    priority = 1000,
-    config = function()
-      vim.cmd.colorscheme 'onedark'
-    end,
-  },
+  -- Theme related configs go here
+  require 'kickstart.plugins.theme',
 
-  {
-    -- Set lualine as statusline
-    'nvim-lualine/lualine.nvim',
-    -- See `:help lualine.txt`
-    opts = {
-      options = {
-        icons_enabled = false,
-        theme = 'onedark',
-        component_separators = '|',
-        section_separators = '',
-      },
-    },
-  },
+  -- Status line related configs go here
+  require 'kickstart.plugins.statusline',
 
   {
     -- Add indentation guides even on blank lines
--- a/lua/kickstart/plugins/statusline.lua
+++ b/lua/kickstart/plugins/statusline.lua
@@ -1,0 +1,14 @@
+return  {
+  -- Set lualine as statusline
+  'nvim-lualine/lualine.nvim',
+  -- See `:help lualine.txt`
+  opts = {
+    options = {
+      icons_enabled = false,
+      theme = 'onedark',
+      component_separators = '|',
+      section_separators = '',
+    },
+  },
+}
+
--- a/lua/kickstart/plugins/theme.lua
+++ b/lua/kickstart/plugins/theme.lua
@@ -1,0 +1,9 @@
+return {
+  -- Theme inspired by Atom
+  'navarasu/onedark.nvim',
+  priority = 1000,
+  config = function()
+    vim.cmd.colorscheme 'onedark'
+  end,
+}
+


feoh:
@mech-a Because you've been so awesome about contributing lately I'd love your opinion on this as well. I don't want to impose my will on anyone but this will make my life a LOT easier as maintainer :) And I suspect will make it easier on others who can't use the chosen theme as well.

mech-a:
I'm hesitant to add this as it makes kickstart no longer a single file config. I do think that it's not a significant change, but that is some of the marketing to use this repository: it's very simple to read through and understand it. While I do agree that this potentially makes it easier for some continuing users of kickstart, I think that most continuing users aren't affected, and new users aren't really helped.

I think that most continuing users don't actually pull in changes from the main repository, considering that it can override some of your bindings (e.g. the vimdiff changes we've seen) and collide with previous configs. Additionally, kickstart is a template -- they most likely configured it to be pretty far away from the standard kickstart config we see. 

With regard to new users, they have a bit harder onboarding to kickstart. However, they do get more exposure to how to split your lazy config over multiple files, which isn't bad. I commented some areas that we could improve our documentation for this change, if we go through with it.

By the way, I saw that commit [cdaa750](https://github.com/nvim-lua/kickstart.nvim/commit/cdaa750c86cfc0c8f2fc1c8638b7296c327bf5a9)  on master already implements this, and edits the autoformat.lua file -- is this intended to be on master?

skhattane:
Hey guys, it's a subjective choice but my 2 cents is that most users would prefer to keep the onedark theme by default. The tokyonight is a high contrast (I understand the need for @feoh). However the WOW effect for new nvim users when they apply this theme (onedark) for the first time seems to be part of the success of this Kickstarter.

feoh:
Thanks all, I appreciate the feedback and ultimately I agree. Closing this for now and I'll rever tthat errant commit.

feoh:
Just in case anyone stumbles on this, I may have hit upon a good solution previously suggested by @tjdevries - I'd been avoiding forking kickstart.nvim but this may be the perfect use case for it.

It'll mean a merge commit whenever init.lua changes upstream, but the diffs should be SUPER easy to read given that they're just 2 very small code blobs replaced with require statements.

I'm glad everyone liked onedark. The way it makes code comments so incredibly light and ghostly actually makes my eyes literally hurt with strain trying to parse the text so it's [Dracula](https://github.com/Mofiqul/dracula.nvim) for me - almost the exact opposite. Super bright/high contrast.

If anyone in the future wants to see my ongoing version of the changes in situ here's [my fork](https://github.com/feoh/kickstart.nvim).

feoh:
Gotta admit I'm thinking about abandoning kickstart.nvim

Trying to maintain an upstream fork and merge changes down has been a nightmare, and every time I try to live with the onedark theme I literally end up in severe pain with a SCREAMING eyestrain headache.

I love contributing to open source but this is a bit much :)

mech-a:
Really sorry to hear that :( I've appreciated your thoughtful comments and changes!

In case you're still trying to maintain the fork, I thought I might recommend Interactive Patching with a tool like [lazygit](https://github.com/jesseduffield/lazygit), or even just their merge conflict handling. It's helped me out in the past, but it's definitely still extra work to merge in upstream changes.

In case you are all looking for a new maintainer, I'd love to step up and help out.

PR #333, opened 2023-06-06
Override or set new options, autocmds, etc. in lua files placed in lua/custom/

gmnz:


feoh:
I'm really concerned that glob is going to cause undesirable side effects.

I'd love to hear from someone more experienceced but I'm personally SUPER reluctant to merge this.

Please remember, this PR fixes a non issue. The nature of this project is as I was reminded recently myself, a largely single file config. Just change the settings you want in init.lua

mech-a:
I think this adds complication for users using a kickstart config, where it is made quite explicit that one should change the default options themselves. As users explore other configs, they may choose to make a multi-file config, and can do so. I don't think this is a useful addition to kickstart at this time, though I can understand the intention.

feoh:
Thanks a bunch for chiming in @mech-a - I'm going to close this. Thanks for your contribution @gmnz but I'm afraid it doesn't fit with the goals of the project.

PR #337, opened 2023-06-08
Fixes #336 - Enabale treesitter indent for Python

### LINKED ISSUE #336: Python indenting now works in tree-sitter
From https://github.com/nvim-treesitter/nvim-treesitter/issues/2947, it seems that python indentation is finally working. See https://github.com/nvim-lua/kickstart.nvim/issues/78, and we no longer need 'disable = { "python" }'.

```
```

feoh:
Tested in my local configuratoin. Indenting works great.

@mech-a Hopefully I'm not abusing your good graces, but would you be willing to give this a shot? :) If you think it's OK I'll merge it.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -295,7 +295,7 @@ require('nvim-treesitter.configs').setup {
   auto_install = false,
 
   highlight = { enable = true },
-  indent = { enable = true, disable = { 'python' } },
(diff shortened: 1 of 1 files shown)


mech-a:
No worries, looks good to me, property indenting seems correct.

PR #342, opened 2023-06-14
Add legacy tag to fidget to avoid deprecation warning

gfaster:
Small fix - `fidget.nvim` is having a full rewrite so we need to check out the "legacy" tag there to avoid a deprecation warning

see: [fidget issue](https://github.com/j-hui/fidget.nvim/issues/131), and [relevant commit](https://github.com/j-hui/fidget.nvim/commit/c93c07c219bcf78c26b144027fcfd20cd23e55bd)

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -84,7 +84,7 @@ require('lazy').setup({
 
       -- Useful status updates for LSP
       -- NOTE: `opts = {}` is the same as calling `require('fidget').setup({})`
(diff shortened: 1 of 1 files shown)


feoh:
Thanks for this! Perhaps we should leave this issue open or file another one staged in the future to reevaluate once the rewrite is complete.

PR #344, opened 2023-06-17
Use call_parentheses = "None"

MariaSolOs:
`no_call_parentheses` has been deprecated: https://github.com/JohnnyMorganz/StyLua/blob/0d4decb462b2eb84d5e7ef1f8feba39c9890bdae/CHANGELOG.md?plain=1#L308-L311


feoh:
Good catch! Merging.

PR #346, opened 2023-06-18
Add descriptions for debugging keybindings.

juangiordana:
Add descriptions for debugging key bindings.

Improve formatting by changing double quotes to single quotes in order to keep compatibility with the rest of the kickstart.nvim project.


feoh:
Good catch.

PR #347, opened 2023-06-18
Fix typo in ignorecase comment

ilian:
In order to perform a case-sensitive search with ignorecase, the pattern should contain `\C` instead of `/C`.


feoh:
Checked the docs and you're right! Merging.

PR #357, opened 2023-07-01
Update README.md with a GitHub friendly video example

kdheepak:
I've added a .mp4 file that shows a copy paste example of using kickstart.nvim.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # kickstart.nvim
 
+https://github.com/kdheepak/kickstart.nvim/assets/1813121/f3ff9a2b-c31f-44df-a4fa-8a0d7b17cf7b
+
 ### Introduction
 
 A starting point for Neovim that is:


mech-a:
Please [see this comment.](https://github.com/nvim-lua/kickstart.nvim/pull/358#issuecomment-1616107362)

mech-a:
I think that we want users to clone the whole repository so that they have routes to add their own custom plugins from the get-go. While I think the change in #358 is sound because it brings us in line to the readme, adding more explicit support to this installation method isn't in line with what we want users to do.we intend for users to do. Kickstart was intentionally expanded with multi-file support, and we want new users to be able to see that.

kdheepak:
That makes sense. I’ll update the video to show an installation using a clone.

kdheepak:
I've pushed another `mp4` file that uses git clone instead.

feoh:
I'm struggling a bit evaluating this one because as a visually impaired person everything in the video is whizzing by at the speed of light.

Would love some more feedback on whether or not fully sighted folks would find this a good addition.

@mech-a , @tjdevries or anyone else?

Thanks much

kdheepak:
I’m also open to suggestions on better ways to visually depict what kickstart.nvim does.

feoh:
@kdheepak Let's keep our minds open and think on that, shall we? :)

Until then, I merged your PR. Thanks for the contribution!

I realized, what harm can it do? It can only help visual learners who aren't in the cyclops minority I inhabit :)

kdheepak:
Great! If anyone has ideas for improving it, feel free to post here and I’ll help in whatever way I can!

feoh:
I would personally like to see a slight pause between segments so I can process what's happening better, but my lacking perceptual capabilities shouldn't be a guide for your efforts :)

mech-a:
Thanks for the video; here are some potential improvements for round 2, in rough priority:

- Decrease compression to limit artifacts
- Decrease viewport of video
- Move into installation section; the main selling point of Kickstart is not its cloneability, but readable docs :)
- Record only the terminal window

kdheepak:
Btw, we are limited to 10mb on GitHub for mp4 files.  The original file without compression was 800mb. 

Maybe it should be a YouTube video?

mech-a:
I think it's best to make it an svg/gif. MP4s at that resolution are a bit too large. See [sharkdp/fd demo](https://github.com/sharkdp/fd/blob/master/README.md#demo) for a good example of an svg-based demo on a small viewport.

PR #358, opened 2023-07-01
Make init.lua copy-paste friendly

kdheepak:
Currently if you copy paste `init.lua` it gives an error because it tries to `require "custom.plugins"`. If we want the copy paste path to be a happy path, I believe this should be commented by default.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -197,12 +197,13 @@ require('lazy').setup({
   -- require 'kickstart.plugins.autoformat',
   -- require 'kickstart.plugins.debug',
 
-  -- NOTE: The import below automatically adds your own plugins, configuration, etc from `lua/custom/plugins/*.lua`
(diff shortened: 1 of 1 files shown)


mech-a:
We explicitly don't want people to copy and paste `init.lua` in -- that installation method is for a very outdated version of kickstart.

kdheepak:
While I generally agree, I do believe in this case it is a small tradeoff for getting a happy path for the copy paste use case, e.g. if I wanted to demo how easy it is to start using nvim. 

Also the readme mentions that this repo is a single file configuration with examples to move to a multiple file configuration.

mech-a:
Thanks for your comment - I agree that the PR brings us more in line with what we state on the readme. I think this is a good change and we won't need to adjust our installation instructions. @feoh I'm curious about your thoughts?

PR #362, opened 2023-07-07
fix: correct typo

andres-lowrie:
Hey y'all 👋 

This is super minor but I think there may have been a typo?


feoh:
Thanks for the fix!

PR #369, opened 2023-07-20
Add hint to uncomment line to autodetect plugins

Dennis-Rall:
I think you should mention that now you have to uncomment line 206 in init.lua to autodetect plugins

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Additional system requirements:
 * Inside of your copy, feel free to modify any file you like! It's your copy!
 * Feel free to change any of the default options in `init.lua` to better suit your needs.
 * For adding plugins, there are 3 primary options:
-  * Add new configuration in `lua/custom/plugins/*` files, which will be auto sourced using `lazy.nvim`
+  * Add new configuration in `lua/custom/plugins/*` files, which will be auto sourced using `lazy.nvim` (uncomment line 206 in init.lua therefore)
   * Modify `init.lua` with additional plugins.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,7 +55,7 @@ Additional system requirements:
 * Inside of your copy, feel free to modify any file you like! It's your copy!
 * Feel free to change any of the default options in `init.lua` to better suit your needs.
 * For adding plugins, there are 3 primary options:
-  * Add new configuration in `lua/custom/plugins/*` files, which will be auto sourced using `lazy.nvim`
+  * Add new configuration in `lua/custom/plugins/*` files, which will be auto sourced using `lazy.nvim` (uncomment line 206 in init.lua therefore)
```
kang8, line 58: Maintaining a specific line number can be challenging, as it requires constant monitoring of every change to `inin.lut` at line 206

feoh: Agreed. Please describe the line to be removed rather than citing a line number and I'll merge this.

Also please consider changing "therefore" to something like 'to enable this'.

Dennis-Rall: Yeah you're right, I changed the hint according to your suggestions

feoh (approved):
Thank you very much for your contribution and for working with us to make kickstart.nvim the best it can be :)

mech-a:
@feoh curious if you intended to merge this as well? LGTM & is approved

feoh:
Good catch! I think I thought I was approving emerging the whole pull request as opposed to one commit

PR #371, opened 2023-07-20
Update init.lua

componhead:
source custom plugins folder


mech-a:
This was an intentional change to comment it out -- see the above lines. We've also updated documentation for this in the README with a recent MR. Is there any specific reason you'd like this change?

componhead:
Sorry, I thought a mistake.

PR #373, opened 2023-07-24
Allow easier overriding of default filetypes for LSP

Numkil:
Hi,

For some language servers it makes sense to override the default filetypes that they attach too. For example when editing a twig file it is nice to have the html language server attached to the buffer. 

With this little addition to the setup_handlers function you can just put the filetypes configuration in the server configuration defined above and it will work. If you don't specify filetypes on your lsp config the value set will be nil and the default behavior will occur.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -424,12 +424,16 @@ end
 --
 --  Add any additional override configuration in the following tables. They will be passed to
 --  the `settings` field of the server config. You must look up that documentation yourself.
+--
+--  If you want to override the default filetypes that your language server will attach to you can
+--  define the property 'filetypes' to the map in question.
 local servers = {
   -- clangd = {},
   -- gopls = {},
   -- pyright = {},
   -- rust_analyzer = {},
   -- tsserver = {},
+  -- html = { filetypes = { 'html', 'twig', 'hbs'} },
 
   lua_ls = {
     Lua = {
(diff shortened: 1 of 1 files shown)


feoh:
This looks good to me, would you please add an example use case to the README under Configuration And Extension
?

Thanks a bunch!

init.lua
```diff
@@ -424,12 +424,16 @@ end
 --
 --  Add any additional override configuration in the following tables. They will be passed to
 --  the `settings` field of the server config. You must look up that documentation yourself.
+--
+--  If you want to override the default filetypes that your language server will attach too you can
```
feoh, line 428: Sorry to be a pest, but you want 'attach to' not 'attach to' otherwise looks great!

feoh:
Hi I'm sorry but rather than correcting the spelling error you removed the docs entirely.

Numkil:
Hi @feoh  Sorry its late in the evening for me and I accidentally rebased my PR on the wrong branch because i was not paying attention :) I fixed it now!

feoh:
No worries! Thanks for your patience and thanks for the contribution! I've actually got a couple use cases for this myself :)

feoh:
Hi.

Unfortunately after laboriously merging down your PR into my fork, I discovered it breaks if you don't do any of these overrides. I now get:

```
Error detected while processing /Users/cpatti/.config/nvim/init.lua:
/Users/cpatti/.config/nvim/init.lua:467: attempt to index a nil value
/Users/cpatti/.config/nvim/init.lua:467: attempt to index a nil value
/Users/cpatti/.config/nvim/init.lua:467: attempt to index a nil value
/Users/cpatti/.config/nvim/init.lua:467: attempt to index a nil value

```

On startup, which isn't great :(

Unless you make a new PR with a fix in a few hours, I'll revert your commit this evening.

Thanks and apoligies to anyone who gets this broken kickstart. My bad for not testing it thoroughly outside my fork.

PR #376, opened 2023-07-25
Fix bug when server not explicitely defined in configuration but installed

### LINKED ISSUE #375: kickstart.nvim currently spews 'attempt to index null value' errors at startup
## Describe the bug

I goofed and merged PR #373 without thoroughly testing it.

So now you get:

```
Error detected while processing /Users/cpatti/.config/nvim/init.lua:
/Users/cpatti/.config/nvim/init.lua:467: attempt to index a nil value
/Users/cpatti/.config/nvim/init.lua:467: attempt to index a nil value
/Users/cpatti/.config/nvim/init.lua:467: attempt to index a nil value
/Users/cpatti/.config/nvim/init.lua:467: attempt to index a nil value

```

On startup.

## To Reproduce

1. Pull latest kickstart.nvim from master
2. Start Neovim

## Desktop

- OS: Any
- Terminal: Any

## Neovim Version

```
NVIM v0.10.0-dev-1839+g0329f5c2f-Homebrew
Build type: Release
LuaJIT 2.1.0-beta3

  fall-back for $VIM: "/opt/homebrew/Cellar/neovim/HEAD-0329f5c/share/nvim"

```

Numkil:
hi @feoh 

The  problem was that kickstart initializes all LSP's that are installed and not only the ones explicitly defined in your servers variable. I locally had not done this so I did not realize this at the time. 

Here I add a conditional check to see if servers[server_name] is explicitly defined in servers and if not to just set nil on the filetypes variable

This fixes #375

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -463,7 +463,7 @@ mason_lspconfig.setup_handlers {
       capabilities = capabilities,
       on_attach = on_attach,
       settings = servers[server_name],
-      filetypes = servers[server_name].filetypes,
+      filetypes = (servers[server_name] or {}).filetypes,
     }
   end
 }


tjdevries:
I think you could do something like this:

`filetypes = servers[server_name].filetypes`, -> `filetypes = (servers[server_name] or {}).filetypes`,

feoh:
Abso-freaking-lutely!

One tip I wasn't aware of with forks:

If you're encountering all kinds of crazy pulling down changes from upstream and you're using rebase, try merge, it will do the git auto-resolve magic much more effectively.

I was having a hell of a time maintaining mine but now that I've switched from rebase to merge I only get dinged when there's an ACTUAL conflict.

If you like maybe put a pointer to your refactor here once you're done. Always worth keeping an open mind and seeing how others are doing things!

Thanks everyone for your comments and help with this. I feel like my Neovim Lua education is ongoing :)

PR #383, opened 2023-08-05
Update README.md

mgua:
added a pointer to my fork where instructions for different operating systems are available

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,3 +147,6 @@ This requires:
 {'nvim-telescope/telescope-fzf-native.nvim', build = 'cmake -S. -Bbuild -DCMAKE_BUILD_TYPE=Release && cmake --build build --config Release && cmake --install build --prefix build' }
 ```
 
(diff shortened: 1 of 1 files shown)


antoineco:
How about contributing changes upstream instead?

feoh:
I agree. Thank you for your contribution, but i'd be much more inclined to merge it if you would simply put those instructions in line.

gregorywaynepower:
@antoineco @feoh I can get these instructions in an inline change in the REAMDE.md if y'all want?

mech-a:
Thanks for your contribution, I'm not sure we really need this, as much of the details are about getting neovim up and running on some less-used OS's, which is better handled by neovim's own docs than us. For the other dependencies we have, it's also better to just reference their own docs.

feoh:
Hi thanks for your contribution but as others have said I don't think this is a desirable addition. Closing for now.

PR #385, opened 2023-08-06
little typo to show line numbers

fgau:


antoineco:
This is correct and intended: https://neovim.io/doc/user/lua-guide.html#_-vim.o

fgau:
> `vim.wo` is correct and intended: https://neovim.io/doc/user/lua-guide.html#_-vim.o
> 
> > `vim.wo`: for window-scoped options (can be double indexed)

your are absolutely right! will read the docu more!

PR #397, opened 2023-08-13
add `smart add to quickfix list` for telescope

gregorywaynepower:
Rebound default mapping of ['<C-q>'] to the telescope.action to use the smart adding of items to a quickfix list in the case users toggle matches of results without breaking the original functionality of all matches being added to a quickfix list.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -283,6 +283,10 @@ require('telescope').setup {
       i = {
         ['<C-u>'] = false,
         ['<C-d>'] = false,
+        ['<C-q>'] = require('telescope.actions').smart_add_to_qflist + require('telescope.actions').open_qflist
+      },
+      n = {
+        ['<C-q>'] = require('telescope.actions').smart_add_to_qflist + require('telescope.actions').open_qflist
       },
     },
   },


mech-a:
Thanks for your contribution! I'm hesitant to add this as I think that the quickfix feature is already not very used; I'm not too sure we need to add a default keybind for this, especially as downstream users could have bound C-q to something already. Could you provide some more details on why you think this would be a good default addition?

feoh:
I don't think I understand the use case or why this is desirable enough to test it properly, so, we'd need:

- Someone who understands why we'd want this
- Is able to put that understanding to work in validating the correctness of the PR and that it won't unlease a barrage of others yelling about it overriding their particular customizations :)

mech-a:
Doing some searching around, I feel like the quickfix list, though useful, is quite niche and relevant in only some workflows (mostly C). I feel like this addition is out of scope for the default options as it adds a good deal of complexity that new users would have to figure out -- the quickfix list is not the easiest feature to understand, and we try to keep the config such that a new user could read just the file and understand every line. I appreciate the submission, but I'd prefer it to be something users set up and not shipped in the repo.

feoh:
@mech-a Thank you for your assistance as always. I have REALLY appreciated your help with this project!

@gregorywaynepower Thanks for your contribution but I'm closing this for now. Please feel free to make this change in your local fork if you haven't already!

gregorywaynepower:
No worries, I appreciate the feedback!

PR #406, opened 2023-08-22
Revert gitsigns keymaps but fix vimdiff and fugitive conflict

gangelop:
Originally, the keymaps for jumping to next and previous git hunks were `]c` and `[c`. This was changed in #323 (83b65a1) because they overwrote the built-in vimdiff keymaps.

However, the more traditional solution is to have `]c` and `[c` *extend* the built-in keymap. This is what fugitive and gitgutter have been doing for years.

Gitsigns doesn't do this by itself, but it has a recommended keymap configuration on which the present patch is based:

https://github.com/lewis6991/gitsigns.nvim#keymaps

The only thing I've added is to have the keymaps work in visual mode as well, which is the same behavior as the built in vimdiff keymaps.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -124,9 +124,20 @@ require('lazy').setup({
         changedelete = { text = '~' },
       },
       on_attach = function(bufnr)
-        vim.keymap.set('n', '<leader>gp', require('gitsigns').prev_hunk, { buffer = bufnr, desc = '[G]o to [P]revious Hunk' })
-        vim.keymap.set('n', '<leader>gn', require('gitsigns').next_hunk, { buffer = bufnr, desc = '[G]o to [N]ext Hunk' })
-        vim.keymap.set('n', '<leader>ph', require('gitsigns').preview_hunk, { buffer = bufnr, desc = '[P]review [H]unk' })
+        vim.keymap.set('n', '<leader>hp', require('gitsigns').preview_hunk, { buffer = bufnr, desc = 'Preview git hunk' })
+
+        -- don't override the built-in and fugitive keymaps
+        local gs = package.loaded.gitsigns
+        vim.keymap.set({'n', 'v'}, ']c', function()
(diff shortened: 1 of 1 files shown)


feoh:
I haven't really delved into gitsign enough so I can't comment on this.

@antoineco seems like you approve this change? Have you tested it?

Will I get a raft of "AAH YOU MOVED MY CHEESE!" angry comments if I merge this? :)

feoh:
Can I ask you to do some homework so we can move this PR along if in fact it makes sense to merge?

Please go back and find the PRs from the folks who made the change you disagree with. Please at least try @-ing them to see if you can come to consensus on a solution.

If you can, it's an easy merge from my end. If you can't, then we'll keep working on this. I really do appreciate the contribution!

gangelop:
Hi @luizcalliari 👋

In #323 (83b65a1) you changed the `]c` and `[c` mappings for gitsigns because they conflicted with vimdiff.
I'm trying to bring them back in a way that doesn't conflict with vimdiff.
The expected result is that `]c` and `[c` should take you to next and previous hunk when there are git differences but when you open the file in a vimdiff, the same keys should take you to the next and previous difference in the diff.

Would you mind testing the patch in the current PR to confirm that it's still accomplishes what you intended with your earlier change?

antoineco:
@feoh sorry I completely missed your mention. Yes I do think that the change is sensible, and it works as expected for me 👍

feoh:
The more I look at this, the more I think it's a good idea. @luizcalliari -f you respond fairly soonish with any objections we can discuss it, otherwise I think I'm gonna merge this.

Nothing is set in stone, this is Git after all :)

luizcalliari:
Hello guys!! I've no objection!! Good job @gangelop!!!

luizcalliari (approved):

feoh:
Thank you everyone! Merged.

PR #423, opened 2023-09-08
Allow the user to set the default autoformat setting

Anthony-Fiddes:
Hello!

I made this change to my autoformat config and just wanted to share it in case it's useful. I don't know if this is the best way to make the default autoformat setting user configurable, but it works for me so far.

### DIFF
--- a/lua/kickstart/plugins/autoformat.lua
+++ b/lua/kickstart/plugins/autoformat.lua
@@ -8,11 +8,16 @@ return {
   config = function()
     -- Switch for controlling whether you want autoformatting.
     --  Use :KickstartFormatToggle to toggle autoformatting on or off
-    local format_is_enabled = true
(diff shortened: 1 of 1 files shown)


feoh:
Can you help me understand the use case here? What problem is this trying to solve?

Anthony-Fiddes:
I just wanted to add `vim.g.kickstart_autoformat_enabled = false` to my work config so that all buffers don't autoformat by default there. I wasn't sure how niche a use case this might be (to want autoformat enabled by default in one environment but not in another). Feel free to close the PR if you feel like it's not useful!

feoh:
Hi, yes I'm afraid I don't see a general use case that would appeal to all or most neovim users here. I'm going to close this for now. If someone from the community has strong feelings, please re-open and chime in.

Thanks for your contribution!

PR #438, opened 2023-09-24
Which key documentation and 2-3 changed keybindings

tcld:
I wanted the new keybindings to be properly documented, since `which-key` has been a great help in learning and exploring vim-commands I had been previously unaware of.

While doing that, I noticed that a few of the existing commands might be better of grouped in with others.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -438,6 +438,17 @@ local on_attach = function(_, bufnr)
   end, { desc = 'Format current buffer with LSP' })
 end
 
+-- document existing key chains
+require('which-key').register({
+  ['<leader>c'] = { name = '[C]ode', _ = 'which_key_ignore' },
+  ['<leader>d'] = { name = '[D]ocument', _ = 'which_key_ignore' },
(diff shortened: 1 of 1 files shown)


feoh:
@tcld Thanks very much for your contribution!

These look good to me, but gitsign keybind changes have been V-E-R-Y controversial in the past.

If you wouldn't mind, please take a look through the relevant closed PRs and at the folks involved to see if they have any objection. If not I'd be happy to merge.

feoh:
@tcld Alternatively if you just want to improve the docs and strip out the keybinding changes I can merge it immediately.

tcld:
I'll do the latter on Friday or Saturday; don't have a PC at the moment.
Any suggestion for the doc string I'll need for the gitsigns "group" with a
single element?

tcld:
I removed the last commit, so only the changes to the which-key docs are left. If I find the time, I'll have a look at related PRs to see how people felt about changing keybindings. :)

feoh:
Thanks very much! Sorry to be a pain about this it just gets hard trying to serve everyone when people have Strong Feelings about a given set of keybindings :)

PR #443, opened 2023-09-28
fix(init.lua): indent blankline v3 setup

emmanuelchucks:
Indent blankline plugin v3 setup throws error when called with `opts = {...}`

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -170,10 +170,12 @@ require('lazy').setup({
     'lukas-reineke/indent-blankline.nvim',
     -- Enable `lukas-reineke/indent-blankline.nvim`
     -- See `:help indent_blankline.txt`
-    opts = {
-      char = '┊',
-      show_trailing_blankline_indent = false,
-    },
+    config = function()
+      require('ibl').setup {
+        char = '┊',
+        show_trailing_blankline_indent = false,
+      }
+    end,
   },
 
   -- "gc" to comment visual regions/lines


init.lua
```diff
@@ -170,10 +170,12 @@ require('lazy').setup({
...
-    opts = {
-      char = '┊',
-      show_trailing_blankline_indent = false,
-    },
+    config = function()
+      require('ibl').setup {
+        char = '┊',
+        show_trailing_blankline_indent = false,
+      }
+    end,
```
tranzystorek-io, line 178: This can be done easier by setting `main = 'ibl'` in the plugin config. Also it seems `show_trailing_blankline_indent = false` is the default now.

confusingbits: The `main = "ilb"` solution (it's in the v3 migration docs) seems to be working for me, the conig function solution didn't work for me. 🤷‍♂️ 

image

benjaminbauer: also the `main = 'ibl'` is the way how [upstream](https://github.com/lukas-reineke/indent-blankline.nvim/wiki/Migrate-to-version-3) recommends it

feoh:
I see the error too. I'll be in work meetings all morning - if someone other than the OP has time to test this fix and validate that it fixes the issue, I'd be happy to merge.

feoh:
Actually, I changed my mind :) We can always revert and we're busted right now so there's no where to go but up. Merging.

PR #445, opened 2023-09-28
Update init.lua

AaryanTrivedi:
To avoid the error "You are trying to call the setup function of Indent-blankline version 2 but you have version 3 installed."


lukas-reineke:
#443 already updated to v3, this is not necessary

PR #459, opened 2023-10-09
Nvterm test

alexvoelker:


alexvoelker:
added terminal functionality in the nvim editor window

PR #462, opened 2023-10-10
Fix help for indent-blankline plugin

Alvaro-Kothe:
Since version 3 `:help indent_blankline` no longer works. Replace it with `:help ibl`.


feoh:
Thanks for the fix! Merged.

PR #465, opened 2023-10-12
disabling ibl scope by default

arthurbailao:
I don't know what you think, but I personally don't like this kind of scope highlighting. And worst, I spent quite some time to find out which plugin introduced this feature. For me is not intuitive that an indentation blank line plugin could also highlight scopes, so maybe we should disabled this feature by default.

image

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -179,7 +179,9 @@ require('lazy').setup({
     -- Enable `lukas-reineke/indent-blankline.nvim`
     -- See `:help indent_blankline.txt`
     main = 'ibl',
-    opts = {},
+    opts = {
+      scope = { enabled = false },
+    },
   },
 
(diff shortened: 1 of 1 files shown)


feoh:
Hi.

Thank you for your contribution, but we would encourage you to make this change in a personal fork.

If others in the community also feel strongly about this change, please speak up, but for now I'm closing this.

We have too many users with too many disparate use cases and needs to try to accommodate everyone when you can just fork as I do and have the best of all possible worlds! :)

(I can't personally use the onedark theme kickstart uses by default, so I maintain https://github.com/feoh/kickstart.nvim )

arthurbailao:
For sure, I do have my own fork for quite some time too. I sent this PR because it was an annoying experience, just trying to help others.

PR #467, opened 2023-10-13
Use Telescope builtin functions for LSP definition and type definition

theopn:
I thought it was a little odd that every other LSP features use Telescope builtin but not those two. It will allow easier navigation when there are multiple symbol definitions.

We can also potentially do something like the following snippet to be safer and provide an example of `pcall`:

```lua
  -- ...
  local status, builtin = pcall(require, 'telescope.builtin')
  if status then
    nmap('gd', builtin.lsp_definitions, '[G]oto [D]efinition')
    -- ...
  else
    nmap('gd', vim.lsp.buf.definition, '[G]oto [D]efinition')
    -- ...
  end
```

Although I am not sure it will be necessary since Telescope is very widely used in Kickstart and it might clutter the config and confuse beginners.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -422,10 +422,10 @@ local on_attach = function(_, bufnr)
   nmap('<leader>rn', vim.lsp.buf.rename, '[R]e[n]ame')
   nmap('<leader>ca', vim.lsp.buf.code_action, '[C]ode [A]ction')
 
-  nmap('gd', vim.lsp.buf.definition, '[G]oto [D]efinition')
+  nmap('gd', require('telescope.builtin').lsp_definitions, '[G]oto [D]efinition')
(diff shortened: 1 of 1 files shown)


maxwelbm (approved):
Thanks for the suggestion, I need to improve my configuration

maxwelbm (approved):

one2three4f:
I thought it was strange the gd key wasn't working with this lsp https://github.com/Feel-ix-343/markdown-oxide , for some reason telescope's gd does not behave the same way vim's. 

took a peek and lunar vim also keeps native gd.
https://github.com/LunarVim/LunarVim/blob/9ee3b7b8846d7ed2fa79f03d67083f8b95c897f2/lua/lvim/lsp/config.lua#L88

PR #472, opened 2023-10-20
Added simple C++ compiler for Windows information

razoreater:


feoh:
Thank you for your contribution, but I have concerns about this.

You're assuming that:

1) People are allowed to use/install chocolatey
2) People can install mingw on their Windows system without adversely affecting other systems they may have in play.

Happy to see this re-opened if there are large numbers of Windows users who can guarantee that both points above are satisfied, but for now I'm closing this.

PR #473, opened 2023-10-22
reorganize init.lua to a multi file setup

dam9000:
I understand the kickstart init.lua is single-file by design, and most likely this PR will not be accepted and that's fine. I just think a multi-file approach is better, even for beginners. In any case thanks for the kickstart.nvim it is the best starting point for nvim configuration.


feoh:
Hi there!

First off, thanks for your contribution! The beauty of open source is that there's no harm in submitting a PR.

As you say, the nature of this particular project is that it's largely a single file. One file to read. One file to understand.

Would you consider starting a fork? That way you could maintain whatever organization makes sense for you, and folks can use your fork instead who want a multiple file approach.

As always I welcome community input. If folks feel very strongly that a multiple file approach makes more sense, then please chime in here. If enough people feel this way we can re-open and reconsider this PR, but I'm closing it for now.

dam9000:
@feoh thanks for the feedback.
Regarding the fork, yes, my multi-file kickstart fork is here:
https://github.com/dam9000/kickstart-modular.nvim

WindSoilder:
> As you say, the nature of this particular project is that it's largely a single file. One file to read. One file to understand

This is what makes me find it usable, I've tried to use neovim many times, but only this one helps me dive into the world of neovim.
It's a single file, it's best for new-bee who doesn't have any neovim experience.

feoh:
@dam9000 That's fantastic! With your approval I'll add a note to the README for folks who might prefer a multi-file approach so they can more easily find your project.

dam9000:
@feoh Sure, of course I approve and welcome if you can add a link to my multi-file fork to the README. And I'll try to do my best to keep it up to date with all the upstream changes, while making no other functional changes.

greg-nagy:
I know that maintaining oss projects is hard 

so I am here to thank you all for both the single file kickstart, the modular kickstart and the starter

as someone new to neovim I find all of them very helpful and a good augmentation for my learning as I get into the weeds

appreciate y'all

drvnprgrmr:
This kickstart finally kickstarted my journey into neovim fully. I'd started and stopped so many times, but I'm going fully in now -- I even deleted VSCode to force myself -- and its been going pretty well.

drvnprgrmr:
Just need to modularize the init file now so that I don't have to keep quitting neovim every time I make a change.

szechp:
Can we re-open this for further discussion?

Context here: https://github.com/nvim-lua/kickstart.nvim/pull/1481

Maybe we can re-evaluate this a bit? Maybe have the bare minimum in one file, but do stuff like telescope or snacks into their separate files?

AlexanderGolys:
I think it's useful to make more explicit the main components you should check out next as a n00b and set the path to 'typical complete experience'.
TJs masterclass for me was the perfection and without them I think I would be completely lost in the tsunami of features "to check out", it's hard to not get stuck on side quests even with the main quest bolded and with great voice actor

Issue #478, opened 2023-10-24
Post Installation… huh?


michaelGregoire:
Please simply explain in the README what the "Post Installation" command `nvim --headless "+Lazy! sync" +qa` is actually doing and why it's needed. **kickstart.nvim** is seemingly geared toward making it easier for folks to approach neovim, so please explain things to us like we’re 5. We want to learn. K thanx bye.

dam9000:
That is supposed to install the configured plugins.
Here is what the individual options do:
  `--headless`: run without a UI, just execute the provided commands
  `"+Lazy! sync"`: Run install, clean and update the plugins
  `+qa`: quit

However, the funny thing is, Lazy sync is run automatically on the first run by the Lazy manager anyway. If you just run nvim without arguments on a fresh install you will see the Lazy plugin manager doing it's thing - installing and synchronizing the plugins.

In addition the `"+Lazy! sync"` is not even recognized as a valid command, perhaps because the Lazy plugin manager was not yet loaded at the time the nvim tries to execute the command line commands, perhaps it executes them before loading the init.lua.

Here is what you get without the `--headless` and `+qa` commands, that will allow you to see the output:

```
Error detected while processing command line:
E492: Not an editor command: Lazy! sync
Press ENTER or type command to continue
```

So it seems the instruction should actually be to just run:
`nvim --headless +qa`
Which would perform the first-time install of plugins.

Or perhaps better, just run nvim without arguments, so that you can see what is actually going on. With the readme saying that on the first run the plugins will be installed, perhaps that would be more instructional.

michaelGregoire:
Thanks for the detailed response. Much appreciated! That explains why the couple videos I've seen only say that you need to grab the `init.lua` to install.

feoh:
@michaelGregoire That's a different problem :( Those videos are out of date and wrong. I won't ask the authors to update them because it's a lot of work but I would love it if they did :)

If anyone wants to generate a PR to update the README.md removing the options so foks just get to invoke Neovim normally and watch the update/install, I'm totally fine with that and would gladly merge it.

dam9000:
For completeness, as I was wondering why without headless sync didn't work here is the answer, for the interested:
https://github.com/folke/lazy.nvim/issues/285#issuecomment-1568456990

dam9000:
Anyway here is a PR with a change to the README:
https://github.com/nvim-lua/kickstart.nvim/pull/479

[...]

PR #479, opened 2023-10-25
Update README Post Installation steps

dam9000:
As discussed here:
https://github.com/nvim-lua/kickstart.nvim/issues/478

Change the recommendation to just run nvim normally instead of the headless mode for the first run. This will show Lazy UI updating the plugins which would match what the video show and may be easier to understand what is going on thant the silent headless run.

Perhaps now it is too verbose? Let me know if you would prefer to word it differently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,15 @@ git clone https://github.com/nvim-lua/kickstart.nvim.git %userprofile%\AppData\L
 
 ### Post Installation
 
-Run the following command and then **you are ready to go**!
+Start Neovim
+
+```sh
+nvim
+```
+
+The `Lazy` plugin manager will start automatically on the first run and install the configured plugins - as can be seen in the introduction video. After the installation is complete you can press `q` to close the `Lazy` UI and **you are ready to go**! Next time you run nvim `Lazy` will no longer show up.
+
+If you would prefer to hide this step and run the plugin sync from the command line, you can use:
 
 ```sh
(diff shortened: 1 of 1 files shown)


lemusthelroy (approved):

README.md
```diff
@@ -51,7 +51,13 @@ git clone https://github.com/nvim-lua/kickstart.nvim.git %userprofile%\AppData\L
...
 ### Post Installation
 
-Run the following command and then **you are ready to go**!
+Start Neovim
+
+```sh
+nvim
+```
+
+The `Lazy` plugin manager will start automatically on the first run and install the configured plugins - as can be seen in the introduction video. After the installation is complete you can press `q` to close the `Lazy` UI and **you are ready to go**! Next time you run nvim `Lazy` will no longer show up. If you would prefer to hide this step and run the plugin sync from the command line, you can use:
```
ro0gr, line 60: nit. Maybe it's just me, but I'd keep the
> If you would prefer...

separated by a newline. Cause the preceding message seems to be complete.

feoh: @dam9000 I agree with @ro0gr.  Would it be possible to add the newline they suggest? I'd sincerely appreciate it and THANK YOU for your contribution in any case!

dam9000: Done.

feoh:
Thanks very much. This is a great contribution. Small improvements like this are super important!

PR #480, opened 2023-10-28
README.md: add a note about NVIM_APPNAME in the FAQ section

dam9000:
Not sure if this is really a frequently asked question, but it was a question I had myself and eventually found the answer elsewhere. Just thought it might be useful to point out the possibility of keeping multiple Neovim configurations in parallel also here.


feoh:
I like that you added it as an FAQ question so as not to complicate the main flow of the README. 

Merging.

PR #486, opened 2023-11-03
README.md: Added a FAQ: why is init.lua a single file

dam9000:
In the PR comment: https://github.com/nvim-lua/kickstart.nvim/pull/473#issuecomment-1774303260
It was suggested that a note would be added to the README regarding the single vs multi file approach of init.lua,
So here is a PR that adds a section in FAQ about the single file vs multi-file approach and a link to the multi-file fork.
Not sure if I formulated it the best, if you have some additional suggestions let me know.
(or close if you feel this does not belong here)

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,16 @@ Each PR, especially those which increase the line count, should have a descripti
 * Are there any cool videos about this plugin?
   * Current iteration of kickstart (coming soon)
   * Here is one about the previous iteration of kickstart: [video introduction to Kickstart.nvim](https://youtu.be/stqUbv-5u2s). Note the install via init.lua no longer works as specified. Please follow the install instructions in this file instead as they're up to date.
+* Why is the kickstart `init.lua` a single file? Wouldn't it make sense to split it into multiple files?
+  * The main purpose of kickstart is to serve as a teaching tool and a reference
+    configuration that someone can easily `git clone` as a basis for their own.
+    As you progress in learning Neovim and Lua, you might consider splitting `init.lua`
+    into smaller parts. A fork of kickstart that does this while maintaining the exact
+    same functionality is available here:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -161,6 +161,12 @@ Each PR, especially those which increase the line count, should have a descripti
 * Are there any cool videos about this plugin?
   * Current iteration of kickstart (coming soon)
   * Here is one about the previous iteration of kickstart: [video introduction to Kickstart.nvim](https://youtu.be/stqUbv-5u2s). Note the install via init.lua no longer works as specified. Please follow the install instructions in this file instead as they're up to date.
+* Why is the kickstart `init.lua` a single file? Wouldn't it make sense to split it into multiple files?
+  * The main purpose of kickstart is to serve as a teaching tool and a reference configuration that someone can easily copy and paste as a basis for their own. As you progress in learning Neovim and Lua, you might consider splitting `init.lua` into smaller parts. A fork of kickstart that does this while maintaining the exact same functionality is available here:
```
feoh, line 165: Firstly this *definitely* belongs here in my book, as this question keeps coming up, and I think yours is a thoughtful treatment and suggested answer that solves the problem for those who care.

I have just a couple tiny suggestions:

* Please consider replacing "copy and paste" with "git clone" because copying and pasting is no longer the write way to install kickstart and we get SOOO MANY PEOPLE coming here from @tjdevries video confused because things are broken I don't want to encourage them :)
* Please consider splitting that very long bullet into separate lines. In Markdown you can make the lines separate and shorter and they'll still render as one line unless you put two empty lines in between.

Thanks much for contributing, I think this will help a lot!

dam9000: Done, force pushed the suggested changes.

feoh (approved):

PR #492, opened 2023-11-06
add Telescope live_grep on Git root

slara:
Adds a keymap to use Live Search with Grep on the entire current git project.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -308,6 +308,40 @@ require('telescope').setup {
 -- Enable telescope fzf native, if installed
 pcall(require('telescope').load_extension, 'fzf')
 
+-- Telescope live_grep in git root
+-- Function to find the git root directory based on the current buffer's path
+local function find_git_root()
+  -- Use the current buffer's path as the starting point for the git search
+  local current_file = vim.api.nvim_buf_get_name(0)
+  -- If the buffer is not associated with a file, return nil
+  if current_file == "" then
+    print("Buffer is not associated with a file")
+    return nil
+  end
+  -- Extract the directory from the current file's path
+  local current_dir = vim.fn.fnamemodify(current_file, ":h")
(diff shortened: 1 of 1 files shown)


feoh:
This looks neat! Thank you for your contribution.

Has it been tested on Windows?

If not and you don't have access to a Windows install I can test it in my fork.

slara:
Hi @feoh , Thanks! I tested it on Linux and Mac. I don't have a Windows install available.

dam9000:
Can we keep the old `<leader>sg` keymap as it was (live grep) and use your newly added new `<leader>sG` for the new functionality (grep git root), this would be less disruptive for existing users.

feoh:
I love this idea!

dam9000:
Hmm, another thing to consider, perhaps this functionality should be made into a plugin that can then be added to kickstart as a single line addition, if we want to keep the "kickstart": small, single file, and easy to learn. Wouldn't that be a better approach?

feoh:
@dam9000 Gonna pump the brakes on that one :) I don't want to get kickstart.nvim into the business of needing to support plugins meant to extend *it* rather than Neovim itself.

I don't personally think this code change is so massive, I think it's a decision of include it or don't. I'm personally leaning towards including it but I'm open to other feedback.

feoh:
I don't really understand how to use the :Git stuff yet, and when I try this here on Windows I get "Buffer not associated with file" so I'd guess that's working as designed?

Man I wish I had the time to delve more deeply into this stuff :)

I'm going to merge and if any ninjas out there understand enough to tell me it's wrong we can work on it from there :)

slara:
Thanks @feoh! This is so cool! 

I think this could happen if you are inside a special buffer like NeoTree or the dashboard. I will try to reproduce it and make a fix.

PR #494, opened 2023-11-07
Live Grep from Git root falls back to cwd on special buffers

slara:
This also falls back to current working dir if dir is not in a git repository.


feoh:
That's fantastic thank you very much for your contributions!

slara:
Thank you @feoh !

PR #498, opened 2023-11-09
4 live grep not working

thewordisbird:
This only adds back code to allow for git root grep. The solution to the issue was to add the `ripgrep` dependency via brew:

```
brew install ripgrep
```


juangiordana:
You should push to your own fork :)

feoh:
Hi. This PR seems like a mistake. I'm closing it for now.

PR #505, opened 2023-11-13
init.lua: remove 'legacy' tag from fidget.nvim

j-hui:
Since https://github.com/j-hui/fidget.nvim/pull/143, the `legacy` tag is no longer necessary (or recommended) for my plugin.

(Then again, I'm not sure if the point of including fidget.nvim here was to be instructive about how Lazy `tag`s work, so maybe this PR defeats the point?)

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -88,7 +88,7 @@ require('lazy').setup({
 
       -- Useful status updates for LSP
       -- NOTE: `opts = {}` is the same as calling `require('fidget').setup({})`
-      { 'j-hui/fidget.nvim', tag = 'legacy', opts = {} },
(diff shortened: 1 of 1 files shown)


feoh:
I thought fidget was included because we wanted its functionality? Happy to merge this in any case, but if you think fidget should be removed from kickstart please open a separate issue and we can discuss there.

j-hui:
What I meant was, I wasn't sure if fidget.nvim was included for the _sole_ purpose of demonstrating how Lazy's `tag` mechanism worked---in which case, this PR would be counter-productive.

But that doesn't seem to be case. I'm very happy to have Fidget included in kickstart (: Thanks for merging!

PR #511, opened 2023-11-22
option to reduce noisy Lua_LS's missing-fields warnings

wommy:


dam9000:
wommy, I would recommend squashing the two commits into one since one is fixing the other and making a force push

feoh:
I agree that would be fantastic.

feoh:
I just squash merged it myself. Thanks for the contribution!

PR #524, opened 2023-11-30
Use alias instead of a wrapper script

wadkar:
It is much easier to stick an alias statement in `~/.{z,ba}shrc` than create a script, invoke another instance of interpreter and then run neovim


dam9000 (approved):

feoh:
Good call thanks for the contribution! And thanks for helping out @dam9000 I've been super busy with work stuff and getting COVID (mostly over it now thank goodness).

PR #525, opened 2023-12-02
Run stylua

dam9000:
Run `stylua` on `init.lua` to fix formatting of recent changes.

Not sure if this is something that the project cares about or not.


feoh:
Would it make sense to have this as a github action or something?

feoh:
More the second one.

The first is a pre-commit hook, which is great, but must be installed by each developer, and my point was making it part of the regular Git side workflow so we don't have to train devs up on using it but we get the win of properly formatted code anyway :)

dam9000:
Hmm I played around with github actions/stylua on my private repo and I could get it to run a stylua check on a PR, which shows the stylua errors, but does not perform the actual changes and still allows to commit the pr as is. I guess that would be a good start, right? I'll open a new PR for that.

dam9000:
force pushed: rebase to latest master to trigger the newly added action for stylua check

dam9000:
This PR passes the stylua check, it's good to merge, so that future PR will start from a clean state.

PR #549, opened 2023-12-20
Switch cmp up mapping from C-d to C-b to match regular vim up key

aripollak:
The existing keybinding seems like a typo, since C-d in vim normally scrolls downwards, but C-b scrolls upwards, which seems like the more accurate opposite of C-f.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -624,7 +624,7 @@ cmp.setup {
   mapping = cmp.mapping.preset.insert {
     ['<C-n>'] = cmp.mapping.select_next_item(),
     ['<C-p>'] = cmp.mapping.select_prev_item(),
-    ['<C-d>'] = cmp.mapping.scroll_docs(-4),
+    ['<C-b>'] = cmp.mapping.scroll_docs(-4),
     ['<C-f>'] = cmp.mapping.scroll_docs(4),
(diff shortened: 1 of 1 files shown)


feoh:
Hi. If you wish to see this merged, please do the following:

* Look through git history/git blame and find whoever checked in the key mapping you want to change.
* @ that person in a comment here and ask them whether or not they agree with this change.

Otherwise, if I see a number of folks in the community voicing opinions one way or another on this I'm happy to act accordingly.

I'm just concerned about unilateral changes to key mappings which might result in a thousand silent users having their workflow blow up.

Thanks for understanding and thanks for your contribution!

aripollak:
Welp, looks like this keybinding has been around since kickstart.nvim was started. I'm guessing it was originally based on nvim-cmp's README, which was eventually fixed in https://github.com/hrsh7th/nvim-cmp/pull/538.
@tjdevries, do you agree with this change?

feoh:
Hey, thanks much for doing that detective work!

If it's been fixed in cmp upstream, I'm WAY more tempted to just merge it.

Let's give it a few hours in case anyone has any vehement objections and I'll merge this.

Thanks!

dam9000:
I agree with the proposed change since it matches with cmp readme.

PR #550, opened 2023-12-21
Add CI support via Github Actions

hyiltiz:
I noticed that the master HEAD is causing multiple errors with the latest nvim (nvim, LuaSnip and vimdoc lsp). It would be useful to have a test before merging any changes to the repo.

Please remember to update README.md badge URL to point to @nvim-lua instead once merged.


dam9000:
@hyiltiz what kind of errors, I have not noticed any and i have the latest release NVIM v0.9.4. Are these errors present on the pre-release versions? 0.10.0-dev?

Also, I would recommend to squash the 14 commits to one or just a couple if it that makes more sense.

hyiltiz:
I can squash the commits as one. It is intended to install the last release of nvim (should not throw any errors) then clone master of `nvim-lua/kickstart.nvim` and start `nvim` a few times, to test whether kickstart has any issues.

dam9000:
Not opposed to this, but just thought to mention,  my understanding is that the reported errors were because of various included plugins breaking changes (in case of luasnip) not because of bad changes in kickstart. From my observation feoh does a pretty good job making sure that what gets merged does not cause any issues. Perhaps automating that can help, but it won't prevent errors that are in the included plugins.

feoh:
Hi.

I have concerns about this.

Neovim is a complex beast, and there are may variables around why or why not it might start cleanly.

First, it's actually a cool bit of work, so I appreciate your contribution on that count.

However: What is the new contributor experience like if your test fails because Neovim fails to start? I'm super worried about some upstream plugin that has nothing to do with their contribution failing to initialize and thus failing your tests.

I am not personally wild about merging this, but if others have very strong feelings to the contrary, please do speak up and re-open this.

hyiltiz:
I doubt that released `nvim` binaries will have issues starting up in `ubuntu-latest`. The Github Actions included for this repo currently only checks for lua styles and linting, but it would be useful to be able to know:
- over time, some plugins did not introduce a breaking change (if so, we'd be able to see it in builds and figure out a response; at the very least, a temporary pin on versions)
- any new changes introduced does not introduce conflicts with what is already in there

What do you think of the *idea* of introducing a PR check that tests a PR with `nvim`? If the idea itself doesn't have much traction, I'd be happy to close it off. If you are open to the idea but do not like the specific way this test is set up, we can certainly discuss on how it should be changed.

feoh:
I'm not averse to it but I am struggling to envision a set of tests that actually make sense.

Kickstart has almost no code of its own. It is a collection of snippets from READMEs of all its component plugins.

So, I don't understand how you test that.

But please, if you have ideas, I'm open to them.

hyiltiz:
I think we'd be testing whether the set of version of plugins installed would be compatible with each other (to the point that nothing throws errors when opening a buffer). As such, when a build fails:
- it is a signal for users to not install the latest version(s) of offending plugins
- it is a signal for kickstart devs/maintainers to pin the versions of the offending plugins until they are fixed
- it is a signal for kickstart devs to maybe file a issue with the upstream plugins

hyiltiz:
Could you please take a look at #553?

feoh:
Hi. I've looked. The problem is that there's some repo configuration I don't think I can change that's requiring that every Github Action script be "approved by an administrator" *with every check-in*.

This seems not great, and I'm not sure what to do about it.

hyiltiz:
I *think* that is because my PR, owned by @hyiltiz, is attempting to trigger a Github Action for this repo under @nvim-lua, which would use up some of the Action rates. Once it is merged into this repo (`master` or elsewhere, so we can merge into `dev` to check if you like), I don't think it would require an approval. If it is not automatic on merge, we might need a repo admin to give perms once.

Is there anyone knowledgeable on this subject, or maybe a wiki page?

hyiltiz:
Also, please ignore #550 in preference of #553.

PR #551, opened 2023-12-21
reenable indentation guide

nealchau:
from v2 to v3 the character was moved into the indent dict


dam9000:
I don't see what the problem is, is this your local change and you made the PR by mistake?
With the latest ibl, I see the indent lines just fine
image
and ibl is the latest v3.4:
image

nealchau:
ok maybe was version related.  thanks

PR #569, opened 2024-01-03
Harpoon should be in Kickstart

jayzone91:
Added Harpoon by ThePrimeagean.
Added a config, keymaps and telescope integration.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -253,6 +253,55 @@ require('lazy').setup({
     },
     build = ':TSUpdate',
   },
+  {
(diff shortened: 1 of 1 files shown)


feoh:
Hi. Respectfully I strongly disagree with this change. 

Harpoon is great, and I use it myself, but It requires a great deal of customization to be truly useful and is not a tool suitable for newcomers to Neovim IMO.

Thanks for your contribution but I'm closing this for now. If there is massive, strenuous opposition we can reconsider it.

PR #590, opened 2024-01-20
Add Onedark Style

Tokarak:
I wanted to change the style in my own config, and thought that it wasn't as trivial as I would like it to be.

Default behaviour unchanged. The option to change the Onedark theme are now more exposed to the user.

Lazy-loading defaults to off as recommended by Lazy.nvim for the main theme.

Is the commenting satisfactory on all counts?

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -193,7 +193,11 @@ require('lazy').setup({
     -- Theme inspired by Atom
     'navarasu/onedark.nvim',
     priority = 1000,
+    lazy = false,
     config = function()
+      require('onedark').setup({
+        style = 'dark' -- dark, darker, cool, deep, warm, warmer, light
+      })
       vim.cmd.colorscheme 'onedark'
     end,
   },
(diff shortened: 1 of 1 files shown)


feoh:
Hi ! I like the idea of this change.

Could you please:

1) Run stylua on your code to make that check pass.
2) Add some documentation into the README showing how one might configure their chosen style?

Thanks for your contribution!

Triavanicus:
So this would've helped me as I switched to 3 different themes, looking to get a darker theme, only to refactor my config a few weeks later, to noticed I had come full circle, and just modified the original theme (which I was unaware I could do). I ended up doing the following, with the theme.

```lua
  {
    'navarasu/onedark.nvim',
    lazy = false,
    priority = 1000,
    opts = {
        style = 'darker',
        colors = {
          bg0 = "#1d1d1d",
          bg1 = "#252525",
          bg2 = "#2d2d2d",
          bg3 = "#424242",
        }
      },
    config = function(_, opts)
      require('onedark').setup(opts)
      require('onedark').load()
    end
  },
```

Tokarak:
The documentation now is unobtrusive, discoverable, and focused; I don't see how I could improve it.

I'm not going to add colour boilerplate like in @Triavanicus's example, because UI design is outside the scope of the target audience. 

Should be ready to merge.

PR #600, opened 2024-01-26
Set status line theme to the global colorscheme

v-bertin:
instead of hardcoding the 'onedark' theme.
It allows users to change the Neovim theme globally and to see those changes automatically reflected to the status line.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -205,7 +205,7 @@ require('lazy').setup({
     opts = {
       options = {
         icons_enabled = false,
-        theme = 'onedark',
+        theme = 'auto',
         component_separators = '|',
         section_separators = '',
       },


feoh:
Hi! I like this idea.

I've actually changed the theme in my fork, and needing to maintain one less change from master is always a boon in my book :)

1) What kind of testing have you done to ensure this won't break other users? (please be detailed)
2) Are there any downsides to this change?

Thanks for your contribution and I look forward to hearing from you!

v-bertin:
Hi !
Thank you for the quick answer.
Regarding the testing, I have considered different scenarios : 
- User uses 'onedark' colorscheme as the default kickstart.nvim theme and does not have the change : nothing breaks and the user does not have to change anything
- User uses 'onedark' colorscheme as the default kickstart.nvim theme and has the change already : nothing breaks and that's one less thing to maintain when rebasing on master
- User uses another colorscheme and does not have the change : nothing breaks but the status line theme now matches the used colorscheme. I don't see it as an issue, except if the user wants to use different themes for the editor itself and the status line ...
- User uses another colorscheme and has the change already : nothing breaks and that's one less thing to maintain when rebasing on master

I don't see any downsides as such ...
I think the main upside is that it makes it easier for new users to change Neovim colorscheme : they just have to add the new colorscheme in the plugins list with the right priority.
A downside could be for users that want to have different themes for the editor and the status line. However I don't think it is the original will as the 'onedark' theme is hardcoded in the status line config, to ensure it matches the global 'onedark' colorscheme.

Hopefully it answers your questions. Let me know :)

Knar89 (approved):

feoh:
Thanks so much for your contribution. I look forward to modifying my fork!

feoh:
you don't need to do anything for this change to work. I'm confused about what you're clueless about. :-)

Knar89:
Well I guess because I'm doing everything from an iPhone :/ I have no PC

PR #611, opened 2024-02-02
Add Build Step to LuaSnip

rgarber11:
`LuaSnip` requires a build step in order to use snippets with Regular Expressions in them. Snippets like these are very common (Header guards in C++). Currently, `kickstart.nvim` does not produce valid header guards. Configuring this would be confusing for beginners, who know that `friendly-snippets` and `LuaSnip` are already installed.
### Before: 
```cpp
#ifndef INCLUDE/home/rgarber11/tempconfigconfig.h_
// This is an invalid guard
...
```
### After: 
```cpp
#ifndef INCLUDE_TEMP_CONFIG_H_
// This guard is valid
...
```

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -105,7 +105,7 @@ require('lazy').setup({
         build = (function()
           -- Build Step is needed for regex support in snippets
           -- This step is not supported in many windows environments
-          -- Remove if your windows setup supports this
+          -- Remove the below condition to re-enable on windows
           if vim.fn.has 'win32' == 1 then
             return
           end
--- a/init.lua
+++ b/init.lua
@@ -102,7 +102,7 @@ require('lazy').setup({
       -- Snippet Engine & its associated nvim-cmp source
       {
         'L3MON4D3/LuaSnip',
-        build = 'make install_jsregexp'
+        build = 'make install_jsregexp',
       },
       'saadparwaiz1/cmp_luasnip',
 
--- a/init.lua
+++ b/init.lua
@@ -100,7 +100,10 @@ require('lazy').setup({
     'hrsh7th/nvim-cmp',
     dependencies = {
       -- Snippet Engine & its associated nvim-cmp source
-      'L3MON4D3/LuaSnip',
+      {
+        'L3MON4D3/LuaSnip',
+        build = 'make install_jsregexp'
+      },
       'saadparwaiz1/cmp_luasnip',
 
       -- Adds LSP completion capabilities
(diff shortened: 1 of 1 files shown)


feoh:
Will this patch work on non UNIX platforms? We have a LOT of mostly silent Windows users who we only hear from when there's a problem.

dam9000:
Hmm there is indeed a windows built error reported related to this:
https://github.com/L3MON4D3/LuaSnip/issues/986

quite unfortunate, perhaps the change should be ifdef-ed to exclude windows?

feoh:
An ifdef or alternative build steps would be grand, yes.

rgarber11:
I could not get this change working on Windows, so I can close the PR if that is a hard requirement.
On the other hand, this PR doesn't really break anything for Windows (on first install, you see a build failed message for LuaSnip, but LuaSnip will still function as before). I would argue that a change that improves things for ~60% of users, while not really doing anything for ~40% of users might still be worth considering.
If there is a way to ifdef within Lua's kickstart.nvim, then I could definitely look into adding that too, which would leave Windows users with an experience exactly the same as before this PR.

feoh:
I'm a little bit concerned about creating a bad user experience for windows users. I can appreciate that this doesn't actually break anything but I also feel like giving them a face full of build failed when they do the install is not what we're going for.

Would this possibly make sense as a section added to the README for C++ programmers?

dam9000:
I think this is one option:

1. skip build on windows:
(not tested)
```
      {
        'L3MON4D3/LuaSnip',
        build = (function()
          -- remove the below condition if you are using windows
          -- and you have the development environment setup
          if vim.fn.has("win32") == 1 then return end
          return 'make install_jsregexp'
        end)()
      },
```


2. alternatively, ignore exit error code:
(also not tested)
```
      {
        'L3MON4D3/LuaSnip',
        build = 'make install_jsregexp || true'
      },
```
the above approaches could be combined to "if on windows ignore exit code"

feoh:
@dam9000 LOVE this! @rgarber11 You game for a change like that? Just test it under Windows and see that it doesn't blow up and I'll merge your PR.

Thanks everyone!

rgarber11:
I can test this under Linux and windows later tonight

rgarber11:
I tested the suggestion from @dam9000 and the first one was successful on Windows. I'm going to change the comment slightly, since even with MinGW along with Lua, I still could not get this library to build on Windows. 
```lua
        build = (function()
          -- Build Step is needed for regex support in snippets
          -- This step is not supported in many windows environments
          -- Remove the below condition to re-enable on windows
          if vim.fn.has 'win32' == 1 then
            return
          end
          return 'make install_jsregexp'
        end)(),
```

PR #614, opened 2024-02-03
NVIM tree configurations.

anurag-prakash-singh:


feoh:
Hi. There are several Neovim filesystem browsers. I use Neotree.

I don't think we should merge this. Please feel free to make this change in your own fork.

PR #625, opened 2024-02-13
Add key mapping to exit terminal mode via ESC

r2p2:
I was unable to exit the terminal mode using the default shortcut `C-\` followed by `C-n` when using 'de' as keyboard layout. Maybe because of the required additional modifier key to access backspace. This caused me to always `exit` the shell in order to get back to the edit buffers.

I would assume that more people are affected when using a different keyboard layout than 'us' and would benefit from a simpler default.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -342,6 +342,9 @@ vim.keymap.set('n', ']d', vim.diagnostic.goto_next, { desc = 'Go to next diagnos
 vim.keymap.set('n', '<leader>e', vim.diagnostic.open_float, { desc = 'Open floating diagnostic message' })
 vim.keymap.set('n', '<leader>q', vim.diagnostic.setloclist, { desc = 'Open diagnostics list' })
 
+-- Exit terminal mode by simply pressing ESC
+vim.keymap.set('t', '<ESC>', '<C-\\><C-n>', { noremap = true })
+
 -- [[ Highlight on yank ]]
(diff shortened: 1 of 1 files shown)


dam9000:
I don't know, this looks too invasive, what if someone runs an interactive program, like for example `make menuconfig` or any kind of tool that uses `dialog` then the `esc` key will not work inside the terminal...

feoh:
thank you for your contribution, but this is not appropriate for the kickstart project. Please create your own fork and make such changes there.

dam9000:
@r2p2 I tried changing the keyboard to german (de) and it works for me, that is, to get `\` i need to press alt ß (or alt - (minus) on an eng keyboard) and pressing ctrl alt ß works fine at getting the ctrl `\`, followed by ctrl-n exits the terminal mode.

r2p2:
> @r2p2 I tried changing the keyboard to german (de) and it works for me, that is, to get `\` i need to press alt ß (or alt - (minus) on an eng keyboard) and pressing ctrl alt ß works fine at getting the ctrl `\`, followed by ctrl-n exits the terminal mode.

Thank you for testing, but I checked again and it does not work for me. Maybe I do something else wrong. Nevertheless, my ESC workaround works for me and I thought others could profit too. But I see that this might not be within the scope of this project and/or that others might not have the same issue.

PR #634, opened 2024-02-20
Fix setting up mason.nvim and mason-lspconfig.nvim

williamboman:
These are currently being set up twice:
- mason.nvim once by lazy.nvim (`config = true`) and then once manually
- mason-lspconfig.nvim twice manually


feoh:
Hi! This is being handled in https://github.com/nvim-lua/kickstart.nvim/pull/635 as near as I can tell. Please free to re-open if I'm wrong, and thanks for your contribution!

PR #646, opened 2024-02-26
fix: Should default to false

GeorgiChochev:


tjdevries:
I actually want this to be true, because it's confusing for people. I just forgot to update the doc. Pushed an update in af4fd23

PR #654, opened 2024-02-27
Fix: Add treesitter option to use pre-compiled binaries

feoh:
On Windows, when new users install kickstart it blows up by default because having a C compiler in PATH is not a reaonle expectation on hat platform.

Also, this should decrease first use startup time substantially on all platforms because curl-ing down a binary is much faster than compiling from scratch.

Also updated README as Windows folks are expected to turn on developer mode so symlinks and 'curl' will work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,11 @@ return {
 
 Installation may require installing build tools, and updating the run command for `telescope-fzf-native`
 
+Users should follow the [Treesitter Windows
+Installation](https://github.com/nvim-treesitter/nvim-treesitter/wiki/Windows-support)
+instructions to allow Treesitter language extensions to be downloaded from
+Github.
+
 See `telescope-fzf-native` documentation for [more details](https://github.com/nvim-telescope/telescope-fzf-native.nvim#installation)
 
 This requires:
--- a/init.lua
+++ b/init.lua
@@ -771,6 +771,9 @@ require('lazy').setup({
     build = ':TSUpdate',
     config = function()
       -- [[ Configure Treesitter ]] See `:help nvim-treesitter`
+      -- Use precompiled binaries for faster startup.
+      -- This also means Windows users won't fail on startup by default.
+      require('nvim-treesitter.install').prefer_git = false
 
       ---@diagnostic disable-next-line: missing-fields
       require('nvim-treesitter.configs').setup {


feoh:
@dam9000 @sodoku @rubin @kdheepak and anyone else - could folks please give this a look? It will help us not 💥 by default at startup on Windows and speed up startup times for everyone by not requiring folks to compile treesitter language extensions when binaries are available from Github releases.

dam9000:
This looks like a great idea, I mostly use Linux so I don't have an immediate way to test it out, but might try sometime later.

feoh:
@dam9000 Thanks! It will help (and work just fine!) for Linux folks as well.

It should speed up the initial Neovim startup because rather than compiling all those treesitter extensions, it'll just automatically fetch the Linux compiled binaries from Github releases.

dam9000:
Hmm, by reading the page you linked:
https://github.com/nvim-treesitter/nvim-treesitter/wiki/Windows-support

it does not sound to me that this change will actually use the prebuilt binaries but just change the way the (source?) packages are downloaded to `curl` instead of `git clone` and it seems on linux curl is already the default method while on windows is `git` because it requires an additional Admin mode step:

> enable creating symbolic links for non-admin users (enable "Developer Mode": https://www.ghacks.net/2016/12/04/windows-10-creators-update-symlinks-without-elevation) or refrain from using parsers that use symbolic links in their repos (tree-sitter-typescript, tree-sitter-ocaml). Otherwise tar will fail on extracting those archives on Windows

So in a way, although it might be faster compared to git on Windows this add an additional preparation step that would not be required otherwise?

As for the binaries it says:

> Pre-compiled parsers (disabled at the moment -> we want to create a proper binary releases in future)
> Important: this feature is disabled at the moment. We want to create proper binary releases for people to download in future

Unless the doc is wrong and the pre-compiled parsers actually work?

feoh:
Yup I think the doc is wrong.

I say that because *with* this change I can start neovim with kickstart and get zero errors, with Treesitter operational.

Without that change, it explodes every single time *unless* I use a Powershell with the C compiler in PATH, which is apparently a BIG antipattern according to every expert on Windows I've talked to.

I'll go poke them about updating that doc or telling me I'm mistaken :)

feoh:
thanks so much for digging so deeply into this. 

can't blame me for trying I guess :-)

PR #659, opened 2024-02-28
Change mini.statusline location format to LINE:COLUMN

dam9000:
Default mini.statusline location format is:
  'cursor line | total lines │ cursor column | total columns'

For example: `512|816|35|95`

Which just doesn't look good to me. Before the latest kickstart rewrite, the `lualine` location format was LINE:COLUMN (`512:35`) which I think is what most would expect. I understand this is a subjective preference so feel free to discard this PR if you think the current default is fine.

PS. an additional note, as this overwrites a mini.statusline function, the lua diagnostics show a warning:  `Duplicate field 'section_location'` at the added line but it works fine. I didn't find any other elegant way to avoid this warning, at least not with this small number of lines of code. An alternative would take about 30 lines, which seems way too much for such a small change. Perhaps the previous `lualine` was a better statusline plugin option.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -762,6 +762,9 @@ require('lazy').setup {
       --  You could remove this setup call if you don't like it,
       --  and try some other statusline plugin
       require('mini.statusline').setup()
+      MiniStatusline.section_location = function()
+        return '%2l:%-2v'
+      end
 
       -- ... and there is more!
       --  Check out: https://github.com/echasnovski/mini.nvim


feoh:
100% agree.

No opinion on lualine versus mini. Maybe chat with @tjdevries about it if you can get him to chime in? :)

dam9000:
Thanks.

dam9000:
@echasnovski right, it's not 30 but 20, I did not count it at the time but just eyeballed it, the point is, TJ mentioned he wanted to keep the init.lua under 300 lines of code, so such a big block would be a substantial portion of the whole config. In any case thanks for the feedback!

PR #660, opened 2024-02-29
feat: enable lua lsp snipppets

antonk52:
This makes the builtin snippets work with luasnip

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -562,6 +562,9 @@ require('lazy').setup {
                 -- If lua_ls is really slow on your computer, you can try this instead:
                 -- library = { vim.env.VIMRUNTIME },
               },
+              completion = {
+                callSnippet = 'Replace',
+              },
               -- You can toggle below to ignore Lua_LS's noisy `missing-fields` warnings
               -- diagnostics = { disable = { 'missing-fields' } },
             },


feoh:
I have zero understanding of how snippets work despite watching @tjdevries excellent video series a couple times :)

So it would be great if others could take a look at this.

antonk52:
To clarify, the lua_ls has snippet support disabled by default. This PR enables it

https://github.com/LuaLS/lua-language-server/blob/master/doc/en-us/config.md#completioncallsnippet

feoh:
OK, thanks for that.

I'm still a bit reluctant to approve/merge based on my own limited understanding.

Is that desirable? Was not enabling that by default just an omission? Are there other snippet-y decisions this change could affect?

Sorry for being a nervous ned, but there are a LOT of folks who use this who have strong preferences :)

antonk52:
Fair questions!

This only affects the completion results when working in lua files. 

two main outcomes

- **Builtin snippets** - A good example is if you type `fo` you will get a few completion items for different variations of `for` loops that you can expand as a snippet. 
- **Snippet like completion for functions** - before this change when you complete a function name that takes arguments it only completes the function name, ie `my_f<tab>` becomes `my_func`. With this change it will also completes the function call with arguments ie `my_f<tab>` becomes `my_func(arg1, arg2)`

> Is that desirable? Was not enabling that by default just an omission?

These are good questions. I am not aware of a reason why this is not an opt out features by default. Installing and using an lsp is an intent to have completion working. I do not know a reason why this is not enabled out of the box. Yet everyone I told about this enabled it and have been using it ever since.

feoh:
Thanks for the thoughtful responses.

We can always revert if folks have an issue with this.

Merging.

PR #672, opened 2024-03-02
Add a section on user commands

clement-or:
**Summary**
This PR adds a custom user command in kickstart.nvim as well as comments explaining how to create such commands and where to find help.
This command is `:Config`, which just switches the current buffer to the user's config.

**Why?**
Kickstart is aimed at beginners, and when you're a beginner you modify your config a lot. And you encounter problems a lot. When that happens, it can be cumbersome to cd to your config directory everytime an error pops up. A simple command such as `Config` (or `Cfg` as I use myself) allows newbies to switch to their config when an error pops up, save quand quit, restart neovim and go on with their work. 

**Possible improvements**
I currently only `:e` the path to the config, which doesn't change the current working directory. We might want to use `:cd` so users can edit multiple files in their config using netrw or telescope.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -202,10 +202,9 @@ vim.api.nvim_create_autocmd('TextYankPost', {
 -- See `:help lua-guide-commands-create`
 
 -- Quickly jump to your config with this awesome custom command
--- Try it by starting nvim in a random directory and typing :Config
-vim.api.nvim_create_user_command("Config", function()
-  local path = "~/.config/nvim/" -- You can customize the path here
-  vim.cmd("e" .. path)           -- vim.cmd allows you to execute vim commands from lua
+--  It aliases :e $MYVIMRC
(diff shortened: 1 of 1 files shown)


dam9000:
I don't like this.
This won't work on windows because the path is different and it also won't respect the `NVIM_APPNAME` which also affects the config path.
All the user needs to do is `:e $MYVIMRC` this works for all above mentioned cases and is simple enough that IMO does not need a new command to be added. It will also expand properly if you type `:e $MY<TAB>`. But I agree it would be worth mentioning in the README.md or directly in the `init.lua`.

clement-or:
I didn't know about the $MYVIMRC environment variable. Adding the updated snippet would allow for a way to add custom commands and a shorthand for :e $MYVIMRC.

feoh:
Thank you very much for your contribution but I don't think this changes is appropriate for this project at this time.

PR #673, opened 2024-03-02
feat: use VimEnter event instead of VeryLazy

### LINKED ISSUE #655: Intro disappears immediately after starting up neovim
## Describe the bug

I got the fresh kickstart installation I haven't changed anything, When i launch neovim the intro screen disappears immediately
## To Reproduce

1. ...

## Desktop

- OS: Arch linux
- Terminal: Alacritty

## Neovim Version

Latest version
```
```

taliraj:
Fixes #655.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -258,9 +258,9 @@ require('lazy').setup {
   -- lazy loading plugins that don't need to be loaded immediately at startup.
   --
   -- For example, in the following configuration, we use:
-  --  event = 'VeryLazy'
+  --  event = 'VimEnter'
   --
-  -- which loads which-key after all the UI elements are loaded. Events can be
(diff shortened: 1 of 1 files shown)


dam9000:
A note, while the fix is correct, now the following comment is no longer correct:
```
  -- For example, in the following configuration, we use:
  --  event = 'VimEnter'
  --
  -- which loads which-key after all the UI elements are loaded. Events can be
```
Because VimEnter happens before UI.

taliraj:
> Perhaps just change it to `before all the UI elements are loaded`

Good point 948004162ace7dba1df7fcce7b7fab1f3efd29f9

dam9000 (approved):
LGTM :)

feoh:
Thank you for your contributions everyone!

PR #691, opened 2024-03-04
Add separate comment for `timeoutlen` option

TinyLittleWheatley:
`timeoutlen` option was under unrelated comment with `updatetime` option.


tjdevries:
Thanks :)

Issue #686, opened 2024-03-03
Disable formatting for some LSP


qianxyz:
In the lspconfig section it is mentioned that we can disable certain LSP features by setting `capabilities`. Is there any example for turning off formatting for e.g. tsserver?

SebasF1349:
This is from LazyVim: https://github.com/LazyVim/LazyVim/discussions/242#discussioncomment-4947842

But basically you can add
```lua
if client.name == "tsserver" then
    client.server_capabilities.documentFormattingProvider = false
end
```
to the LspAttach autocomd, inside the nvim-lspconfig config and after defining the client (so after line 503)

dam9000:
@qianxyz do you want to disable lsp formatting completely, or just the autoformat on save? If it's the autoformat on save, there's a recipe from conform.nvim that I use, perhaps I should open a PR for that...

qianxyz:
@dam9000 Thanks! I'm aware of the conform.nvim recipe. Basically I'm missing the old behavior with [`autoformat.lua`](https://github.com/nvim-lua/kickstart.nvim/blob/7af594fd319fbae6b2aaa06337f3df8acbbb7f18/lua/kickstart/plugins/autoformat.lua) where there is a `:Format` command, and you can toggle format on save globally and filter on `client.name`. All of these are configurable through conform.nvim now so functionally I'm all set. 

I'm just curious since this is explicitly mentioned in the comments 
https://github.com/nvim-lua/kickstart.nvim/blob/b83b2b061c1fab0e1a3a28c185345be7957e74cd/init.lua#L593-L600
and I'm wondering how to do that here since people may want to disable lsp formatting completely. But as you said, just disabling format on save makes more sense to me now. Thanks!

dam9000:
I opened a PR for the "autoformat on save only for specified filetypes"
https://github.com/nvim-lua/kickstart.nvim/pull/694

Chewie:
I think the comments are misleading: the `capabilities` field in the `servers` variable are for defining the client capabilities, but `documentFormattingProvider` is a server capability. It could instead be modified through a `on_attach` callback by modifying `client.server_capabilities`.

But for this particular case, I instead added a filter field to the `vim.lsp.buf.format` to exclude tsserver:

```lua
nmap('<leader>f', function()
    vim.lsp.buf.format {
        async = true,
        filter = function(client) return client.name ~= "tsserver" end,
    }
end, '[F]ormat')
```

PR #694, opened 2024-03-04
conform: autoformat on save only for specified filetypes

dam9000:
Fixes: #686

Enable the autoformat on save only for specified filetypes.
By default this is only for lua, similar as LSP is by default only enabled for lua.
Based on conform recipe:
https://github.com/stevearc/conform.nvim/blob/master/doc/recipes.md

In my view having the autoformat on save enabled by default for all filetypes is a bit too intrusive, hence this PR.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -603,10 +603,16 @@ require('lazy').setup({
     'stevearc/conform.nvim',
     opts = {
       notify_on_error = false,
-      format_on_save = {
-        timeout_ms = 500,
-        lsp_fallback = true,
-      },
+      format_on_save = function(bufnr)
+        -- Disable "format_on_save lsp_fallback" for languages that don't
+        -- have a well standardized coding style. You can add additional
+        -- languages here or re-enable it for the disabled ones.
+        local disable_filetypes = { c = true, cpp = true }
+        return {
+          timeout_ms = 500,
+          lsp_fallback = not disable_filetypes[vim.bo[bufnr].filetype],
+        }
+      end,
       formatters_by_ft = {
         lua = { 'stylua' },
         -- Conform can also run multiple formatters sequentially


fredrikaverpil:
@dam9000 nice!
I will personally create a global toggle which toggles formatting on/off instead.

dam9000:
@fredrikaverpil right, there are multiple ways to customize the `write_on_save` logic and there are many examples in the conform recipes. Including all of them would be an overkill. It's true the selection of when to enable/disable will be opinionated that's why it's good to provide a place where this can be customized. I think the important thing is to change the `format_on_save` from a hardcoded always enable table to a function to give the user a hint and a place where to customize this. Also I think enabling it for `lua` is a reasonable default, because most likely the only lua files user will be editing is the kickstart `init.lua` itself, which has a well defined style in `.stylua.toml`. The same cannot be said for all the other file types that someone works on.

dam9000:
@feoh I said "intrusive" not "intensive". English is not my primary language so maybe the term used was not the best choice although I think it describes my feeling exactly:
- intrusive: causing disruption or annoyance through being unwelcome or uninvited.

By that I mean the situation I was put in - I updated the kickstart  config to the newest version (after the rewrite), went to edit my usual source code files and was taken by surprise that they all got reformatted, that was an unexpected and unwelcome change. I don't know if any editor out there enables "reformat on save" by default, that surely has to be something that a person will want to enable explicitly, only after he makes sure that the formatting style matches the expectation. So I first needed to find out what exactly is causing the reformat and then find out a way to limit it to only lua and that resulted in this PR.

feoh:
your English is fantastic as always! My reading comprehension was at fault :-) to be honest I haven't been bitten by this yet in personally when I'm programming I really appreciate having my code auto format for me on save but perhaps I'm in the minority. In any case, thank you for your contribution and I'm going to leave this one for others to comment and or possibly act on.

dam9000:
Maybe it can be nice for when you are starting a project from scratch. But when editing existing code the style will likely not match and if the files are in git then you get a bunch of diffs all over the place. Also some languages such python have a standardized coding style while for C/C++ there are uncounted number of styles and each person (or team) will defend the style that they prefer. So for C and C++ I think this should definitely not be enabled by default. In my view this should be a opt-in as no other editor that I'm aware of has this enabled by default, I checked: sublime, jetbrains, vs, vs code. Also I expect most people trying out Neovim/kickstart will already have some existing projects so let's not take them by surprise with such default behaviour.

tjdevries:
What if we do the reverse, and we make it easy to *disable* a language (we don't have to populate the list with anything, we can just show them that they would need to add it)?

I think this might be a nice middle ground?

tjdevries:
```lua
local disable_filetypes = { c = true, cpp = true }
return {
  timeout_ms = 500,
  lsp_fallback = not disable_filetypes[vim.bo[bufnr].filetype],
}
```

dam9000:
force pushed: changed as suggested, and tested that it works as expected.

Issue #692, opened 2024-03-04
Completion is not complete....


abeldekat:
## 

![1709574575](https://github.com/nvim-lua/kickstart.nvim/assets/58370433/693b4fdf-db82-423c-8d34-4ce8d964db87)

## Describe the bug

See [reddit feedback](https://www.reddit.com/r/neovim/comments/1b12oyb/comment/kt6y1gl/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button)

Thanks for a new kickstart version!

I was wondering: Why has neodev.nvim been removed?

For example, when typing `require("cmp").s` inside init.lua, the setup method is not in the list of suggestions. 

I read in this [pr](https://github.com/neovim/neovim/pull/24592#issuecomment-1671044254) that `neodev.nvim` is not needed anymore. However, the PR addresses Neovim `nightly` and the usage of `.luarc.json` when working on the `Neovim` codebase.

On version 0.9.5, without a `.luarc.json`, I *think* `neodev.nvim` is still relevant.
The plugin adds the `lua` path for each plugin to `workspace.library`, and also the `types`, which are now included setup in Neovim nightly.

On nightly, somehow the `lua` paths of the plugins need to be added for the completion experience of the previous `kickstart`.

## To Reproduce

1. Install kickstart in `~/.config/kickstart`
1. `NVIM_APPNAME=kickstart nvim init.lua`
2. type `require("cmp").s`
3. Result: only text completion

## Desktop

- OS: Arch linux
- Terminal: Alacritty

## Neovim Version

0.9.5
NVIM v0.10.0-dev-2507+g3df1211eb
```
```

PR #696, opened 2024-03-05
Use `cmp-nvim-lua` as `nvim-cmp` source for neovim Lua API

jamylak:
Things like `vim.keymap.set` were not autocompleting for me and I thought I heard @tjdevries mention that we should be able to type commands in our config without having to "wing it" unless I missed a later part of the video where this was added and removed, but I think it's very helpful for editing your config

https://github.com/hrsh7th/cmp-nvim-lua/tree/main

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -629,9 +629,6 @@ require('lazy').setup {
     dependencies = {
       -- Snippet Engine & its associated nvim-cmp source
       {
-        -- nvim-cmp source for neovim Lua API
-        -- so that things like vim.keymap.set, etc. are autocompleted
-        'hrsh7th/cmp-nvim-lua',
         'L3MON4D3/LuaSnip',
         build = (function()
           -- Build Step is needed for regex support in snippets
@@ -650,6 +647,9 @@ require('lazy').setup {
       --  into multiple repos for maintenance purposes.
       'hrsh7th/cmp-nvim-lsp',
       'hrsh7th/cmp-path',
+      -- nvim-cmp source for neovim Lua API
+      -- so that things like vim.keymap.set, etc. are autocompleted
+      'hrsh7th/cmp-nvim-lua',
 
       -- If you want to add a bunch of pre-configured snippets,
       --    you can use this plugin to help you. It even has snippets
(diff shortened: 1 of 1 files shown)


dam9000:
I can confirm that this used to work with the old kickstart before the last rewrite and that with this change it works again.

One more thing, the old kickstart also provided a detailed description of the selected item, for example:
```
vim.keymap.s 
           set(mode, lhs, rhs, opts) Function  function                                    
                                               function keymap.set(mode: string|table,     
                                               lhs: string, rhs: string|function, opts:    
                                               table|nil)                                  
                                               ──────────────────────────────              
                                                Adds a new |mapping|.                      
                                                Examples:                                  
                                                  -- Map to a Lua function:                
                                                  vim.keymap.set('n', 'lhs', function()    
                                                  print("real lua function") end)          
```
Now it just says `function` without any details:

```
vim.keymap.s                                                                             
           set                       Function   function      
```

Any idea why, and how to improve that to match the old kickstart version?

abeldekat:
Perhaps related: [Completion is not complete...](https://github.com/nvim-lua/kickstart.nvim/issues/692)

dam9000:
Ok figured out the `vim.keymap.set` fix, here is a PR:
https://github.com/nvim-lua/kickstart.nvim/pull/754

dam9000:
Hmm, with neodev change merged, this PR is not necessary, in fact now there are two completions offered, the first coming from cmp-nvim-lua: without parameters and without help and the second from neodev (actually nvim_lsp which neodev sets up), which has the parameters and help:

image


image

or at least the sources order should be changed like this:
```
index fe25648..7c2cf87 100644
--- a/init.lua
+++ b/init.lua
@@ -704,8 +704,8 @@ require('lazy').setup({
           end, { 'i', 's' }),
         },
         sources = {
-          { name = 'nvim_lua' },
           { name = 'nvim_lsp' },
+          { name = 'nvim_lua' },
           { name = 'luasnip' },
           { name = 'path' },
         },
```
So that the neodev set nvim_lsp takes precedence and it looks like this:

image

PR #702, opened 2024-03-05
Replace `vim-sleuth` with `guess-indent.nvim`

chrisgrieser:
[guess-indent.nvim](https://github.com/NMAC427/guess-indent.nvim) is a lua version of `vim-sleuth`. It also has the advantage of having an option to override `.editorconfig`, if needed.

However, it does require a `.setup()` call, so a `opts = {}` will be required, which slightly disturbs the "introductory flow" of kickstart. To keep the introduction working, I placed `Comment.nvim` first (as it does not require a `.setup()` call), and put `guess-indent` second.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -220,7 +220,7 @@ vim.opt.rtp:prepend(lazypath)
 -- NOTE: Here is where you install your plugins.
 require('lazy').setup {
   -- NOTE: Plugins can be added with a link (or for a github repo: 'owner/repo' link).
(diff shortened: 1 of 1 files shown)


feoh:
I don't personally have any feelings about this change, as I have had no problem with Tim Popes plug-in, and have no need to override editor config. Other folks, please chime in here.

dam9000:
I have no preference, I disable this in my personal config anyway as I find the guessing unreliable.

tjdevries:
I like tpope's version, so we'll leave it for now. I may revisit later though. Thanks for the PR

Issue #700, opened 2024-03-05
Strange glyph icons appear.


joram56:
## Describe the bug
Following the guide of cloning the repository and opening up neovim, the setup completes, but I see strange glyphs everywhere instead of the intended icons. 
![image](https://github.com/nvim-lua/kickstart.nvim/assets/25789485/b0efc6f0-5aad-4e38-adf4-9c2d79d5494a)

## To Reproduce
I assume this has to do with the fonts on a system, so not sure how to reproduce.

## Desktop

- OS: Ubuntu:20.04 LTS
- Terminal: zsh

## Neovim Version
0.9.1

dam9000:
@joram56 you need to install one of the `Nerd fonts`: https://www.nerdfonts.com/
There's a huge selection, pick one, install it and then configure your terminal to use it.

joram56:
Since this is a requirement, can it be added to the "Install External Dependencies" section of the readme?

dam9000:
@joram56 You are right, I opened a PR: https://github.com/nvim-lua/kickstart.nvim/pull/703

dam9000:
Here is a new PR for this issue: https://github.com/nvim-lua/kickstart.nvim/pull/716
Which makes the Nerd Font an optional requirement that can be toggled via a global flag.

PR #703, opened 2024-03-05
README.md: add a Nerd Font requirement

dam9000:
Fixes nvim-lua/kickstart.nvim#700
A Nerd Font is required by:
- mini.statusline
- Lazy plugin manager
- Neo-tree.nvim - not enabled by default however instructions on how to add it are in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ If you are experiencing issues, please make sure you have the latest versions.
 External Requirements:
 - Basic utils: `git`, `make`, `unzip`, C Compiler (`gcc`)
 - [ripgrep](https://github.com/BurntSushi/ripgrep#installation)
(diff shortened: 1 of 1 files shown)


tjdevries:
Hmm, I think it might be better to try and disable the nerdfonts by default for mini and lazy. Do you by any chance happen to know how to do that? Otherwise I will look.

dam9000:
I think I do, I can open a PR for that, although to be honest I like the look of nerd font icons better.

feoh:
I really love the look of nerd font, icons too! Why are we getting rid of that again?

I guess asking people to download and install fonts is an extra step. Let's just make them easy to turn back on again for those of us who love them and have them installed by default as the first thing we do when we come to a new system :-)

dam9000:
Closing this since a better alternative was merged.

Issue #680, opened 2024-03-03
uncommenting friendly-snippets does not work


sqrtnull:
## Describe the bug

just uncommenting rafamadriz/friendly-snippets does not work

## To Reproduce

1. uncomment rafamadriz/friendly-snippets in the dependencies for nvim-cmp

## Desktop

- OS: macos
- Terminal: alacritty

## Neovim Version

```
NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1703358377
```
# Fix
add `require("luasnip.loaders.from_vscode").lazy_load()` when setting up cmp
(maybe also add rafamadriz/friendly-snippets to the dependencies for LuaSnip)

tjdevries:
Yes, you still have to do the setup for friendly snippets -- could maybe be noted better in the documentation though. Do you want to send a PR w/ a description to add?

sqrtnull:
Since my PR was closed and I don't have any better ideas, here is an explanation on how to solve this for anyone with the same problem. Add the line
`require("luasnip.loaders.from_vscode").lazy_load()`
below the following code. (since the old kickstart had the same line here)

https://github.com/nvim-lua/kickstart.nvim/blob/b529bc33590cbb81a5916408b2d6001a643e596c/init.lua#L650-L654

also maybe just read https://github.com/rafamadriz/friendly-snippets

tjdevries:
@VlaDexa had a really smart suggestion, which we just merged. Should make this easier!

feakuru:
I have a config forked from latest Kickstart (https://github.com/feakuru/nvim-config), and when I try to uncomment the `friendly-snippets` section there, the snippets for Python, HTML or JS don't seem to show up in completions. Am I missing some additional steps?

PR #709, opened 2024-03-05
add documentation on how to set up friendly-snippets

sqrtnull:
Added documentation to clarify #680

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -652,6 +652,10 @@ require('lazy').setup {
       --    you can use this plugin to help you. It even has snippets
       --    for various frameworks/libraries/etc. but you will have to
       --    set up the ones that are useful for you.
+      -- You will need to add the following line somewhere
+      --    eg: under luasnip.config.setup {}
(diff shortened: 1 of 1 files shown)


sqrtnull:
I am unsure whether it would be better to just outright enable vscode-style snippets for LuaSnip so people would not encounter this problem, since a comment telling you to insert code somewhere else seems weird (to me).

feoh:
Hi.

This would need substantial rework in order to be useful. Also I'm not sure that providing 1000% VSCode compat is the goal of this project.

In particular, should you choose to re-open and pursue this, you'd need explicit instructions as to where to put what so as to be more friendly for newer folks which is the target audience of this project.

Thank you for your contribution but I'm closing for now.

sqrtnull:
@feoh I am not sure what you are getting at here; the friendly-snippets plugin requires LuaSnip to accept vscode-style snippets for it to work properly, thus the comment/suggestion. That being said, I agree with you in that this commit itself is not the way to go.

PR #711, opened 2024-03-05
Remove Nerd Font requirements for lazy.nvim and mini.statusline

dam9000:
Fixes nvim-lua/kickstart.nvim#700

This has two changes:
- lazy.nvim: use unicode instead of Nerd Font icons.
- mini.statusline: disable icons for git branch, lsp, filetype (filetype icons are only enabled if nvim-web-devicons is also loaded)

For reference, the unicode icons are copied from the suggestion from:
https://github.com/folke/lazy.nvim/blob/aedcd79811d491b60d0a6577a9c1701063c2a609/README.md?plain=1#L473
> If you don't want to use a Nerd Font, you can replace the icons with Unicode symbols.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -218,7 +218,7 @@ vim.opt.rtp:prepend(lazypath)
 --    :Lazy update
 --
 -- NOTE: Here is where you install your plugins.
-require('lazy').setup {
+require('lazy').setup({
   -- NOTE: Plugins can be added with a link (or for a github repo: 'owner/repo' link).
   'tpope/vim-sleuth', -- Detect tabstop and shiftwidth automatically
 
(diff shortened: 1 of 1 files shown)


dam9000:
For reference, here is how it looks like on a ubuntu 22.04 without nerd fonts and with this change:

mini-statusline-no-icons

lazy-unicode

I personally prefer the look of Nerd Font icons, but you decide.

joram56:
Went through the recent youtube tutorial, and I felt like nerd fonts does not really fit in with the general level of basic configurations the rest of the project does. It feels like a very user specific configuration.

dam9000:
Closing this since a better alternative was merged.

PR #715, opened 2024-03-06
Use lazy's init instead of disabling lazy loading for tokyonight

name-tar-xz:
Lazy's init function runs at startup and colorschemes are auto loaded when called so it removes the need of disabling lazy loading.


brxxlstxrs:
Good solution, strange, for some reason I didn’t seem to see this in the examples of lazy.nvim

PR #718, opened 2024-03-06
Add colorscheme to lazy on its startup (aka plugins installation)

brxxlstxrs:
Make lazy.nvim look cute
(Think about unifying colorscheme settings, maybe make something like a global variable?)

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -820,6 +820,8 @@ require('lazy').setup({
   --    For additional information, see `:help lazy.nvim-lazy.nvim-structuring-your-plugins`
   -- { import = 'custom.plugins' },
 }, {
+  -- Make lazy.nvim try to load one of these colorschemes when starting an installation during startup
+  install = { colorscheme = { 'tokyonight-night', 'habamax' } },
   ui = {
     -- If you have a Nerd Font, set icons to an empty table which will use the
     -- default lazy.nvim defined Nerd Font icons otherwise define a unicode icons table


feoh:
Thanks for your contribution but I'm concerned about adding yet another point of tight coupling with he color scheme.

I think this would make it more difficult for folks who need to change this in a fork.

feoh:
OK. Let me give you a very concrete example.

Before we switched kickstart to use tokyo-night, which I can in fact use and handle, they were using onedark.

I'm partially blind, and as such that color scheme literally gave me a blinding headache, so I forked the kickstart project and changed it.

Back then, there were 3 places sprikled across the code that you had to change to make that happen.

Now, there is exactly one. I am loathe to increment that number and create yet more hassle for folks looking to customize kickstart in their own forks.

brxxlstxrs:
> OK. Let me give you a very concrete example.
> 
> Before we switched kickstart to use tokyo-night, which I can in fact use and handle, they were using onedark.
> 
> I'm partially blind, and as such that color scheme literally gave me a blinding headache, so I forked the kickstart project and changed it.
> 
> Back then, there were 3 places sprikled across the code that you had to change to make that happen.
> 
> Now, there is exactly one. I am loathe to increment that number and create yet more hassle for folks looking to customize kickstart in their own forks.

Why then is there an icon setting now?
I hope if you are so worried about this there are no other similar points in kickstart.nvim now

feoh:
it sounds like you think I'm over indexing on this. I want to be fair, so if that's what you think, please feel free to reopen this and it won't take any action on it and will allow other members of the community to comment and review it. Thank you for your contribution and thank you for being willing to talk about your goals in a constructive way.

PR #719, opened 2024-03-07
chore: rename <C-T> to <C-t> for consistency

ChillerDragon:
So far init.lua lowercased all ``<C-x>`` key mappings. As a vim nob I then assumed that ``<C-T>`` means ctrl+shift+t Because this one was suddenly uppercased.
But seems like the casing is ignored and I had to actually press ctrl+t or is this my terminal emulator messing with me and ctrl+shift+t also works. And ctrl+t is just the fallback because of the ctrl+shift problem with terminal emulators?

```
$ grep -Er '<C-[A-Z]>' init.lua lua/ | wc -l
1

$ grep -Er '<C-[a-z]>' init.lua lua/ | wc -l
12
```

I use gnome BTW

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -456,7 +456,7 @@ require('lazy').setup({
 
           -- Jump to the definition of the word under your cursor.
           --  This is where a variable was first declared, or where a function is defined, etc.
-          --  To jump back, press <C-T>.
+          --  To jump back, press <C-t>.
           map('gd', require('telescope.builtin').lsp_definitions, '[G]oto [D]efinition')
 
           -- Find references for the word under your cursor.


dam9000:
The ctrl-x mapping in vim/nvim is not case sensitive:

```
							*CTRL-{char}*
CTRL-{char}	{char} typed as a control character; that is, typing {char}
		while holding the CTRL key down.  The case of {char} is
		ignored; thus CTRL-A and CTRL-a are equivalent.  But in
		some terminals and environments, using the SHIFT key will
		produce a distinct code (e.g. CTRL-SHIFT-a); in these
		environments using the SHIFT key will not trigger commands
		such as CTRL-A.
```

In terminal normally pressing ctrl-[a-z] or ctrl-shift-[az] will produce the same code, that is the shift is ignored. Perhaps there are some terminals that will emit a different code, but gnome-terminal emits the same, you can try by running `cat` in terminal and pressing c-x and c-s-x, you'll get `^X` in both cases. However in gui mode for example nvim-qt, ctrl-shift registers as a different keymap, you can try it out with:
```
vim.keymap.set('n', '<c-X>', ':echo "c-X"<cr>')
vim.keymap.set('n', '<c-x>', ':echo "c-x"<cr>')
vim.keymap.set('n', '<c-s-x>', ':echo "c-s-x"<cr>')
```

However, you are right ctrl-t and ctrl-shift-t both work even in nvim-qt, which is supposed to differentiate between the two, why is that I have no idea.

In any case I agree with your change, I think it will be less confusing.

PR #721, opened 2024-03-07
Added a customizable highlight on yank. Color and duration can easily…

dezlymacauley:
I have changed the highlight on yank auto-command to be more customizable. 

Those who are new to Neovim will be easily able to change the color of the highlight (if they don't like the default orange flash).

They can also shorten or increase how long the highlight appears for.

I've been using this auto command for months now, without any bugs or clashes with UI plugins.


dam9000:
Hmm, this is replacing pure lua code with "lua calling vimscript calling lua", this does not seem too elegant to me. I think the project strives to be completely written in lua (or at least as much as possible). I doubt it will be accepted as is.

PR #728, opened 2024-03-08
Make telescope lazy loaded by command and key mappings

hronro:
Make telescope lazy loaded by command and key mappings.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -296,7 +296,6 @@ require('lazy').setup({
 
   { -- Fuzzy Finder (files, lsp, etc)
     'nvim-telescope/telescope.nvim',
-    event = 'VimEnter',
     branch = '0.1.x',
     dependencies = {
       'nvim-lua/plenary.nvim',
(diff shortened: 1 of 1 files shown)


feoh:
Hi. Why is this a desirable change please? Is it a performance enhancement?

Just trying to avoid change solely for the sake of yak shaving.

hronro:
The primary goal is to enhance Neovim's startup speed. The increasing popularity of the lazy.nvim plugin manager can be attributed to its feature that allows users to easily configure their plugins for lazy loading. This means that even if a user has installed numerous plugins, Neovim can still open instantly because only a few will load during startup, with others loading as needed. Since kickstart.nvim serves as a template for new users to create their own Neovim configurations, I think this is also a good chance to show them how to lazy load plugins by commands and key mappings.

In the current master branch, the telescope plugin loads immediately after Neovim starts up, which is unnecessary since it may not be needed right away. This PR ensures that telescope will only load when the `Telescope` command is executed or when users press the designated keys for it. While this change may only slightly improve startup time for this specific plugin, learning how to lazily load plugins from this example and applying it to all future additions could lead to significant improvements overall.

dam9000:
Technically the change is correct, one drawback is that it increases the line count, because all calls to telescope need to be wrapped in functions.One way around that could be to call `:Telescope` commands instead, but that's not the preferred lua way. The number of lines is not something that bothers me but TJ said on the he wants to keep the number of lines of code down. Anyway let's look at the actual performance numbers with `nvim --startuptime`:
```
times in msec

unchanged:

042.718  000.002: --- NVIM STARTED ---
044.463  000.002: --- NVIM STARTED ---
042.072  000.002: --- NVIM STARTED ---

lazy loaded telescope (this PR):

037.538  000.002: --- NVIM STARTED ---
035.488  000.002: --- NVIM STARTED ---
036.175  000.002: --- NVIM STARTED ---
```
let's say on average it takes 43 milliseconds to start with preloaded telescope and 36 millisec with lazy telescope, so an improvement of 7 milliseconds. The nvim is so fast anyway that the improvement will not really be noticed. Perhaps this improvement is noticeable when you have a large number of big plugins where each one adds a little bit up...

hronro:
The line count is increased because the formatter requires me to write `{ '<leader>sh', function() require('telescope.builtin').help_tags() end, desc = '[S]earch [H]elp' }` in 3 lines.

To avoid this, we can simply write a helper function for this:

```lua
{ -- Fuzzy Finder (files, lsp, etc)
  'nvim-telescope/telescope.nvim',

  keys= function()
    -- `builtin 'help_tags'` is the same as `require('telescope.builtin').help_tags`,
    -- but it does not need the 'telescope.builtin' module to be available when declaring the key mappings.
    function builtin(b)
      return function()
        require('telescope.builtin')[b]()
      end
    end

    return {
      { '<leader>sh', builtin 'help_tags', desc = '[S]earch [H]elp' },
      -- ......
    }
  end,
}
```

This will make the formatter happy and won't increase the line count.

If you agree, I can apply this change to the PR to decrease the line count.

dam9000:
@hronro that looks like a nice solution, it would probably need to be improved a bit to also accept the optional parameters  for some of the keymaps. In any case, it's not up to me, I'm just commenting. You'll need to get @feoh or @tjdevries approval.

hronro:
@dam9000 What do you mean by optional parameters for some of the keymaps? It should already cover all the key mappings defined for telescope in the master branch.

dam9000:
@hronro I mean for example:
```
require('telescope.builtin').find_files { cwd = vim.fn.stdpath 'config' }
```
you'll need to pass the parameter `{ cwd = vim.fn.stdpath 'config' }`
I guess something like this:
```
keys= function()
    -- `builtin 'help_tags'` is the same as `require('telescope.builtin').help_tags`,
    -- but it does not need the 'telescope.builtin' module to be available when declaring the key mappings.
    function builtin(b, opts)
      return function()
        require('telescope.builtin')[b](opts)
      end
    end
```
(not the added `opts`)
Also, I think the code would be clearer if the function `builtin` is renamed to `telescope_builtin`

tjdevries:
Thanks for the PR, but I think this makes it a lot less readable for newcomers to kickstart. Feel free to do this in your own config though, I think it's a nice change once people are more used to how lazy.nvim works!

PR #732, opened 2024-03-08
feat: allow treesitter default options to be overwritten by user preferences

rbaumgardner93:
If we move the default options for treesitter to opts you could then overwite them based on your preferences without altering the default config by doing something like this:

```lua
-- lua/custom/plugins/treesitter.lua

return {
  {
    opts = {
      -- add your own config options here
      ensure_installed = { 'css', 'javascript', 'styled' },
    }
  }
}
```

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -780,17 +780,18 @@ require('lazy').setup({
   { -- Highlight, edit, and navigate code
     'nvim-treesitter/nvim-treesitter',
     build = ':TSUpdate',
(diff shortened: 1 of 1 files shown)


ChillerDragon:
Thanks for this @rbaumgardner93 I was annoyed there was a git conflict in my config until I realized that this will be the last conflict. Cool change :heart:  

The example is also very helpful!

PR #754, opened 2024-03-12
Added folke/neodev.nvim for proper nvim api completion and annotation

dam9000:
Fixes nvim-lua/kickstart.nvim#692

`neodev` configures Lua LSP for your Neovim config, runtime and plugins used for completion, annotations and signatures of Neovim apis

With neodev, there's no more need to manually set lua_ls workspace settings which don't seem to work properly anyway as currently nvim api completion does not work.


dam9000:
Some screenshots, before:

Screenshot from 2024-03-12 20-34-19

after, with this fix:

Screenshot from 2024-03-12 20-39-13

feoh:
This smells like a clear win to me, AND a code line count reduction. Woot!

PR #759, opened 2024-03-14
Move friendly snippets to dependencies of LuaSnip

VlaDexa:
As the friendly-snippets docs say:
> Warning: If you're using LuaSnip make sure to use require("luasnip.loaders.from_vscode").lazy_load(), and add friendly-snippets as a dependency for LuaSnip, otherwise snippets might not be detected. If you don't use lazy_load() you might notice a slower startup-time

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -632,6 +632,13 @@ require('lazy').setup({
           end
           return 'make install_jsregexp'
         end)(),
+        dependencies = {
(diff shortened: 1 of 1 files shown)


VlaDexa:
Will make #680 work

tjdevries:
Oh, this is a great idea!! really nice thinking!

PR #764, opened 2024-03-16
Fix check health

relardev:
On nvim v0.9.5 when trying to run `:checkhealth` I kept getting lua errors from kickstart

1. the `vim.health....` functions should all start with `report_`, help results for `vim.health`: 
image

1. `vim.version.cmp` expects 2 strings but was getting 2 tables:
image

Not sure if this are correct fixes, this is what I had to do to make `:checkhealth` pass.

### DIFF
--- a/lua/kickstart/health.lua
+++ b/lua/kickstart/health.lua
@@ -8,14 +8,14 @@
 local check_version = function()
   local verstr = string.format('%s.%s.%s', vim.version().major, vim.version().minor, vim.version().patch)
   if not vim.version.cmp then
-    vim.health.error(string.format("Neovim out of date: '%s'. Upgrade to latest stable or nightly", verstr))
+    vim.health.report_error(string.format("Neovim out of date: '%s'. Upgrade to latest stable or nightly", verstr))
     return
   end
 
-  if vim.version.cmp(vim.version(), { 0, 9, 4 }) >= 0 then
-    vim.health.ok(string.format("Neovim version is: '%s'", verstr))
+  if vim.version.cmp(verstr, '0.9.4') >= 0 then
+    vim.health.report_ok(string.format("Neovim version is: '%s'", verstr))
   else
-    vim.health.error(string.format("Neovim out of date: '%s'. Upgrade to latest stable or nightly", verstr))
+    vim.health.report_error(string.format("Neovim out of date: '%s'. Upgrade to latest stable or nightly", verstr))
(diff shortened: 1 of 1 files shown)


dam9000:
@relardev what kind of lua errors are you getting? Can you post a log or screenshot of the errors?
I also have nvim 0.9.5 and I'm not seeing any errors:
```
kickstart: require("kickstart.health").check()

kickstart.nvim ~
- NOTE: Not every warning is a 'must-fix' in `:checkhealth`
  
  Fix only warnings for plugins and languages you intend to use.
  Mason will give warnings for languages that are not installed.
  You do not need to install, unless you want to use those languages!
- System Information: {
  machine = "x86_64",
  release = "6.7.6-200.fc39.x86_64",
  sysname = "Linux",
  version = "#1 SMP PREEMPT_DYNAMIC Fri Feb 23 18:27:29 UTC 2024"
  }
- OK Neovim version is: '0.9.5'
- OK Found executable: 'git'
- OK Found executable: 'make'
- OK Found executable: 'unzip'
- OK Found executable: 'rg'
```

dam9000:
Also in health.lua there is:
```
function M.start(msg)
  M.report_start(msg)
end
function M.info(msg)
  M.report_info(msg)
end
function M.ok(msg)
  M.report_ok(msg)
end
function M.warn(msg, ...)
  M.report_warn(msg, ...)
end
function M.error(msg, ...)
  M.report_error(msg, ...)
end
```
So the code as is should work as is, although, I guess the shorthand functions were added later and if we would want to report an error properly for older versions, the code should indeed be adapted to the old APIs. Buf this further points to your version being older than 0.9.5.

relardev:
```
> nvim --version

NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1692716794

  fall-back for $VIM: "/__w/neovim/neovim/build/nvim.AppDir/usr/share/nvim"

```

dam9000:
@relardev did you build the nvim manually or is it installed from a package? Is it possible you have both - a system installed nvim and a local build and the runtimes clash? Which OS do you use?

relardev:
I updated the nvim today from 0.9.1 by moving new binary into old one(previously I updated it the same way from who knows what version), but didn't touch the share files, didn't know they need to be updated too. My expectation was that the neovim binary has all that it needs. Sorry for fake pr, at least i learned something

Issue #765, opened 2024-03-16
zsh: command not found: nvim


elieven:
## Describe the bug

To the best of my knowledge I'm quite sure I did my best to do exactly what the readme specified, installed all the requirements (in wsl ubuntu) and ran the command. It cloned the repo and after that nothing. Trying to run nvim as specified just fails as it cannot find it. Also I restarted and still same. Am I missing something obvious?

```
[ 11:36PM ]  [ elieven@DESKTOP-I52EQ2M:~/.config ]
 $ git clone https://github.com/nvim-lua/kickstart.nvim.git "${XDG_CONFIG_HOME:-$HOME/.config}"/nvim
zsh: correct 'clone' to 'rclone' [nyae]? n
Cloning into '/home/elieven/.config/nvim'...
remote: Enumerating objects: 884, done.
remote: Counting objects: 100% (19/19), done.
remote: Compressing objects: 100% (16/16), done.
remote: Total 884 (delta 9), reused 8 (delta 3), pack-reused 865
Receiving objects: 100% (884/884), 423.58 KiB | 2.46 MiB/s, done.
Resolving deltas: 100% (442/442), done.
[ 11:36PM ]  [ elieven@DESKTOP-I52EQ2M:~/.config ]
 $ nvim
[ 11:36PM ]  [ elieven@DESKTOP-I52EQ2M:~/.config/nvim(master✔) ]
 $ nvim
zsh: correct 'nvim' to 'vim' [nyae]? n
zsh: command not found: nvim
```

## To Reproduce
- Install dependencies mentioned
- Run the git clone command as specified
- Try to run `nvim`

## Desktop

- OS: Ubuntu (Windows wsl)
- Terminal: Oh my zsh

## Neovim Version

Can't get nvim to run so...

```
```

dam9000:
It appears you have not installed the actual nvim.
Here, this should help:

```
wsl --install
wsl
sudo add-apt-repository ppa:neovim-ppa/unstable -y
sudo apt update
sudo apt install neovim make gcc ripgrep unzip
git clone https://github.com/nvim-lua/kickstart.nvim.git "$HOME/.config/nvim"
nvim
```

elieven:
[...]

PR #767, opened 2024-03-17
README: additional install recipes for various OS

dam9000:
I'm noticing in many opened issues people have problem installing all the proper dependencies or the right neovim version. So I've added more detailed OS-specific instructions if this would perhaps help more people.
I don't have access to MacOS, os only Windows and some Linux variants are covered.

Example issue: https://github.com/nvim-lua/kickstart.nvim/issues/765

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,8 @@ External Requirements:
   - etc.
 
 > **NOTE**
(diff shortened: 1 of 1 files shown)


feoh:
Great work, thanks for this!

Just tried the chocolatey route on Windows with a fresh install & it works great!

PR #771, opened 2024-03-18
Some suggestions and capitalised a few words

togglebyte:
I read through this (very nice btw) and went through and capitalised a few words and added some punctuation etc. Feel free to ignore it and just close it if it's not relevant.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -786,8 +786,7 @@ require('lazy').setup({
       -- ... and there is more!
       --  Check out: https://github.com/echasnovski/mini.nvim
     end,
-  },docker
-
+  },
   { -- Highlight, edit, and navigate code
     'nvim-treesitter/nvim-treesitter',
     build = ':TSUpdate',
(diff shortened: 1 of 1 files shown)


feoh:
One thing though - you may notice that one of the checks failed. This is because we now require code changes to be formatted with the Stylua tool.

Let me know if you need help running your changes through.

init.lua
```diff
@@ -787,7 +786,7 @@ require('lazy').setup({
       -- ... and there is more!
       --  Check out: https://github.com/echasnovski/mini.nvim
     end,
-  },
+  },docker
```
dam9000, line 789: The added `docker` here looks like a mistake and causes stylua to complain and also running nvim with this config gives an error:
```
Error detected while processing .../init.lua:
E5113: Error while calling lua chunk: .../init.lua:791:
  attempt to call global 'docker' (a nil value)
```

togglebyte: Cripes, that must've been an accidental paste or such.
Fixed!

PR #774, opened 2024-03-18
Add nvim-nio as dependency for nvim-dap-ui

alosarjos:
It's a dependency now

https://github.com/rcarriga/nvim-dap-ui/commit/f7d75cca202b52a60c520ec7b1ec3414d6e77b0f

### DIFF
--- a/lua/kickstart/plugins/debug.lua
+++ b/lua/kickstart/plugins/debug.lua
@@ -14,6 +14,9 @@ return {
     -- Creates a beautiful debugger UI
     'rcarriga/nvim-dap-ui',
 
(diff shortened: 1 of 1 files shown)


feoh:
Hi!

Shouldn't this be specified using Lazy's dependencies = clause like we do [here](https://github.com/nvim-lua/kickstart.nvim/blob/65a5ac404b56c4718d79f65ac642e19e89346eda/init.lua#L293) ?

Thanks for your contribution!

feoh:
Don't mind me I misread. You're all good. :) Thanks for the contribution. Merging.

PR #779, opened 2024-03-20
Add gitsigns essential keymaps

dam9000:
An alternative attempt at including only the essential keymaps compared to the full list in: https://github.com/nvim-lua/kickstart.nvim/pull/740
Changes compared to the full list:
- reduced the list of keymaps as suggested by @lewis6991 (the author of gitsigns):
  https://github.com/nvim-lua/kickstart.nvim/pull/740#pullrequestreview-1948295290
- simplified navigation keymap code - verified that it works also in diff mode, the only difference is a slight change in behaviour as gitsigns navigation does wrap-around while original diff mode does not. Also confirmed with @lewis6991 that it should work.
- Added a note pointing to gitsigns README for the full list of keymaps

Compared to the original PR this one has about half the lines of code, hopefully this will make it more likely to be included.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -255,6 +255,33 @@ require('lazy').setup({
         topdelete = { text = '‾' },
         changedelete = { text = '~' },
       },
+      on_attach = function(bufnr)
+        -- Include essential gitsigns keymaps, for more see gitsigns README:
+        -- https://github.com/lewis6991/gitsigns.nvim
+        local gs = package.loaded.gitsigns
+        local function map(mode, l, r, opts)
+          opts = opts or {}
+          opts.buffer = bufnr
+          vim.keymap.set(mode, l, r, opts)
+        end
+        -- Navigation
+        map({ 'n', 'v' }, ']c', gs.next_hunk, { desc = 'Jump to next git [c]hange' })
+        map({ 'n', 'v' }, '[c', gs.prev_hunk, { desc = 'Jump to previous git [c]hange' })
+        -- Actions
+        map('n', '<leader>hs', gs.stage_hunk, { desc = 'git [s]tage hunk' })
+        map('n', '<leader>hr', gs.reset_hunk, { desc = 'git [r]eset hunk' })
+        map('n', '<leader>hu', gs.undo_stage_hunk, { desc = 'git [u]ndo stage hunk' })
+        map('n', '<leader>hp', gs.preview_hunk, { desc = 'git [p]review hunk' })
+        map('n', '<leader>hb', gs.blame_line, { desc = 'git [b]lame line' })
+        map('n', '<leader>hd', gs.diffthis, { desc = 'git [d]iff against index' })
+        -- Visual mode
+        map('v', '<leader>hs', function()
+          gs.stage_hunk { vim.fn.line '.', vim.fn.line 'v' }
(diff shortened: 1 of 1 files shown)


feoh:
LGTM but I'll defer to @tjdevries as I haven't had a problem with this stuff from jump, but can appreciate the desire to keep things small and understandable.

feoh:
Hi!

Just wanted to mention what I did and ask if this might be a good option for you:

https://github.com/feoh/kickstart.nvim/commit/52722b7524b152fb77772ad968a54fc56a51ff50

If I install gitsigns like a regular old plugin, I get it in all its full fat goodness. Keymaps, configs. The whole enchilada.

Why not just do that? What am I missing? That way @tjdevries can get his super minimal starting point for new folks, and we can enjoy our luxurious git manipulating goodness in our forks? :)

Sorry if I'm missing something obvious, and in NO way do I mean to cut down the value of all your amazing contributions and work on this project.

You really have been an outstanding contributor, and I for one *sincerely* appreciate it!

Should you ever come to Boston, MA, USA. Please do look me up! I'd love to buy you a beverage or dinner or whatever :)

dam9000:
@feoh your suggestion works. But I think it's not clean that a plugin has the spec split into two files. Please open a PR with your solution so there are more options for TJ to pick from. It's not a question what works for *me*, I have the full enchilada no worries, the question is what is the best for the average or new Neovim user. I think a new Neovim user isn't necessarily a fresh developer, he can be an experienced developer looking for a new editor. A perfect example: DHH moving to Neovim: https://twitter.com/dhh/status/1764340531877105824

Anyway let's wait for @tjdevries feedback and we'll see.

And thanks for the kind words, I have no travel plans to US at the moment, but I'll keep it in mind :)

init.lua
```diff
@@ -286,7 +313,13 @@ require('lazy').setup({
         ['<leader>r'] = { name = '[R]ename', _ = 'which_key_ignore' },
         ['<leader>s'] = { name = '[S]earch', _ = 'which_key_ignore' },
         ['<leader>w'] = { name = '[W]orkspace', _ = 'which_key_ignore' },
+        ['<leader>h'] = { name = 'Git [H]unk', _ = 'which_key_ignore' },
+        ['<leader>t'] = { name = '[T]oggle', _ = 'which_key_ignore' },
```
ro0gr, line 317: This seems unused? I can't see any new `T` maps in the commit.

dam9000: done, force pushed removal of toggle.

feoh:
Just coming back to this in my usual PR check:

I don't know that generating a PR of my change is appropriate because my change has the gitsigns plugin installed in lua/custom/plugins which I think we'd prefer to reserve for end users to fill themselves :)

SebasF1349:
Gitsigns has just updated and `next_hunk()` and `prev_hunk()` has been deprecated. This is the commit with the new suggested way to do keymaps: [https://github.com/lewis6991/gitsigns.nvim/commit/59bdc1851c7aba8a86ded87fe075ef6de499045c](https://github.com/lewis6991/gitsigns.nvim/commit/59bdc1851c7aba8a86ded87fe075ef6de499045c)

feoh:
Closed in favor of #858. Thanks for the hard work on this!

PR #781, opened 2024-03-20
Update README.md

bozoputer:
The recommended step of forking the repo coming sequentially after the step instructing users to clone the current repo doesn't make sense.

This commit orders the install instructions in a manner that's more logical.


feoh:
Good catch. Thanks for your contribution!

PR #814, opened 2024-03-31
Add LSP keymap to format current buffer

GCrispino:
Having a key map to format the current buffer seems like something that is useful and general enough to have in the kickstart config, so I'm proposing this addition here

Feel free to close if you think that this is not true, though 🙏🏼

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -501,6 +501,11 @@ require('lazy').setup({
           --  For example, in C this would take you to the header.
           map('gD', vim.lsp.buf.declaration, '[G]oto [D]eclaration')
 
+          -- Formats the current buffer
+          map('<space>f', function()
+            vim.lsp.buf.format { async = true }
+          end, '[F]ormat code')
+
           -- The following two autocommands are used to highlight references of the
           -- word under your cursor when your cursor rests there for a little while.
           --    See `:help CursorHold` for information about when this is executed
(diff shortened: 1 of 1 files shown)


feoh:
What do folks think? I might find this useful.

GCrispino:
To add some more context, I'm coming from a setup of vim + coc.nvim, whose default example configuration exposes a "Format" command for doing this, and was trying to find if I could do the same thing after setting kickstart up

init.lua
```diff
@@ -501,6 +501,11 @@ require('lazy').setup({
           --  For example, in C this would take you to the header.
           map('gD', vim.lsp.buf.declaration, '[G]oto [D]eclaration')
 
+          -- Formats the current buffer
+          map('<space>f', function()
```
lazyguru, line 505: Should this be `<leader>` instead of `<space>`? It might be the same at the moment, but if someone changes leader it might be confusing as to why formatting "stopped working"

GCrispino: Good point :). Updated 👍🏼

dam9000:
In my view it would be better to use `conform` which is already included and does lsp fallback and also allows for visual range formatting. In fact that is what I use, I can open a PR with that if there is interest.

dam9000:
Well here is the alternative using conform:
https://github.com/nvim-lua/kickstart.nvim/pull/817

Kickstart is already using conform for formatting and in lua case the stylua formatting differs from LSP formatter. So using this will be more consistent.

feoh:
Closing in favor of #817 @GCrispino thanks for the suggestion. I hope you don't mind that we went with @dam9000 's implementation. We introduced conform in the rewrite for exactly this kind of thing :)

GCrispino:
@feoh no problem. Thanks all for considering this :)

PR #830, opened 2024-04-07
README: add git install to "Ubuntu Install Steps"

hoangvietdo:
I included the Git installation in the 'Ubuntu Install Steps' to facilitate users when attempting to clone the kickstart repository.


feoh:
Hi. This also contains changes to init.lua which probably shouldn't get merged.

dam9000:
This needs to be squashed so it's a single commit without extra changes, use force push to correct.
Can be done by @feoh as well, but it's nicer if the submitter takes the effort to clean up a PR.

PR #831, opened 2024-04-07
Move plugin examples from README to optional plugin files

dam9000:
Proposal: move examples on how to add popular plugins (autopairs and neo-tree) from README
to optional commented out plugin files in the same way as other plugins such as debug
and indent_line are done. The information on how to add plugins is already inside the
init.lua so perhaps there's no need to duplicate that in the README.
That way README can focus on installation while the init.lua can focus
on configuration. Making these popular plugins optional will make it easier
to enable them with less friction.

The plugins received minor modification: both were made to be lazy loaded and a
default keymap added for neo-tree which is recommended in the neo-tree README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,40 +99,9 @@ That's it! Lazy will install all the plugins you have. Use `:Lazy` to view
 current plugin status. Hit `q` to close the window.
 
 Read through the `init.lua` file in your configuration folder for more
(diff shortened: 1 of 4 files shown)


feoh (approved):
Love this! Grat work!

dam9000:
@feoh any plan to merge this?

PR #843, opened 2024-04-14
Enable inlay hints for the supporting servers

VlaDexa:
Since latest neovim nightly supports inlay hints maybe you should look into enabling them by default, like vscode does.

This can be an unwanted change for some people, since the implemented hints are in the middle of the code and I know some people that don't like that.

Maybe that can be left in, but commented out so the users may know that this feature is supported, just not enabled by default.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -286,6 +286,7 @@ require('lazy').setup({
         ['<leader>r'] = { name = '[R]ename', _ = 'which_key_ignore' },
         ['<leader>s'] = { name = '[S]earch', _ = 'which_key_ignore' },
         ['<leader>w'] = { name = '[W]orkspace', _ = 'which_key_ignore' },
+        ['<leader>t'] = { name = '[T]oggle', _ = 'which_key_ignore' },
       }
     end,
   },
--- a/init.lua
+++ b/init.lua
@@ -518,6 +518,16 @@ require('lazy').setup({
               callback = vim.lsp.buf.clear_references,
             })
           end
+
+          -- The following autocommand is used to enable inlay hints in your
+          -- code, if the language server you are using supports them
+          --
+          -- This may be unwanted, since they displace some of your code
+          if client and client.server_capabilities.inlayHintProvider and vim.lsp.inlay_hint then
+            map('<leader>th', function()
+              vim.lsp.inlay_hint.enable(0, not vim.lsp.inlay_hint.is_enabled())
+            end, '[T]oggle Inlay [H]ints')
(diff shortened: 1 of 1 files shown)


WindSoilder:
I think it's good to have a way to toggle inlay hint, because it can be too verbosely

feoh:
Agreed. A toggle or configurable is needed here.

VlaDexa:
Maybe since this affects code stuff and uses LSP it should be <leader>ci?

feoh:
@VlaDexa Thanks for your work thus far. If you're still intrerested in getting this merged, please add a toggle as folks have detailed. I don't honestly care which key bind you use. People can always change it :)

WindSoilder:
Sorry to have more thought here: I don't think `<leader>ci` match the command naming style in kickstart, because `c` doesn't show in help message.

For example:
```lua
map('<leader>ca', vim.lsp.buf.code_action, '[C]ode [A]ction')
map('<leader>rn', vim.lsp.buf.rename, '[R]e[n]ame')
map('<leader>ds', require('telescope.builtin').lsp_document_symbols, '[D]ocument [S]ymbols')
```
Every character has a highlight in help message(which wrapped with `[]`)

feoh:
More thoughts are good so long as the contributor of said thoughts helps find concordance and we can move forward, assuming folks agree that we all want this optional feature.

dam9000:
`<leader>ti` `[T]oggle [I]nlay` still makes the most sense to me :)

VlaDexa:
If we're going by the starting letters of the help message logic, then sure, I agree that `ci` is not conforming.  Then, I think that it being `<leader>th` is easier to remember, since you don't need to know that it toggles something inlay, but enables hints.

dam9000:
`<leader>th` is fine with me :+1:

init.lua
```diff
@@ -286,6 +286,7 @@ require('lazy').setup({
         ['<leader>r'] = { name = '[R]ename', _ = 'which_key_ignore' },
         ['<leader>s'] = { name = '[S]earch', _ = 'which_key_ignore' },
         ['<leader>w'] = { name = '[W]orkspace', _ = 'which_key_ignore' },
+        ['<leader>T'] = { name = '[T]oggle', _ = 'which_key_ignore' },
```
dam9000, line 289: `<leader>T` should be lower case here so `<leader>t` to match the `<leader>th` below

VlaDexa: Thanks!

Issue #849, opened 2024-04-15
which-key bug preventing normal key function


BirdeeHub:
## Describe the bug

On fresh install of kickstart, pressing ```"``` key causes this error, and cancels operator pending
```
E5107: Error loading lua [string ":lua"]:1: ')' expected near '", {mode = "'
```
This is because of the which-key plugins.registers option

Also on a fresh install, pressing ```<C-w>``` will not work.

Instead, it will show ```<80> ^DW``` in the bottom right key pending display, and then pretends the key you pressed was the number 23

This is because of the which-key plugins.presets.windows option

## To Reproduce

1. go through setup instructions for kickstart.nvim on any linux system (have tried on ubuntu, kali, arch, and nixos)
2. press ```"``` or ```<c-w>```
3. observe above behavior

To "fix"
```lua
      require('which-key').setup {
        -- some features of which key prevent normal functioning of the key
        -- this is a bug, so we will disable the problematic features.
        -- which-key is very useful for those new to vim motions, so it is still included.
        plugins = {
          -- this will currently prevent the " keybind from working in normal mode.
          registers = false, -- shows your registers on " in NORMAL or <C-r> in INSERT mode
          presets = {
            -- this will currently prevent the <C-w> keybinds from working in normal mode.
            windows = false, -- default bindings on <c-w>
          },
        },
      }
```

## Desktop

- OS: ubuntu, kali, arch, nixos on WSL, bare metal, or vm
- Terminal: alacritty (I also have experienced this bug on xfce4-terminal and windows terminal, I have not tried any others for this bug)

## Neovim Version

Keep in mind, this has been a bug since at least 0.9, but here is my current version

```
NVIM v0.10.0-dev
Build type: Release
LuaJIT 2.1.1693350652
Run ":verbose version" for more info
```

dam9000:
This works fine for me (on fedora, nvim 0.9.5). I will try other platforms later.
Are you sure this isn't an issue with your setup?

dam9000:
I made some more tests on various platforms and it seems to work fine for me everywhere.
I think this must be some odd config or install issue on your end.
Here are the tests for the `"` in:

ubuntu 22.04 nvim 0.10 alacritty:
![nvim0100-ubuntu2204-alacritty](https://github.com/nvim-lua/kickstart.nvim/assets/148590335/48f967c1-8c27-408e-ac80-8c771b17a478)

ubuntu 22.04 nvim 0.10 gnome-terminal:
[...]

PR #850, opened 2024-04-15
disabled problematic which-key features

BirdeeHub:
addresses issue https://github.com/nvim-lua/kickstart.nvim/issues/849

Edit: I probably shoulda not ammended the 2 formatting edits together, it would have been less confusing without doing that. I was just fixing formatting to satisfy stylua

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -277,7 +277,19 @@ require('lazy').setup({
     'folke/which-key.nvim',
     event = 'VimEnter', -- Sets the loading event to 'VimEnter'
     config = function() -- This is the function that runs, AFTER loading
-      require('which-key').setup()
+      require('which-key').setup({
+        -- some features of which key prevent normal functioning of the key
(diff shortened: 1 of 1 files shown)


dam9000:
This shouldn't be merged before understanding better what the issue is. Because from my testing of the issue #849 the which-key behaviour works fine as is. So I suspect this to be a specific config or install issue.

BirdeeHub:
Ill have to get back to you on this. I just tried an nvim with ONLY which-key installed and it worked???? But its been happening with just the kickstart config and across multiple distros, Ill let you know when I find out specifically what the issue is.

Just for my sanity im going to set up an entirely new ubuntu vm from scratch and try it...

BirdeeHub:
Oh also I figured out what the issue is a while back

I skill issued

vim.opt.cpoptions:append('I')

is what I needed

vim.o.cpoptions = 'I'

Is what I had

I just remembered about this

PR #852, opened 2024-04-15
Arch, btw

rdvm:


dam9000:
I assume `unzip` is already installed by default? Perhaps it's best to make it explicit and add unzip to the list in case someone starts from some kind of minimal install if such exists for arch, perhaps some minimal docker image... other than that LGTM.

dam9000:
Right, I did not include unzip for Fedora because it is in the default set but let's make it explicit for Fedora too. I can confirm that the package name on Fedora is `unzip`:
```
# dnf install unzip
Last metadata expiration check: 2:02:18 ago on Tue 16 Apr 2024 02:27:33 PM CEST.
Package unzip-6.0-62.fc39.x86_64 is already installed.
Dependencies resolved.
Nothing to do.
Complete!
```
So please go ahead and fix also Fedora entry, thanks.

rdvm:
I added `unzip` for Fedora and also added the `--needed` option for Arch so it won't needlessly reinstall any of these packages if already present.

dam9000 (approved):
LGTM, thanks.

feoh:
Thanks everyone. Great additions!

Issue #847, opened 2024-04-14
Treesitter: Error during download, please verify your internet connection


nfgferreira:
## Describe the bug
The treesitter instalation gave the error above for me.

## To Reproduce
Install Kickstart and run Neovim

## Desktop
- OS: WSL, Ubuntu
- Terminal: Stock terminal

## Neovim Version
0.9.5

## Additional notes

The problem was fixed for me after adding `require("nvim-treesitter.install").prefer_git` = true as described here: [https://github.com/nvim-treesitter/nvim-treesitter/issues/3232]

I suspect that the issue may have happened because I am using a corporate PC with some network access limitations. I think it is worth mentioning in the good description provided by the init.lua file.

feoh:
Hi guys. This could be fixed if we add prefer_git as per [this treesitter issue](https://github.com/nvim-treesitter/nvim-treesitter/issues/3232).

```lua
require("nvim-treesitter.install").prefer_git = true
```

Want me to generate a pull request? I'd fixed it this way locally but didn't know how well the change would be received so I didn't PR.

Also, I seem to have lost the fix in my local fork. I sometimes wonder how much churn and lost code I suffer from just maintaining the silly fork :)

dam9000:
@feoh you opened a similar PR already in the past:
https://github.com/nvim-lua/kickstart.nvim/pull/654

But the comment was wrong because it implied precompiled binaries, so removing the mention of precompiled binaries and opening a new PR sound like a good idea.

feoh:
OK. I'm honestly not entirely sure how to translate that syntax into lazy.nvim

I *THINK* maybe I just add `prefer_git = true` to the opts = section?

dam9000:
@feoh I checked the treesitter code and it appears that the prefer_git is not a global treesitter cofig option but a local to the `nvim-treesitter.install` module, so using lazy `opts` will not work. This should work:
```
diff --git a/init.lua b/init.lua
index 496f45b..a3128a1 100644
--- a/init.lua
+++ b/init.lua
@@ -817,6 +817,7 @@ require('lazy').setup({
     config = function(_, opts)
       -- [[ Configure Treesitter ]] See `:help nvim-treesitter`
 
+      require('nvim-treesitter.install').prefer_git = true
       ---@diagnostic disable-next-line: missing-fields
       require('nvim-treesitter.configs').setup(opts)
```
I tested it with:
```
:lua vim.print(require('nvim-treesitter.install').prefer_git)
[...]

PR #856, opened 2024-04-17
Fix: #847 - add prefer_git to treesitter config

feoh:
This PR configures Treesitter to prefer git rather than expecting curl to be installed. This prevevents a failure mode on WSL where curl is optional.

(Doing this small PR on the main fork because my fork is far too far diverged to be able to source a minor change like this, an y'all don't want my 9000 partially blind guy enhancements :)

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -817,6 +817,7 @@ require('lazy').setup({
     config = function(_, opts)
       -- [[ Configure Treesitter ]] See `:help nvim-treesitter`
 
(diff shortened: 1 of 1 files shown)


dam9000:
Perhaps add a comment such as:
```
-- Prefer git instead of curl in order to improve connectivity in some environments
```

feoh:
Done. Great suggestion. Thanks.

dam9000 (approved):

PR #858, opened 2024-04-17
Add gitsigns recommended keymaps as an optional plugin

dam9000:
Add gitsigns recommended keymaps as an optional plugin

Compared to previous PR: https://github.com/nvim-lua/kickstart.nvim/pull/740 this also updates the code to match the latest example in gitsigns README:
- changed `gs` to `gitsigns`
- changed the navigation to the new `nav_hunk()` api

Note, the which-key leader keys are still inside the main init.lua, hope that's fine, if not that can also be moved.


feoh:
Well done and thank you! Great to make this functionality available to folks!

Issue #859, opened 2024-04-18
Debian install procedure has you changing the entire distribution to unstable to install Neovim


EvanCarroll:
Currently the [documentation for Debian](https://github.com/nvim-lua/kickstart.nvim/blame/master/README.md#L196) instructs you to do this,

```sh
sudo apt update
sudo apt install make gcc ripgrep unzip git
echo "deb https://deb.debian.org/debian unstable main" | sudo tee -a /etc/apt/sources.list
sudo apt update
sudo apt install -t unstable neovim
```

The problem is this line,

```
echo "deb https://deb.debian.org/debian unstable main" | sudo tee -a /etc/apt/sources.list
```

That effectively moves you to the unstable version of debian. If you run `apt-get dist-upgrade` and download all of your packages, and then you run the above code and you rerun `apt-get dist-upgrade` you will be on the unstable version of Debian.

This is probably not what you want.

The problem that you're trying to solve is that Debian stable (right now "bookworm") ships Neovim 0.7.2. Unfortunately, there is no newer version of Neovim in the repository, nor in the backports repository (because of a shortage of manpower). However, upgrading your entire distribution to unstable is a very poor suggestion that can cause all kinds of problems and pain, including total failure.

**At the very least, the user should be informed**

EvanCarroll:
This edit was done last month by @dam9000 https://github.com/nvim-lua/kickstart.nvim/commit/a222805c6120b88f451c3e026b29ea6f761a491e

A better way for the Linux's is probably just to use upstream neovim's prebundled binary if that works for you, https://github.com/neovim/neovim/blob/master/INSTALL.md#pre-built-archives-2

feoh:
@EvanCarroll I see what you mean.

As an interim step, would you or anyone commenting here please consider updating the verbiage in the install instructions to warn appropriately?

A problem I see is that in order to update Debian to the point where you can *build* Neovim, you kind of have to go unstable.

Perhaps simply stating that up front and offering the pre-built binary as an alternative is the way to go here? I'll leave it to the Debian folks to craft an appropriate and best practice conformant set of install instructions.

EvanCarroll:
> A problem I see is that in order to update Debian to the point where you can build Neovim, you kind of have to go unstable.

I don't believe this is true. I don't know why specifically you couldn't build Neovim 0.9 on Debian stable (bookworm).

> Debian folks to craft an appropriate and best practice

[...]

PR #860, opened 2024-04-18
Update README.md

EvanCarroll:
Attempted fix for #859, provide reasonable Debian install instructions -- comment on GitHub issue with refinement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,9 +200,14 @@ sudo apt install make gcc ripgrep unzip git neovim
 ```
 sudo apt update
 sudo apt install make gcc ripgrep unzip git
-echo "deb https://deb.debian.org/debian unstable main" | sudo tee -a /etc/apt/sources.list
-sudo apt update
-sudo apt install -t unstable neovim
+
+# Now we install nvim
+curl -LO https://github.com/neovim/neovim/releases/latest/download/nvim-linux64.tar.gz
+sudo rm -rf /opt/nvim
+sudo tar -C /opt -xzf nvim-linux64.tar.gz
+
+# make it available in /usr/local/bin, distro installs to /usr/bin
+sudo ln -sf /opt/nvim-linux64/bin/nvim /usr/local/bin/
 ```
 </details>
 <details><summary>Fedora Install Steps</summary>


README.md
```diff
@@ -200,9 +200,14 @@ sudo apt install make gcc ripgrep unzip git neovim
 ```
 sudo apt update
 sudo apt install make gcc ripgrep unzip git
-echo "deb https://deb.debian.org/debian unstable main" | sudo tee -a /etc/apt/sources.list
-sudo apt update
-sudo apt install -t unstable neovim
+
+# Now we install nvim
+curl -LO https://github.com/neovim/neovim/releases/latest/download/nvim-linux64.tar.gz
+sudo rm -rf /opt/nvim
```
dam9000, line 206: The destination is `/opt/nvim-linux64` not `/opt/nvim` is `/opt/nvim` here intentional or did you mean `/opt/nvim-linux64` ?

README.md
```diff
@@ -200,9 +200,14 @@ sudo apt install make gcc ripgrep unzip git neovim
...
+sudo tar -C /opt -xzf nvim-linux64.tar.gz
```
dam9000, line 207: when testing this I noticed that `/opt/nvim-linux64` was inaccessible to my user because of umask 027 and the `nvim-linux64` itself is not in the archive so tar would not set it's permissions, so perhaps this needs to be added:
```
sudo mkdir -p /opt/nvim-linux64
sudo chmod a+rx /opt/nvim-linux64
```

README.md
EvanCarroll: I just copied it from neovim upstream, https://github.com/neovim/neovim/blob/master/INSTALL.md#pre-built-archives-2 I have no idea why they nuke that path, but it sounds like it's innocuous to me. nvim upstream should own `/opt/nvim*` as an unmodified install location. So if they want it gone if it's there, I'm cool with killing it.  All that said I left a comment on the commit https://github.com/neovim/neovim/commit/5c36701228e166f8aa89983044404ff322fe481f#r141101101

README.md
EvanCarroll: Or just put it in a subshell `(umask 022; sudo tar...)` , i mean it's not a hard workaround but you're going to be changing every guide online for it.

if you set `0277` you don't want things installed as root run by non-root, that's what you're saying. some people like that. if so, you could install it to `~/opt/` and symlink to `~/.bin/` rather thank `/usr/local/bin`.

README.md
dam9000: got it, ok.

README.md
dam9000: That's not what I'm saying, a user having a umask  027 intention is that the user's files are not readable by other users by default. I think that's not an uncommon setting. That doesn't translate to the intention of root installed packages not being able to be used by users. If the `/opt/nvim-linux64` was a part of the tar package this would not be an issue because tar will preserve permissions but it is not, only the subdirs are part of the tar package.
Anyway, this is just a suggestion to make it less error prone for a wider audience, feel free to ignore it and resolve this comment if you think the current behaviour is preferred.

README.md
```diff
@@ -200,9 +200,14 @@ sudo apt install make gcc ripgrep unzip git neovim
 ```
 sudo apt update
 sudo apt install make gcc ripgrep unzip git
-echo "deb https://deb.debian.org/debian unstable main" | sudo tee -a /etc/apt/sources.list
-sudo apt update
-sudo apt install -t unstable neovim
+
+# Now we install nvim
+curl -LO https://github.com/neovim/neovim/releases/latest/download/nvim-linux64.tar.gz
```
dam9000, line 205: On a fresh debian install I get an error that curl is not installed:
```
vboxuser@debian125:~$ curl -LO https://github.com/neovim/neovim/releases/latest/download/nvim-linux64.tar.gz
bash: curl: command not found
```
so please add also `curl` to the above list of dependencies:
```
sudo apt install make gcc ripgrep unzip git curl
```

dam9000 (approved):
@feoh let's merge this, my comments can be addressed later.

PR #863, opened 2024-04-18
I have implemented a simple way to toggle format on save

daniele821:
This is a very simple way to disable format on save functionality provided by `Conform`.  I think it can be very useful, as there can be some files where for you may not want to have it be automatically formatted. 

Note that `<leader>f` still always work, even if format on save is disabled.

I believe it is also educational, as it shows how to create a user command, which I didn't find anywhere in the entire repo.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -630,6 +630,10 @@ require('lazy').setup({
     opts = {
       notify_on_error = false,
       format_on_save = function(bufnr)
+        -- Do not format on save if you run :KickstartToggleFormatOnSave
+        if vim.g.kickstart_disable_format_on_save then
+          return
+        end
         -- Disable "format_on_save lsp_fallback" for languages that don't
         -- have a well standardized coding style. You can add additional
         -- languages here or re-enable it for the disabled ones.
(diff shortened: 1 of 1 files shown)


feoh:
I don't know if this is desirable. You're welcome to create a form and implement this.

daniele821:
i already did. Did i post this pull request wrong?

If so sorry, it's my first pull request.

I implemented this, as I use it myself, and I think it can be useful to be able to disable autoformatting when you save. I though it could be useful to have this, as suppose you have a file with different formatting for example where you only want to make a simple edit.

feoh:
Please don't apologize!

It's just that it seems to me you're doing this with a global configurable variable and I'm not sure if that's the solution the community would want for this.

I'll let others comment on their opinions. Thank you for your contribution in any case!

daniele821:
well you need a global variable, since the command and the usage are in different contexts, and i don't see different ways of doing it.

I can change to use a buffer scoped variable instead, if it's better. But then you would the autoformatting enabled in some file, and not in others, which could become difficult to keep track of.

feoh:
@dam9000 What do you think of this one?  Does it need to be keystroke toggle-able? Is this something we want?

dam9000:
@feoh 
Toggling the format-on-save is one of the many recipes provided by the conform docs:
https://github.com/stevearc/conform.nvim/blob/master/doc/recipes.md#command-to-toggle-format-on-save

We have included one of the recipes  - the "Disable auto format on certain filetypes" (for C and C++)
Which is short and was in my view essential as it also changed from a fixed config (table) to a dynamic configuration (function) and thus gave a hint on where and how to extend and customize that behaviour further.

Including all the possible recipes would be too long and I think this is now in the personal customization territory, not something that a majority of people will need. If someone needs such a fine grained control on the auto format on save, perhaps he might consider disabling it by default and use the provided keymap `<leader>f` to auto-format the buffer on demand. Or setting per-project (folder) behaviour, something that is also provided in one of the recipes.

So from my view this is not really needed, but I have no objections if you decide to include it.

feoh:
I agree!

To the original submitter: thank you very much for your contribution and I do appreciate the time you took preparing it, but I'm afraid it's not compatible with our goals for the project.

Issue #554, opened 2023-12-23
The Return of "cannot find package bash-debug-adapter"


johnpekcan:
## Describe the bug

I believe we have a regression of a bug fix 😅 
The fix here: https://github.com/nvim-lua/kickstart.nvim/pull/210/commits/de1db70ea904e08e031c4f52f99375422e2c284b
Was removed recently: https://github.com/nvim-lua/kickstart.nvim/commit/6ffc5a169f090b0b4359deb6deff0b99ab123679

```
[ERROR Sat Dec 23 09:37:46 2023] ...l/share/nvim/lazy/mason.nvim/lua/mason-registry/init.lua:79: Cannot find package "\"bash-debug-adapter\"".
```

## To Reproduce

1. git clone per read me.
2. uncomment `require 'kickstart.plugins.debug'`
3. Look at `:MasonLog`
4. fwiw repro also works in https://github.com/dam9000/kickstart-modular.nvim  (I can open a separate issue there) 

## Desktop

- OS: 23.2.0 Darwin Kernel Version 23.2.0:  root:xnu-10002.61.3~2/RELEASE_ARM64_T6020 arm64
- Terminal: zsh 5.9 (x86_64-apple-darwin23.0)

## Neovim Version

```
NVIM v0.9.4
Build type: Release
LuaJIT 2.1.1700008891

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/opt/homebrew/Cellar/neovim/0.9.4/share/nvim"
```

feoh:
Thanks for the repott! Care to file a PR to fix this?

johnpekcan:
Cheers @feoh , I've submitted a PR to restore that line.

feoh:
Thanks very much for the fix! Really appreciate the contribution!

PR #865, opened 2024-04-19
fix: restore Mason config timing for DAP startup (again)

GameFuzzy:
Addresses regression of issue #554 fixed in #555 but later reintroduced in 8b5d48a and adds note to prevent regression from occurring for a third time.


feoh:
Thanks for that!

PR #874, opened 2024-04-22
fix: highlight group clear on each attach

sudo-tee:
In my other pull request when opening a 2 buffers it would clear the whole group and removing the commands for other buffers.

It was an oversight on my part. and I introduced an unwanted side effect.


dam9000 (approved):
LGTM, thanks.

PR #876, opened 2024-04-22
Remove nvim-treesitter prefer_git

aripollak:
My `~/.config/nvim` is in a git submodule, and starting up neovim from within `~/.config/nvim` with the latest kickstart.nvim init.lua with prefer_git set to true seems to have broken my local git repo. I'm now getting "fatal: unable to read ..." errors, so I have to reinitialize the submodule from scratch to fix it.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -850,8 +850,6 @@ require('lazy').setup({
     config = function(_, opts)
       -- [[ Configure Treesitter ]] See `:help nvim-treesitter`
 
-      -- Prefer git instead of curl in order to improve connectivity in some environments
-      require('nvim-treesitter.install').prefer_git = true
       ---@diagnostic disable-next-line: missing-fields
       require('nvim-treesitter.configs').setup(opts)
 


feoh:
I too use Git to manage my home directory and dot files.

I don't think these problems are related, and I'm extremely reluctant to undo this change.

aripollak:
I haven't determined why, but it seems to break only while inside a neovim config repo that's a submodule. I just reproduced at least the nvim-treesitter git errors, if not total repo breakage, with the following while my `~/.config/nvim` is checked out to the latest kickstart.nvim master:

```
rm ~/.local/share/nvim/lazy/nvim-treesitter/parser/*
cd ~/.config/nvim
git rebase -i
```

image

aripollak:
Related issue: https://github.com/nvim-treesitter/nvim-treesitter/issues/5353

scmx:
My git worktree breaks every time I do git commit now 😨 . Here's a repro attempt, to show how it might cause problems for kickstart users:
```bash
git clone https://github.com/nvim-lua/kickstart.nvim.git "${XDG_CONFIG_HOME:-$HOME/.config}"/nvim-kickstart-repro
NVIM_APPNAME="nvim-kickstart-repro" nvim # let everything install
mkdir repro
cd repro
git init
git commit -m init --allow-empty
git worktree add ../repro-worktree
cd ../repro-worktree
NVIM_APPNAME="nvim-kickstart-repro" EDITOR=nvim git commit --allow-empty
# Start writing a message, move around a bit, now you'll see the message below:
nvim-treesitter[gitcommit]: Error while checking out revision
fatal: unable to read tree (edd817e0532f179b7f7f371dc180629070945f0c)
```
In my real repo I got `fatal: not a git repository (or any of the parent directories): .git` when doing git status.
Here I got a slightly different behaviour with an empty repo:
```bash
❯ git status -bs
## repro-worktree
AD .github/workflows/ci.yml
...
AD tree-sitter-gitcommit.wasm
```
The TSInstall gitcommit seems to have tried to install those files into the repo, git added them and then moved them? Doing git reset I'm back to a blank repo. In my project it broke `.git`.

The problem doesn't happen when I first comment out prefer_git

feoh:
hi, I'm sorry you're having trouble but kickstart needs to enable and encompass. Different potential configurations.

we made this change in order to make the startup process, simpler and easier for windows users and not be but I think you need to comment out one entry in your local configuration is a reasonable trade-off :-)

aripollak:
The worst-case scenario here is totally breaking the local git checkout, which IMHO seems like a less than ideal default. I appreciate wanting a working default for everyone. But it seems to me that it would be better to have Windows users uncomment a line to get a certain feature to work rather than potentially irreversibly losing data for other users. Also, based on https://github.com/nvim-treesitter/nvim-treesitter/wiki/Windows-support, it seems like this is not a requirement for all Windows users anyway.

PR #877, opened 2024-04-22
Remove treesitter auto_install

aripollak:
It seems like some of the treesitter language are not ready for general use. In my case, Running `git commit -v` resulted in the `gitcommit` language being installed, which resulted in no syntax highlighting at all in the diff below the commit message.


feoh:
Please feel free to make this change in your own personal fork. I don't think it makes sense in the scope of the broader project.

Issue #878, opened 2024-04-23
feat: Add spell checker ?


afifurrohman-id:
# Probably needed for starting?

```lua
vim.opt.spell = true
vim.opt.spelllang = 'en_us'
```

dam9000:
Well, enabling it by default and setting the language is probably a user customization.
However, I can propose a keymap to toggle the spell checking, here is a PR: https://github.com/nvim-lua/kickstart.nvim/pull/879

afifurrohman-id:
> Well, enabling it by default and setting the language is probably a user customization. However, I can propose a keymap to toggle the spell checking, here is a PR: #879

Hmm alright, but i think add example within commenting may is worth?

feoh:
Hi.

The universe of potential extensions to Neovim is *VAST*.

Kickstart doesn't try to be all things to all people.

Spell checking is a feature with MANY contexts. I personally enable it via filetype plugins for Markdown and Text, and that's all.

The feature is extremely well documented in the standard Neovim docs [here](https://neovim.io/doc/user/spell.html).

All I had to do was do a web search on "Neovim spelling" to find it.

Thanks for your contribution, but I don't think this is a good fit for kickstart at this time.

PR #879, opened 2024-04-23
Toggle spell checking

dam9000:
Suggested spell check toggle in response to issue: #878
Feel free to close if you think this does not belong in kickstart.


feoh:
I don't have a problem with this, but spell checking is so well documented and such a matter of individual taste I'm not sure about whether or not this is a good idea.

Thanks for creating the PR though in any case!

dam9000:
@feoh ok, I'll just close this one then.

Issue #884, opened 2024-04-27
Clipboard tool not listed as dependency


stgpepper:
## Describe the bug
Installation instructions do not include installation of clipboard tool. Still suing the system clipboard is on by default. This might cause confusion to new users when starting out. On my case I found this out when trying things out in fresh Ubuntu install. On my situation installing xclip solved the issue.

## To Reproduce
1. Use newly installed Ubuntu (mine running on wsl2).
2. Only install dependencies specified in documentation.
3. Install neovim and neovim kickstart as per documentation.
4. Run neovim and try to copy anything from a file to a system clipboard.

## Desktop
- OS: Ubuntu LTS 22.04
- Terminal: ?

## Neovim Version
```
NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1692716794
```

## Also...
There might be also some other application

dam9000:
@stgpepper opened a PR to address this, please review: https://github.com/nvim-lua/kickstart.nvim/pull/886

stgpepper:
> @stgpepper opened a PR to address this, please review: https://github.com/nvim-lua/kickstart.nvim/pull/886
> 

I would review, but sadly I lack foundational knowledge of the scope of this issue. The solution in PR looks like it would work in my case. And if it were included in the docs from the get go it probably would have saved me some debugging time. I simply dont know what other options are available and what limitations migth be present and therefore feel incompetent to say wheter or not it works for the community.

That said, looks good to me and appreciate the PR 👍. Probably better that someone more experienced on the subject would actually aprove it if nessessary.

PR #886, opened 2024-04-27
README: add clipboard tool dependency

dam9000:
Fixes: #884
Neovim requires an external tool for proper system clipboard integration.
Some systems install this already by default:
- on Fedora xsel is already installed by default
- on Windows using the choko install the win32yank is alredy installed

This is not installed by default on ubuntu or debian so adding that to the dependencies list and to the install instructions snippets.


feoh:
I use [Smartyank](https://github.com/ibhagwan/smartyank.nvim) so I can forget about all these shenanigans and have cut/paster Just Work :)

Happy to merge this in any case. Thanks for the contribution!

PR #896, opened 2024-05-02
docs(init.lua): improve readability

vladdoster:
- grammar
- punctuation
- consistent use of `Kickstart`

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -24,15 +24,15 @@ What is Kickstart?
 
   Kickstart.nvim is *not* a distribution.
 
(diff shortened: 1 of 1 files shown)


feoh:
Can you help me understand the motivation for these changes? Are you making it easier for non English speakers or something?

Otherwise I feel (and this is subjective) like this may be draining the color from the prose, which I don't think we want?

feoh:
Sorry, re-reading this I don't think these changes are a good fit for the project.

But thank you very much for your contribution in any case!

Issue #899, opened 2024-05-02
Kickstart calls `kickstart-lsp-highlight` augroup regardless of support


cpiber:
## Describe the bug

Kickstart registers a autocmd for `LspDetach`, which calls the augroup `kickstart-lsp-highlight`, but if the server does not support the capability `documentHighlightProvider`, this augroup does not exist. The following error is thrown:

```                                                                                         
Error detected while processing LspDetach Autocommands for "*":
Error executing lua callback: <cut>/.config/nvim/init.lua:618: Invalid 'group': 'kickstart-lsp-highlight'
stack traceback:
        [C]: in function 'nvim_clear_autocmds'
        <cut>/.config/nvim/init.lua:618: in function <<cut>/.config/nvim/init.lua:616>
        [C]: in function 'nvim_exec_autocmds'
        /usr/share/nvim/runtime/lua/vim/lsp.lua:388: in function </usr/share/nvim/runtime/lua/vim/lsp.lua:386>
```

## To Reproduce

1. Install a LSP server that does not support highlighting like grammarly (I believe)
2. Open a file for that server, e.g. markdown

## Desktop

- OS: Ubuntu 23.10 `Linux <cut> 6.5.0-28-generic #29-Ubuntu SMP PREEMPT_DYNAMIC Thu Mar 28 23:46:48 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux`
- Terminal: kitty

## Neovim Version

```
NVIM v0.10.0-dev
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

feoh:
Ooh. This is interesting. Thanks for the report!

Would you be willing to take a look around at other people's configurations to see how they might properly support this use case?

cpiber:
I'm not familiar, and not very involved in the community, so that would be difficult, but I'll have a look

But it should easy to check if the group exists and just not run it then, right? I'll look into this a bit later and open a PR if that is fine

dam9000:
This was introduced with: https://github.com/nvim-lua/kickstart.nvim/pull/864 by @sudo-tee 
Perhaps we can ask @sudo-tee to refine this further.

feoh:
That would be amazing!

For an honest to god functionality breaking issue like this I and I'm sure others would be happy to help whip it into shape if the first go around isn't perfect.

Thanks *so much* for being willing to contribute, and thanks again for the report!

feoh:
[...]

PR #900, opened 2024-05-02
Move LspDetach handler near kickstart-lsp-highlight group

dam9000:
Moved to make sure the kickstart-lsp-highlight group exists when
the LspDetach handler is invoked. The LspDetach handler is used
to clean up any lsp highlights that were enabled by CursorHold
if the LSP is stopped or crashed.

Fixes: #899

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -526,6 +526,14 @@ require('lazy').setup({
               group = highlight_augroup,
               callback = vim.lsp.buf.clear_references,
             })
+
+            vim.api.nvim_create_autocmd('LspDetach', {
+              group = vim.api.nvim_create_augroup('kickstart-lsp-detach', { clear = true }),
+              callback = function(event2)
+                vim.lsp.buf.clear_references()
+                vim.api.nvim_clear_autocmds { group = 'kickstart-lsp-highlight', buffer = event2.buf }
+              end,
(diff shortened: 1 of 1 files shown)


feoh:
@cpiber Could you please test these changes with your configuration to see if this fixes your problem?

It would be great to get independent verification that we're fixing this problem.

@dam9000 Thank you for being johnny on the spot with an excellent PR once again!

dam9000:
Johnny on the spot :smile:

feoh:
No, my johnny is a non conformist. He spells his first name with a lowercase 'j' :P 

Thanks for the correction.

cpiber (approved):
This also works, not quite the move I had in mind :) LGTM

Only thing I'm not sure about, is it a problem if `vim.lsp.buf.clear_references()` is not executed? Since @sudo-tee added it, could you address this please?

feoh:
I think we've got this one nailed. @sudo-tee should you have time to check back and have any other thoughts to contribute, please do.

Merging this change isn't a one way door, we can always improve things as we go :)

sudo-tee:
Very late to the party. But I confirm that this works for the use case I had.  Thanks for pinging me in the discussion

PR #913, opened 2024-05-06
Add command to toggle autoformatting

ethuleau:
This is a convenient command if you want to limit your code contribution to existing code base focused.

(that's my fist nvim setting contib)


feoh:
Thanks for your contribution but I don't think it's a good fit for this project at this time.

PR #916, opened 2024-05-06
Update init.lua to prevent mason package error

sahanagana:
I kept repeatedly getting mason errors where the results of functions were 'nil' and checkhealth was failing. Rearranging the code this way fixed that problem.

**NOTE**

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -91,7 +91,7 @@ vim.g.mapleader = ' '
 vim.g.maplocalleader = ' '
 
 -- Set to true if you have a Nerd Font installed and selected in the terminal
-vim.g.have_nerd_font = false
+vim.g.have_nerd_font = true
 
 -- [[ Setting options ]]
 -- See `:help vim.opt`
@@ -417,6 +417,7 @@ require('lazy').setup({
       'WhoIsSethDaniel/mason-tool-installer.nvim',
 
       -- Useful status updates for LSP.
+      -- NOTE: wow i can type a note too look at mee ooooo
       -- NOTE: `opts = {}` is the same as calling `require('fidget').setup({})`
       { 'j-hui/fidget.nvim', opts = {} },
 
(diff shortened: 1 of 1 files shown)


init.lua
```diff
@@ -91,7 +91,7 @@ vim.g.mapleader = ' '
 vim.g.maplocalleader = ' '
 
 -- Set to true if you have a Nerd Font installed and selected in the terminal
-vim.g.have_nerd_font = false
+vim.g.have_nerd_font = true
```
stgpepper, line 94: I actually would prefer this being false. As i see it Kickstart project is about learning to configure stuff. If this is enabled by default it would:
1. Make people not go through the installation process of nerdfont
2. Make it harder to detect what is native part of neovim and what is not

init.lua
```diff
@@ -417,6 +417,7 @@ require('lazy').setup({
       'WhoIsSethDaniel/mason-tool-installer.nvim',
 
       -- Useful status updates for LSP.
+      -- NOTE: wow i can type a note too look at mee ooooo
```
stgpepper, line 420: Does this note serve a purpose?

sahanagana: I apologize for the trouble! I must have accidentally copied over my modified init.lua (which included the nerd font change and me testing out the note feature) instead of the one with only the code rearrangement. I'll make another pull request with the latter file.

sahanagana:
> I can't see how this rearrange could fix anything.

I was getting issues with mason on startup, and when I ran checkhealth, it completely failed. I have screenshots of this behavior, as well as these issues being resolved after I made the change.
I found the solution online with moving the setup() call, which fixed the error. I'm not entirely sure how it works, but I believe it has to do with calling mason-lspconfig before setting up mason. It may also have to do with my neovim version (v0.9.4).

dam9000:
You're moving the call to `mason setup()` from after the `servers=` assignment to before of it. The assignment does not execute any code, I can't see how this change could affect anything. I understand this now works for you but I believe the fix is not related to your change, but perhaps just the fact that you re-run the nvim and Mason was then able to complete with whatever failed before. Are you able to reproduce the issue by reverting the change?

feoh:
I agree.

I'm glad this helped you, but if you want us to merge this you'll need to give us some sense of why the change is necessary.

Either please close this PR or provide the screenshots and error messages you're getting that prompted this change.

feoh:
No trouble, but I'm not sure that other PR is necessary.

PR #925, opened 2024-05-10
Automatically set detached state as needed.

PerMalmberg:
Hi,

Here's a fix to automatically set the detached state for delve based on the OS. Gives a better first-time experience when delve doesn't crash :)

### DIFF
--- a/lua/kickstart/plugins/debug.lua
+++ b/lua/kickstart/plugins/debug.lua
@@ -85,6 +85,11 @@ return {
     dap.listeners.before.event_exited['dapui_config'] = dapui.close
 
     -- Install golang specific config
-    require('dap-go').setup()
+    require('dap-go').setup {
+      delve = {
+        -- On Windows delve must not be run attached or it crashes.
+        detached = vim.loop.os_uname().sysname ~= 'Windows_NT',
+      },
+    }
   end,
 }
--- a/lua/kickstart/plugins/debug.lua
+++ b/lua/kickstart/plugins/debug.lua
@@ -89,7 +89,7 @@ return {
       delve = {
         -- On Windows delve must be run attached or it crashes.
         -- See https://github.com/leoluz/nvim-dap-go/blob/main/README.md#configuring
-        detached = not vim.fn.has 'win32',
+        detached = vim.fn.has 'win32' == 0,
       },
     }
   end,
--- a/lua/kickstart/plugins/debug.lua
+++ b/lua/kickstart/plugins/debug.lua
@@ -87,8 +87,9 @@ return {
     -- Install golang specific config
     require('dap-go').setup {
       delve = {
-        -- On Windows delve must not be run attached or it crashes.
-        detached = vim.loop.os_uname().sysname ~= 'Windows_NT',
+        -- On Windows delve must be run attached or it crashes.
+        -- See https://github.com/leoluz/nvim-dap-go/blob/main/README.md#configuring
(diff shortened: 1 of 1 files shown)


lua/kickstart/plugins/debug.lua
```diff
@@ -85,6 +85,11 @@ return {
     dap.listeners.before.event_exited['dapui_config'] = dapui.close
 
     -- Install golang specific config
-    require('dap-go').setup()
+    require('dap-go').setup {
+      delve = {
+        -- On Windows delve must not be run attached or it crashes.
+        detached = vim.loop.os_uname().sysname ~= 'Windows_NT',
```
dam9000, line 91: maybe this should be used:
```
vim.fn.has 'win32' == 1
```
which is for example already used in init.lua line 676

PerMalmberg: If that is preferred, please amend the PR (maintainers should have access). I've never worked with the nvim API before so I barely know it.

dam9000: I'm not a maintainer, just occasional contributor :)
It was just a suggestion, this is the first time I see `vim.loop.os_uname`, but I saw `vim.fn.has` many times before, but I see both are used in various plugins, for example if we check in: `~/.local/share/nvim/` :
```
grep -r 'fn.has.*win' . | wc -l
40

grep -r 'vim.loop.os_uname.*Win' . | wc -l
3
```
we can see that both are used but fn.has is more common.
To conclude, it's fine as is. You can disregard my comment.

PerMalmberg: It is valuable input to maintainers 👍 Thank you for taking the time to look at it.

feoh: Thanks for your contribution!

Please just try the API @dam9000 suggested locally and amend your PR if it works.

It doesn't make any sense for one of us to amend your PR. That's the purpose of review feedback, to help *you* make your contribution better :)

Thanks for understanding.

PerMalmberg: Sure thing. Right after dinner. 🍢

PerMalmberg:
Updated as requested, works on my machine :)

lua/kickstart/plugins/debug.lua
```diff
@@ -85,6 +85,12 @@ return {
     dap.listeners.before.event_exited['dapui_config'] = dapui.close
 
     -- Install golang specific config
-    require('dap-go').setup()
+    require('dap-go').setup {
+      delve = {
+        -- On Windows delve must be run attached or it crashes.
+        -- See https://github.com/leoluz/nvim-dap-go/blob/main/README.md#configuring
+        detached = not vim.fn.has 'win32',
```
dam9000, line 92: Hmm I think this will always be false
vim.fn.has will return 0 or 1, and in lua `not 0` and `not 1` will both be false
so I think this needs to be:
```
detached = vim.fn.has 'win32' ~= 0,
```
can you please test that?

PerMalmberg: Oh right, Lua does that funny thing with numbers and boolean ops. My bad, will fix.

PerMalmberg: Fixed

dam9000: Thanks!

dam9000 (approved):
LGTM

PR #934, opened 2024-05-16
added support if nvim extention is being used in vscode

sumanth-lingappa:
[VSCODE neovim extention](https://marketplace.visualstudio.com/items?itemName=asvetliakov.vscode-neovim) pre-requisite is to have `nvim` installed and it opens `nvim's` config by default in vscode.
I use both nvim and vscode depending on my project need.

Some of the nvim's UI does not work well with VSCODE. 

This if-else clause will differentiate if we are using vscode or nvim directly.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -1,3 +1,8 @@
+-- if this file is opened in a vscode, exit from here
+if vim.g.vscode then
+  return
+end
+
 --[[
 
 =====================================================================


dam9000:
I don't think vscode users would be the target audience for kickstart so I'm not sure if this is really needed? In my view it just pollutes the code. Besides the suggested change is documented in the vscode neovim plugin install instruction (the link you provided) which one will have to go through and will be able to make the change for himself.

sumanth-lingappa:
I agree with your point @dam9000 
For me this is a must condition clause. I thought this will be useful for others out-of-the-box.

jiaad:
> I agree with your point @dam9000 For me this is a must condition clause. I thought this will be useful for others out-of-the-box.

I also think, it should be available out-of-the-box

feoh:
Hi. Thank you for your contribution but making a one line change to your config as specified in the vscode neovim extension feels like the right place for this to me. As such I don't think it's compatible with our project's goals and I'm closing for now.

PR #936, opened 2024-05-16
Neovim 0.10 updates

VlaDexa:
This is a PR that cleans up the config in accordance to new stable features.

I have made the following changes:
**Key Mappings:**
- Remove `K`, `[d` and `]d` mappings, since they are included by default
- Remove `<leader>e` mapping, because neovim started using `<C-W>d` for this by default

**Plugin changes:**
- Remove `Comment.nvim`, since neovim now includes their commenting tool (see :help commenting)

**Chores:**
- There is a new recommended way of checking for LSP capabilities, so change to that
- Remove check for `vim.lsp.inlay_hint` presence, since stable now supports it

**Fixes:**
- Fix problems with the default parameter of the buffer of `vim.lsp.inlay_hint.is_enabled`

**Accumulated PRs**
- #961
- #956

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -204,7 +204,7 @@ vim.api.nvim_create_autocmd('TextYankPost', {
 -- [[ Install `lazy.nvim` plugin manager ]]
 --    See `:help lazy.nvim.txt` or https://github.com/folke/lazy.nvim for more info
 local lazypath = vim.fn.stdpath 'data' .. '/lazy/lazy.nvim'
-if not vim.loop.fs_stat(lazypath) then
+if not vim.uv.fs_stat(lazypath) then
   local lazyrepo = 'https://github.com/folke/lazy.nvim.git'
   local out = vim.fn.system { 'git', 'clone', '--filter=blob:none', '--branch=stable', lazyrepo, lazypath }
   if vim.v.shell_error ~= 0 then
--- a/init.lua
+++ b/init.lua
@@ -162,9 +162,6 @@ vim.opt.hlsearch = true
 vim.keymap.set('n', '<Esc>', '<cmd>nohlsearch<CR>')
 
 -- Diagnostic keymaps
-vim.keymap.set('n', '[d', vim.diagnostic.goto_prev, { desc = 'Go to previous [D]iagnostic message' })
-vim.keymap.set('n', ']d', vim.diagnostic.goto_next, { desc = 'Go to next [D]iagnostic message' })
-vim.keymap.set('n', '<leader>e', vim.diagnostic.open_float, { desc = 'Show diagnostic [E]rror messages' })
 vim.keymap.set('n', '<leader>q', vim.diagnostic.setloclist, { desc = 'Open diagnostic [Q]uickfix list' })
 
 -- Exit terminal mode in the builtin terminal with a shortcut that is a bit easier
@@ -237,11 +234,6 @@ require('lazy').setup({
   --
   -- Use `opts = {}` to force a plugin to be loaded.
   --
-  --  This is equivalent to:
-  --    require('Comment').setup({})
-
-  -- "gc" to comment visual regions/lines
-  { 'numToStr/Comment.nvim', opts = {} },
 
   -- Here is a more advanced example where we pass configuration
   -- options to `gitsigns.nvim`. This is equivalent to the following Lua:
@@ -498,10 +490,6 @@ require('lazy').setup({
           -- or a suggestion from your LSP for this to activate.
           map('<leader>ca', vim.lsp.buf.code_action, '[C]ode [A]ction')
 
-          -- Opens a popup that displays documentation about the word under your cursor
-          --  See `:help K` for why this keymap.
-          map('K', vim.lsp.buf.hover, 'Hover Documentation')
-
           -- WARN: This is not Goto Definition, this is Goto Declaration.
           --  For example, in C this would take you to the header.
           map('gD', vim.lsp.buf.declaration, '[G]oto [D]eclaration')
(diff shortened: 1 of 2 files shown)


VlaDexa:
I should add that this probabbly shouldn't be merged too soon, since 0.10 released today, and it's probably not in many repos. But, I think these changes are fit to be in the main, since the point of kickstart is to be small and support the latest stable

dam9000:
I think merging this should be delayed by a couple of months, since it will take a while for all the distributions and package managers to upgrade to the latest 0.10 release. In the meantime, the kickstart config as is works perfectly fine also on 0.10. Merging this does not fix anything but it will break functionality for people still on 0.9.5.

dam9000:
In addition, if these changes are integrated then the `:chekhealth` would also need to be updated to require minimum version to be 0.10.0 instead of the current 0.9.4 - see `lua/kickstart/health.lua`

Alternatively these changes could be made conditional with something like:
```
if vim.fn.has 'nvim-0.10' == 1 then
...
end
```
curiously, the below version check does not work for me in nvim 0.10:
```
if vim.version.ge(vim.version(), '0.10') then
...
end
```
any idea why? I tested it with the prebuild nvim-linux64.tar.gz 0.10 release.

VlaDexa:
> prebuild nvim-linux64.tar.gz 0.10 release

Check your `vim.version()`. Just downloaded the release files from github, and it seems that they have packaged a version with a dev mark on it. Their version comparator looks for that and if it finds some prerelease mark it shows that 0.10-dev is greater than 0.9 but still less than 0.10

dam9000:
@VlaDexa you are right, vim.version() says:
`0.10.0-dev+g27fb62988`

init.lua
```diff
@@ -540,7 +528,7 @@ require('lazy').setup({
           -- code, if the language server you are using supports them
           --
           -- This may be unwanted, since they displace some of your code
-          if client and client.server_capabilities.inlayHintProvider and vim.lsp.inlay_hint then
+          if client and client.supports_method(vim.lsp.protocol.Methods.textDocument_inlayHint) then
             map('<leader>th', function()
               vim.lsp.inlay_hint.enable(not vim.lsp.inlay_hint.is_enabled())
```
arthurbailao, line 533: LSP is warning the following: `This function requires 1 argument(s) but instead it is receiving 0.`

```suggestion
              vim.lsp.inlay_hint.enable(not vim.lsp.inlay_hint.is_enabled {})
```

agingorange: Thanks, that works.

mmirus: YMMV, but I had issues until I passed the buffer in the filter table:

```suggestion
              vim.lsp.inlay_hint.enable(not vim.lsp.inlay_hint.is_enabled({ bufnr = event.buf }))
```

blarz:
Regarding the delay merging the PR, the README says:

> Kickstart.nvim targets only the latest ['stable'](https://github.com/neovim/neovim/releases/tag/stable) and latest ['nightly'](https://github.com/neovim/neovim/releases/tag/nightly) of Neovim.

So it would make sense to merge this now or soonish IMHO.

At which point do you declare it ready to go, when you say we should wait for distributions to package it?

But I also see this point:

> Merging this does not fix anything but it will break functionality for people still on 0.9.5.

How was this handled during the 0.9 release?

VlaDexa:
Can I collect here commits from other PR's that also suggest stuff for neovim 0.10? I think GitHub will autoclose those ones if this big PR containing their commits get merged
I'm talking about #961 and #956, for example

VlaDexa:
Got tired of waiting for someone to reply. If anyone wants for these commits from other PR's to not be included here please tell me.
For PR authors whose commits I merge here: please don't close your requests, as inclusion here doesn't mean it will get merged upstream

rmacklin:
> I think GitHub will autoclose those ones if this big PR containing their commits get merged

That would only happen if 1) you did a `merge` of https://github.com/nvim-lua/kickstart.nvim/pull/961 rather than a `cherry-pick` (as it stands now, only https://github.com/nvim-lua/kickstart.nvim/pull/956 could be auto-closed, since that PR's exact history is included in this branch's history) and 2) this PR was merged using a merge commit rather than a squash-merge - but the maintainers normally do a squash-merge so... 🤷‍♂️ 

Edit: Side note - you inspired me to add some more reasons to https://github.com/rmacklin/why_i_dont_recommend_squash_and_merge

rivenirvana:
Are these going to be merged anytime soon?

VlaDexa:
Fixes #992

feoh:
Hi all.

I had to take a breather because I was getting kinda cranky after a wave of "YOUR CODE IS CRAP WHY CAN'T YOU NOT SUCK?" issues :)

I'm not averse to merging this now. Does it contain the :checkhealth changes @dam9000 recommended?

VlaDexa:
> Does it contain the :checkhealth changes @dam9000 recommended?

Neovim still hasn't fixed the versioning issue in the official releases, so the 0.10 version check still won't work. We can make it compare against '0.10-dev' version, which works for currents stable but will also pass anyone on old nightlies.

feoh:
is there any reason why allowing it to pass for nightly would be bad?

@dam9000 do you stink this pull request is premature still?

VlaDexa:
> is there any reason why allowing it to pass for nightly would be bad?

Couldn't think for any but wrote that for your information. Seeing that you're also fine with this, made a new commit.

feoh:
As I'm fond of saying about Git  based projects, this isn't a one way door :) I'm going to merge this.

rmacklin:
Glad to see this merged! I do think we should fix this error (previously discussed in https://github.com/nvim-lua/kickstart.nvim/pull/961#discussion_r1646808700):

image

so I've opened https://github.com/nvim-lua/kickstart.nvim/pull/1040 to address it

dam9000:
@feoh for the record since you asked: I'm fine with this merge, enough time has passed since the 0.10 release.

feoh:
thank you for chiming in! As always, I very much appreciate your contributions. You're huge help. I wish I had the power to grant you merge permissions.

PR #952, opened 2024-05-26
Add plugin: oil file system editor and explorer

rajive:
**NOTE**

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -878,6 +878,7 @@ require('lazy').setup({
   -- require 'kickstart.plugins.lint',
   -- require 'kickstart.plugins.autopairs',
   -- require 'kickstart.plugins.neo-tree',
+  -- require 'kickstart.plugins.oil',
   -- require 'kickstart.plugins.gitsigns', -- adds gitsigns recommend keymaps
 
   -- NOTE: The import below can automatically add your own plugins, configuration, etc from `lua/custom/plugins/*.lua`
--- a/lua/kickstart/plugins/oil.lua
+++ b/lua/kickstart/plugins/oil.lua
@@ -1,0 +1,30 @@
+return {
+  { -- Oil filesystem editor and explorer
+    'stevearc/oil.nvim',
+    dependencies = {
+      -- Useful for getting pretty icons, but requires a Nerd Font.
+      { 'nvim-tree/nvim-web-devicons', enabled = vim.g.have_nerd_font },
+    },
+    config = function()
+      require('oil').setup {
+        columns = { 'icon' },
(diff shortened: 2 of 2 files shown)


init.lua
```diff
@@ -878,6 +878,7 @@ require('lazy').setup({
   -- require 'kickstart.plugins.lint',
   -- require 'kickstart.plugins.autopairs',
   -- require 'kickstart.plugins.neo-tree',
+  require 'kickstart.plugins.oil',
```
VlaDexa:
```suggestion
  -- require 'kickstart.plugins.oil',
```
I think this should follow the other plugins in being disabled by default

lua/kickstart/plugins/oil.lua
```diff
@@ -0,0 +1,27 @@
+return {
+  { -- Oil filesystem editor and explorer
+    'stevearc/oil.nvim',
+    dependencies = { 'nvim-tree/nvim-web-devicons' },
```
VlaDexa: Can you maybe make it so it respects the web-devicons config from init.lua, since it's optional?

init.lua
rajive: Great suggestion!

lua/kickstart/plugins/oil.lua
rajive: Great suggestion! Yes, updating accordingly.

rajive:
Hi @VlaDexa Updated to incorporate your feedback.

VlaDexa:
Hi! I've seen those changes, but I don't have any power to merge them since I'm not a maintainer.

By the way, since suggesting them I've learned that it's not necessary to specify when to enable web-devicons, because lazy already know when to do that from a specification from init

feoh:
Hiya!

I love and use oil, but I guess one question I have is: As awesome as our favorite file tree manipulator is, is it actually appropriate for Neovim newcomers?

The answer may well be "heck yeah!" in which case we can certainly merge this.

I'd love to hear from the community on this. Again I apologize for my long absence. IMO if you can't contribute in an even handed, well tempered way, it's time to step back and take a breath :)

dlugoschvincent:
Also, you already have mini.nvim installed. So just enabling mini.files gives basically the same functionality without needing a new plugin.
https://github.com/echasnovski/mini.nvim/blob/main/readmes/mini-files.md

feoh:
I don't know enough about mini, but I think the point stands that in general there are many many choices for this kind of functionality. We currently offer one. Let's leave it at that for now and those of us who like something different are welcome to customize accordingly.

PR #953, opened 2024-05-26
Enable `smartindent` option for default configuration

iharkatkavets:
This PR enables a pretty useful option that probably exists in most IDE's by default - indentations. 
If you see if that can be useful to be included in `basic` configuration, please merge my PR.
If not, I think it has to be at least mentioned(and commented) in the `basic` configuration.


iton0:
What's the purpose of vim-sleuth if this in an option in vim

feoh:
Yup thanks for your contribution but like @iton0 said we have vim-sleuth for this.

PR #956, opened 2024-05-27
refactor: replace vim.loop with vim.uv

mrr11k:
This PR uses vim.uv instead of the deprecated vim.loop. 

See https://github.com/neovim/neovim/pull/22846 for more info.


purarue:
The [lazy instructions](https://github.com/folke/lazy.nvim?tab=readme-ov-file#-installation) use a `or`, incase the user is on older systems to allow using `vim.loop` still:

`if not (vim.uv or vim.loop).fs_stat(lazypath) then`

I know kickstart targets stable, but should probably match that instead of requiring the user to be on v0.10+, would lead to one less error if a user happened to be on an old version

feoh:
I think this is obviated by #936 please feel free to re-open if I'm wrong.

Issue #960, opened 2024-06-02
neodev has been deprecated in favour of lazydev


jpoliv:
The neodev plugin is now end-of-life and should be replaced by the new lazydev plugin.

neodev plugin:
https://github.com/folke/neodev.nvim

lazydev plugin:
https://github.com/folke/lazydev.nvim

Caveat:
The lazydev plugin requires Neovim >= 0.10.0

neodev usage:
https://github.com/nvim-lua/kickstart.nvim/blob/master/init.lua#L423-L425

maztaim:
It's deprecated enough that updating is blocked using Lazy.

feoh:
Fixed in a recent merge.

PR #961, opened 2024-06-02
Upgrade folke/neodev (sunsetting) to folke/lazydev

qgates:
Address issue #960

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -420,9 +420,9 @@ require('lazy').setup({
       -- NOTE: `opts = {}` is the same as calling `require('fidget').setup({})`
       { 'j-hui/fidget.nvim', opts = {} },
 
-      -- `neodev` configures Lua LSP for your Neovim config, runtime and plugins
+      -- `lazydev` configures Lua LSP for your Neovim config, runtime and plugins
       -- used for completion, annotations and signatures of Neovim apis
-      { 'folke/neodev.nvim', opts = {} },
+      { 'folke/lazydev.nvim', ft = 'lua', opts = {} },
     },
     config = function()
       -- Brief aside: **What is LSP?**


iton0:
Since this is a dependency of the lsp would this not be loaded along with the lsp config regardless of filetype?

init.lua
```diff
@@ -420,9 +420,9 @@ require('lazy').setup({
       -- NOTE: `opts = {}` is the same as calling `require('fidget').setup({})`
       { 'j-hui/fidget.nvim', opts = {} },
 
-      -- `neodev` configures Lua LSP for your Neovim config, runtime and plugins
+      -- `lazydev` configures Lua LSP for your Neovim config, runtime and plugins
       -- used for completion, annotations and signatures of Neovim apis
-      { 'folke/neodev.nvim', opts = {} },
+      { 'folke/lazydev.nvim', ft = 'lua', opts = {} },
```
rmacklin, line 425: To avoid an "undefined field `fs_stat`" LSP error on https://github.com/nvim-lua/kickstart.nvim/blob/20fa60a709e30fcf9ab6060e20750841141f7937/init.lua#L207

perhaps this should use the configuration suggested in https://github.com/folke/lazydev.nvim/tree/6184ebbbc8045d70077659b7d30c705a588dc62f#-installation
```suggestion
      {
        'folke/lazydev.nvim',
        ft = 'lua',
        opts = {
          library = {
            -- Load luvit types when the `vim.uv` word is found
            { path = 'luvit-meta/library', words = { 'vim%.uv' } },
          },
        },
      },
      { 'Bilal2453/luvit-meta', lazy = true },
```

qgates: I got the sense that these are optional configs for those who want specific vim.uv LSP typings. Not everyone will need/want them. However, if there's no downside and or most people need them, I see no problem with it.

At the risk of bloat perhaps someone else can chime in?

gustavosinacio: I feel like, for a kickstart config, the suggested would be the preferred aproach. Anyone who doens't want this will propably be more aware of how to remove it.

rmacklin: As it stands, on `master` using `folke/neodev.nvim` we don't see an "undefined field `fs_stat`" LSP error. But with the change in this PR, we do see that error (unless we apply the suggestion above). That's why I think we might want to follow the configuration suggested in https://github.com/folke/lazydev.nvim/tree/6184ebbbc8045d70077659b7d30c705a588dc62f#-installation - Otherwise, merging this PR as-is will leave the fresh `kickstart.nvim` installation with an LSP error in `init.lua`, which IMO degrades the desired "first timer" experience.

feoh:
Has this PR been abandoned? It seems like merging it will leave us with errors which is no bueno. Could we either fix or close? Thanks a bunch!

rmacklin:
@feoh This PR was cherry-picked into #936, which was just merged, so this PR can be closed.

But this also means the LSP error discussed above is present in `master` now. I've opened a PR with my suggestion from https://github.com/nvim-lua/kickstart.nvim/pull/961#discussion_r1646808700 here:
- https://github.com/nvim-lua/kickstart.nvim/pull/1040

feoh:
Thank you *so* much for being so incredibly thorough. I really appreciate having the links ready at hand so I can read up on the context and understand the nature of the fix and why it's important.

Closing this, merging that :)

PR #973, opened 2024-06-08
Update lazy.nvim bootstrap

iton0:
Used lazy.nvim bootstrap from [lazy.nvim site](https://github.com/folke/lazy.nvim?tab=readme-ov-file#-installation)


11xdeveloper:
This isn't needed due to #936

PR #982, opened 2024-06-12
fix mini.ai example command

syhner:


VlaDexa:
Actually, selection of quote by q also works fine

syhner:
> Actually, selection of quote by q also works fine

init.lua
```diff
@@ -801,7 +801,7 @@ require('lazy').setup({
       --
       -- Examples:
       --  - va)  - [V]isually select [A]round [)]paren
-      --  - yinq - [Y]ank [I]nside [N]ext [']quote
+      --  - yin' - [Y]ank [I]nside [N]ext [']quote
```
rmacklin, line 804: > Actually, selection of quote by q also works fine

@VlaDexa @syhner Perhaps then, the comment should instead be changed to:
```lua
      --  - yinq - [Y]ank [I]nside [N]ext [Q]uote
```
to match the use of `q` instead of `'` - thoughts?

rmacklin: Opened https://github.com/nvim-lua/kickstart.nvim/pull/985

PR #999, opened 2024-07-02
feat: add query parser to treesitter config

iton0:
Based on the [nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter?tab=readme-ov-file#modules) README.md which states query as one of the five parsers that should always be installed.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -835,7 +835,7 @@ require('lazy').setup({
     'nvim-treesitter/nvim-treesitter',
     build = ':TSUpdate',
     opts = {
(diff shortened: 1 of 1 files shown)


VlaDexa:
I think it means that nvim treesitter will always keep parsers from that list installed. I have opened [a discussion](https://github.com/nvim-treesitter/nvim-treesitter/discussions/6865) in nvim-treesitter repo for someone to clarify the meaning.

VlaDexa (approved):

tjdevries:
I just added the ones from nvim-treesitter, thanks for the PR though, would not have noticed otherwise

PR #1022, opened 2024-07-13
which-key v3 update

VlaDexa:
which-key recently had a v3 update and depreciated the register function, replacing it with an add function.

Also it's a bit more clear now, since there is now no empty map

PR #1023, opened 2024-07-13
Fix the which-key spec to follow the new version.

ghost:
With the kickstart.nvim config, I noticed the annoying which-key health warning popups. The warning suggested to change the specs to follow the new version standard. I changed it and that resolved the problem.


ghost:
Just noticed that that's a duplicate of PR #1022.

PR #1042, opened 2024-07-22
Add explicit dependency of nvim-lspconfig on cmp-nvim-lsp

jstrot:
The plugin spec for `neovim/nvim-lspconfig` does a `require('cmp_nvim_lsp')` in the `LspAttach` event handler but does not explicitly add a dependency on `hrsh7th/cmp-nvim-lsp`.

This is fine in the default `kickstart.nvim` layout but may not work if someone decides to break the specs into separate folders or decides to remove/disable `nvim-cmp`.

This PR simply adds the explicit dependency where used.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -424,6 +424,9 @@ require('lazy').setup({
         },
       },
       { 'Bilal2453/luvit-meta', lazy = true },
+
+      -- Allows extra capabilities provided by nvim-cmp
+      'hrsh7th/cmp-nvim-lsp',
     },
     config = function()
       -- Brief aside: **What is LSP?**


shaoyanji (approved):
I don't know how this was deleted in the first place

iton0:
Should the cmp-nvim-lsp dependency be removed from the nvim-cmp setup?

feoh:
I'm seeing lots of questions here and not many answers.

@jstrot thanks for your contribution, could you please respond to some of these comments? Happy to merge this if we can get consensus around what will work.

jstrot:
> I don't know how this was deleted in the first place

I think it's been missing since day-1 (4c4096cf30acace7bf3cf5a0a4b6c5346a0d5a59).
When plugins are initialized in a specific order (like a monolithic init.lua does), the dependency is already loaded. The issue  arises if someone splits apart the config into separate files, e.g, lua/plugin/*.lua.

jstrot:
> Should the cmp-nvim-lsp dependency be removed from the nvim-cmp setup?

The issue I'm addressing is neovim/nvim-lspconfig missing a dependency on hrsh7th/cmp-nvim-lsp since neovim/nvim-lspconfig's config function there's a call to `require('cmp_nvim_lsp')`

As for nvim-cmp itself, once of the sources is "nvim_lsp" so its dependency on cmp-nvim-lsp should be kept.

jstrot:
> I'm seeing lots of questions here and not many answers.
> 
> @jstrot thanks for your contribution, could you please respond to some of these comments? Happy to merge this if we can get consensus around what will work.

Sorry for taking time to respond, I hardly get time during the week...
I'm happy to answer further questions and comments.

Thanks.

iton0:
> > Should the cmp-nvim-lsp dependency be removed from the nvim-cmp setup?
> 
> The issue I'm addressing is neovim/nvim-lspconfig missing a dependency on hrsh7th/cmp-nvim-lsp since neovim/nvim-lspconfig's config function there's a call to `require('cmp_nvim_lsp')`
> 
> As for nvim-cmp itself, once of the sources is "nvim_lsp" so its dependency on cmp-nvim-lsp should be kept.

I thought that if the LSP config loads before the nvim-cmp plugin, the cmp-nvim-lsp should be available in the nvim-cmp setup. If so, is an explicit dependency needed in the nvim-cmp setup?

PR #1049, opened 2024-07-24
performance: defer clipboard because xsel and pbcopy can be slow

abeldekat:
See [this](https://github.com/LazyVim/LazyVim/discussions/4112) discussion in LazyVim

The approach discussed has been merged in [LazyVim](https://github.com/LazyVim/LazyVim/pull/4120) and [AstroNvim](https://github.com/AstroNvim/astrocore/pull/28)

TLDR: 
Startup time performance is affected quite significantly when the clipboard provider is `xsel`(linux) or `pbcopy`(macos). I expect an improvement in these cases, especially on older pc's.

This PR makes sure that `vim.opt.clipboard` is only applied after `UiEnter`.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -111,9 +111,12 @@ vim.opt.mouse = 'a'
 vim.opt.showmode = false
 
 -- Sync clipboard between OS and Neovim.
+-- Schedule the setting after `UiEnter` because it can increase startup-time.
 --  Remove this option if you want your OS clipboard to remain independent.
 --  See `:help 'clipboard'`
-vim.opt.clipboard = 'unnamedplus'
+vim.schedule(function()
+  vim.opt.clipboard = 'unnamedplus'
+end)
 
 -- Enable break indent
 vim.opt.breakindent = true
--- a/init.lua
+++ b/init.lua
@@ -111,7 +111,7 @@ vim.opt.mouse = 'a'
 vim.opt.showmode = false
 
 -- Sync clipboard between OS and Neovim.
--- Schedule the setting after `UiEnter` because it can increase startup-time.
+--  Schedule the setting after `UiEnter` because it can increase startup-time.
 --  Remove this option if you want your OS clipboard to remain independent.
 --  See `:help 'clipboard'`
 vim.schedule(function()


dam9000:
In my view this just complicates the config, makes it confusing for the newcomer who would be the target audience, without providing a tangible improvement. What tasks can one do before UiEnter? I expect if you measure the time it takes from start to actually editing a file it would be the same? In what use case is this useful? Perhaps for running vim or lua scripts using neovim? Just trying to understand...

abeldekat:
Thanks for the reply!
I understand your concern about complicating the config. That said, I think the extra line is not overly complicated. The accompanying comment is important.

A similar question was asked on [reddit](https://www.reddit.com/r/neovim/comments/1e4je65/save_40_startuptime_for_free/) :

> Delaying a function call like this just means your startup time shows a smaller number, but the function is still called later in the event loop. You aren't saving any actual time by doing this. 

My response, based on my own [config](https://github.com/abeldekat/nvim_pde):

> Very true. However, in my opinion, the purpose of lazy loading is to render text on the screen as quickly as possible. The actual time is not that important. When I open Neovim with a file argument, that file is rendered immediately, treesitter included. Most of the plugins are loaded after. 


I think I do provide a tangible improvement. However that improvement varies between environments. Worst case: No improvement at all. Best case someone reported: A startup-time improvement of 300ms( see [this](https://github.com/LazyVim/LazyVim/discussions/4112) discussion). That's huge for a setting that's not even needed before `UiEnter`.

Also, I would argue that the intent of `kickstart` is not primarily to avoid complications for the novice neovim user. The config educates and provides ample comments. For example, `nvim-cmp` is loaded on `InsertEnter`. 
On my system, just `kickstart` takes 135 ms to load. Without `InsertEnter` on `nvim-cmp`, it takes 161 seconds, a difference of 26 ms. To compare, when scheduling the clipboard, startup time is 94 seconds: A difference of 41 ms.

igniscyan:
I'd like to chime in here and say that I do agree with @abeldekat's sentiment. Kickstart is a base config that takes a lot of the initial hurdle of configuring nvim out of the equation for a novice, and furthermore, it's a good base configuration **period.** I don't think the philosophy should be to shy away from performance improvements that may take a bit of further reading to completely understand.

With the above being said, what is overly confusing about wrapping an already existing line in a schedule function? It seems like it'd be a good opportunity to introduce vim.schedule(), and with the potential upside of lowering the base configuration's startup time in some environments this seems like a good change to me. Just my single cent.

abeldekat:
@igniscyan Thanks for the reply!

@dam9000, in my previous reply, I would like to rephrase my answer to your most important question:

> What tasks can one do before UiEnter?

I think it is not about what one can do *before* `UiEnter`. It's about how fast "the UI enters".  
In general I don't mind to wait a bit before having full editing capabilities(for example, the `lsp` and `completions`). I do want to see the code or the startupscreen almost immediately, so that I can read the code, think and navigate.

dam9000:
Out of curiosity I tested it on my pc and:
original:
048.315  000.002: --- NVIM STARTED ---
this change:
035.874  000.002: --- NVIM STARTED ---
so a 13 millisecond improvement, interesting :) ... but small. For this negligible gain I would rather keep the config simple.

shouldn't this improvement go to nvim core so that everyone benefits instead of having work-arounds in the configuration? Did you bring up this with the core neovim project?

abeldekat:
> so a 13 millisecond improvement, interesting :) ... but small. For this negligible gain I would rather keep the config simple.

That's a 27% decrease in startup-time by just using `vim.schedule`...)

Your environment is not the target of this PR. Applied to that environment, `kickstart` could be simplified even more by removing the following events from the spec: `InsertEnter`, `VimEnter`, `BufWritePre`. The human eye would not notice the change.

I know the statement above is a bit absurd. After all, those events make sense, and when the config expands each improvement counts.

On older hardware, the improvement can be very [noticeable](https://github.com/LazyVim/LazyVim/discussions/4112). According to [this](https://github.com/LazyVim/LazyVim/discussions/4112#discussioncomment-10105456) comment, even on a fairly recent setup using `wsl2`. 

> shouldn't this improvement go to nvim core so that everyone benefits instead of having work-arounds in the configuration? Did you bring up this with the core neovim project?

I did consider doing so. In `clipboard.vim`, `xsel` and `pbcopy` are handled a bit [differently](https://github.com/LazyVim/LazyVim/discussions/4112#discussioncomment-10095201). I think it's not trivial without breaking changes. 

After feedback in [this](https://www.reddit.com/r/neovim/comments/1e4je65/save_40_startuptime_for_free/) reddit post, I wanted this simple improvement to also be available for people that don't follow `Neovim` topics that closely.

My PR's in [LazyVim](https://github.com/LazyVim/LazyVim/pull/4120) and [Astronvim](https://github.com/AstroNvim/astrocore/pull/28) were accepted. For `nvchad` the solution was a little harder to apply. The maintainer decided that the change should be the [responsibility](https://github.com/NvChad/NvChad/pull/2940) of the user.

Last, but certainly not least, `kickstart` is the perfect target for this change in my opinion. Many people having their own `pde` use `kickstart`, either directly or as a reference.

dam9000:
I think this should be brought up to neovim core team:

Tested this change in core neovim runtime and the improvement is the same:
```
--- clipboard.vim-orig	2024-07-27 11:36:19.166852465 +0200
+++ clipboard.vim	2024-07-27 11:36:36.640979483 +0200
@@ -105,7 +105,7 @@
     let s:copy['*'] = s:copy['+']
     let s:paste['*'] = s:paste['+']
     return 'wayclip'
-  elseif !empty($DISPLAY) && executable('xsel') && s:cmd_ok('xsel -o -b')
+  elseif !empty($DISPLAY) && executable('xsel')
     let s:copy['+'] = ['xsel', '--nodetach', '-i', '-b']
     let s:paste['+'] = ['xsel', '-o', '-b']
     let s:copy['*'] = ['xsel', '--nodetach', '-i', '-p']
```
I take it by reverting from xclip back to xsel:
https://github.com/neovim/neovim/issues/20862 clipboard: prefer xsel before xclip
then this is again an issue:
https://github.com/neovim/neovim/issues/7237 Slow startup with clipboard.vim + xsel

I guess it could be fixed by the above change, or if ancient backward compatibility is an issue then the cmd_ok could be cached in a file instead of executing xsel every time. Or the whole clipboard.vim could be rescheduled after uienter in a similar way as you do it in the config.

ps. I'm just commenting, not trying to block this change - it's up to the maintainers anyway.

abeldekat:
> ps. I'm just commenting, not trying to block this change - it's up to the maintainers anyway.

No problem. It's good to have the discussion.

> ...  take it by reverting from xclip back to xsel: ...

Yes, I read those core issues, they were mentioned in the link I provided to the [discussion](https://github.com/LazyVim/LazyVim/discussions/4112) in LazyVim. They were one of the reasons I decided not to file a new issue.

> I guess it could be fixed by the above change...

I don't know whether that cmd_ok is there for historical reasons. However, that won't fix the problem with `pbcopy` on `macos`. Also, scheduling the whole `clipboard.vim` is a breaking change I guess. There would be no way to have the clipboard configured earlier when that behavior is required for some reason.

feoh:
OK I read the code in this PR and while it does add complexity I'm not sure it really raises the bar to entry too much for new users.

What I want to know is:

- You talk about Linux and Mac, what about Windows? Or Windows/WSL?
- Will your change break the Smartyank plugin? It's the only thing that keeps me from tearing my nearly non existent remaining hair out WRT Neovim clipboards.

If you can confidently answer that your proposed change will work under these additional environments, I'm all for it.

I'm also happy to do testing for these configurations as I have access to all. I run Smartyank by default and WSL is one of my primary environments along with Mac (I do run Ubuntu on the side as my mad scientist's lab, so I can test there too when I get around to it :)

abeldekat:
> You talk about Linux and Mac, what about Windows? Or Windows/WSL?

I don't know about windows without `wsl`. I did receive positive [feedback](https://github.com/LazyVim/LazyVim/discussions/4112#discussioncomment-10105456) from a `wsl2` user.

> Will your change break the Smartyank plugin?

No. I just tested this with your `smartyank` setup. The plugin is activated and I can copy to the system  clipboard. Your remaining hair won't be affected by this PR...;)
I also tested the `yanky` plugin, included as extra in `LazyVim`. 

> I'm also happy to do testing for these configurations as I have access to all.

Thank you! I would like to know on which of those platforms the improvement is most noticeable. I don't think this PR requires extensive testing( see also [Folke`s comment](https://github.com/LazyVim/LazyVim/discussions/4112#discussioncomment-10099287) )

EDIT: Is the comment above the change good enough?

PR #1057, opened 2024-07-26
Modify Conform comments to prevent deprecation warning when used

igniscyan:
At some point `Conform.nvim` switched from the sub-list syntax to including a `stop_after_first` argument. This PR updates the comment to use the new syntax so users don't get a deprecation warning after use.


rmacklin:
> At some point

Here's the commit: https://github.com/stevearc/conform.nvim/commit/9f111be14818c91832db8f320c4a4aa68de0e00b (part of https://github.com/stevearc/conform.nvim/pull/491)

PR #1061, opened 2024-07-28
Remove treesitter prefer_git option (unsafe!)

jstrot:
- It's not safe and can corrupt other git repos (when GIT_DIR and GIT_INDEX_FILE environment variables are set)
- nvim-treesiter maintainers consider `prefer_git` as deprecated and no longer needed.

See nvim-treesitter PR for details: https://github.com/nvim-treesitter/nvim-treesitter/pull/6959

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -856,8 +856,6 @@ require('lazy').setup({
     config = function(_, opts)
       -- [[ Configure Treesitter ]] See `:help nvim-treesitter`
 
-      -- Prefer git instead of curl in order to improve connectivity in some environments
-      require('nvim-treesitter.install').prefer_git = true
       ---@diagnostic disable-next-line: missing-fields
       require('nvim-treesitter.configs').setup(opts)
 


rmacklin:
For context, this setting was adding in https://github.com/nvim-lua/kickstart.nvim/pull/856.

Re:
> * It's not safe and can corrupt other git repos (when GIT_DIR and GIT_INDEX_FILE environment variables are set)

Wasn't this fixed by https://github.com/nvim-treesitter/nvim-treesitter/pull/6960 (which was recently merged, and is linked at the end of the PR you linked)?

Re:
> * nvim-treesiter maintainers consider `prefer_git` as deprecated and no longer needed.

I'm not sure if that's true. The comment was _"The actual resolution is deprecating `prefer_git`"_, which I don't interpret as _"`prefer_git` is deprecated"_ (rather, I interpreted it as proposing to resolve the issue by deprecating `prefer_git` - but that comment was made prior to opening https://github.com/nvim-treesitter/nvim-treesitter/pull/6960 to resolve the issue without deprecation).

rmacklin:
Hmm, just noticed nvim-treesitter has both a `master` branch _and_ a `main` branch, and the `main` branch includes https://github.com/nvim-treesitter/nvim-treesitter/commit/ad2cb2b6c943718c6f1e6ccd5d036cd5183dc25d. So I guess they did also decide to deprecate it even after fixing the issue?

rmacklin:
Hmm, just noticed nvim-treesitter has both a `master` branch _and_ a `main` branch, and the `main` branch includes https://github.com/nvim-treesitter/nvim-treesitter/commit/ad2cb2b6c943718c6f1e6ccd5d036cd5183dc25d. ~So I guess they did also decide to deprecate it even after fixing the issue?~

Edit: Ah, it seems that commit was actually authored `13 Jun 2024 12:30:57 +0200`, so it actually precedes https://github.com/nvim-treesitter/nvim-treesitter/pull/6960 by over a month. The `main` branch was rebased on `master` 12 hours ago, but GitHub uses the COMMITTER_DATE timestamp when showing _"committed 12 hours ago"_ here:

image

I've left a comment on that commit to see if it's still needed/wanted given the more recent merge of https://github.com/nvim-treesitter/nvim-treesitter/pull/6960.

rmacklin:
Ok, the maintainer does still seem to want to remove git support in nvim-treesitter 1.0. That means the diff here is probably appropriate - though I'd suggest changing the PR title/description to clarify that the reason for this PR is because the `prefer_git` option is being removed from nvim-treesitter, not because it's unsafe (because that issue was fixed in https://github.com/nvim-treesitter/nvim-treesitter/pull/6960)

Issue #1060, opened 2024-07-28
Code action keybinding isn't available in visual mode


byrmsh:
## Describe the bug

Code action keybinding `vim.lsp.buf.code_action` is not available in visual mode. This prevents using refactoring functionality such as "extract selected code into function". Is this intentional? If it is not, I can quickly send a PR. I searched for this on the issues and pull requests but couldn't find anything matching.

## To Reproduce

1. Select any code block in visual mode in a buffer with an LSP client running
2. Use default keybinding `<leader>ca` for code action
3. Selected code is deleted and mode is set to insert mode

## Desktop

- OS: Linux
- Terminal: Alacritty

## Neovim Version

```
NVIM v0.10.0
Build type: RelWithDebInfo
LuaJIT 2.1.1713773202
Run "nvim -V1 -v" for more info
```

iton0:
Yeah make a PR for this

VlaDexa:
> 3\. Selected code is deleted and mode is set to insert mode

That's the default behavior of `c` keybind.

Refactoring actions are a bit broken in nvim as of my test, they don't remove the selection and don't offer an automatic way to change function name when you extract into a function. If you seek a bit more proper support I recommend looking into [refactoring.nvim](https://github.com/ThePrimeagen/refactoring.nvim).

byrmsh:
> That's the default behavior of `c` keybind.

Yeah I am aware of this as it is one of the essential keybindings of Vim/Neovim. I just wanted to write the steps in detail.

> Refactoring actions are a bit broken in nvim as of my test, they don't remove the selection and don't offer an automatic way to change function name when you extract into a function. If you seek a bit more proper support I recommend looking into [refactoring.nvim](https://github.com/ThePrimeagen/refactoring.nvim).

[...]

PR #1064, opened 2024-07-30
Add visual mode support for LSP code action keymap (#1060)

byrmsh:
See #1060

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -279,7 +279,7 @@ require('lazy').setup({
 
       -- Document existing key chains
       require('which-key').add {
-        { '<leader>c', group = '[C]ode' },
+        { '<leader>c', group = '[C]ode', mode = { 'n', 'x' } },
         { '<leader>d', group = '[D]ocument' },
         { '<leader>r', group = '[R]ename' },
         { '<leader>s', group = '[S]earch' },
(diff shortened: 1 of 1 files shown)


init.lua
```diff
@@ -469,47 +469,47 @@ require('lazy').setup({
           -- to define small helper and utility functions so you don't have to repeat yourself.
           --
           -- In this case, we create a function that lets us more easily define mappings specific
-          -- for LSP related items. It sets the mode, buffer and description for us each time.
-          local map = function(keys, func, desc)
-            vim.keymap.set('n', keys, func, { buffer = event.buf, desc = 'LSP: ' .. desc })
+          -- for LSP related items. It sets the buffer and description for us each time.
+          local map = function(mode, keys, func, desc)
+            vim.keymap.set(mode, keys, func, { buffer = event.buf, desc = 'LSP: ' .. desc })
```
nhld, line 475: you could do this:
```lua
          local map = function(keys, func, desc, mode)
            mode = mode or 'n'
            vim.keymap.set(mode, keys, func, { buffer = event.buf, desc = 'LSP: ' .. desc })
          end

```
and set the mode `{ n, x }` for that one line without changing the rest:
```lua
          map('<leader>ca', vim.lsp.buf.code_action, '[C]ode [A]ction', { 'n', 'x' })

```

feoh: Hey @bayramkzk what do you think of this suggestion?

byrmsh: I think having mode at the end of arguments seems a bit weird but I agree that this commit changed more lines than it should and as this is a very small change in the config, I don't think there is any need to debate any more about it. Reverting and recommitting with minimal changes now!

PR #1070, opened 2024-08-01
fix: remove deprecated opt for conform.nvim

iton0:
fix: remove deprecated opt for conform.nvim and add space keymap

- changed lsp_fallback -> lsp_format
- updated format_on_save function to reflect change above

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -637,7 +637,7 @@ require('lazy').setup({
       {
         '<leader>f',
         function()
-          require('conform').format { async = true, lsp_fallback = true }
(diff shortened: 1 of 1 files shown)


feoh:
Removing deprecated opts I get, but why remove the space keymap?

iton0:
I thought it would stop the cursor from moving to the right as noted in the issue in my initial comment. However, after reviewing the keymap does not stop that behavior. I will update the commit to remove that.

Edit: it should be updated now just for the conform config

PR #1095, opened 2024-08-20
Check for loop or uv for lazypath

### LINKED ISSUE #1073: vim.uv field is nil causing error in init.lua
## Describe the bug

When trying to initialize and use the `uv` module in Neovim, I encountered an error indicating that the `vim.uv` field is `nil`. This issue arises when using the `fs_stat` function to check if a path exists.

## To Reproduce

1. Use Neovim version `v0.9.5`.
2. Add the following code to `init.lua`:
   ```lua
   if not vim.uv.fs_stat(lazypath) then
     local lazyrepo = 'https://github.com/folke/lazy.nvim.git'
     local out = vim.fn.system { 'git', 'clone', '--filter=blob:none', '--branch=stable', lazyrepo, lazypath }
     if vim.v.shell_error ~= 0 then
       error('Error cloning lazy.nvim:\n' .. out)
     end
   end
   vim.opt.rtp:prepend(lazypath)
   ```
3. Start Neovim.

**Expected Behavior:**
The `uv` module should be initialized correctly, allowing the use of `fs_stat` to check if a path exists.

**Actual Behavior:**
Neovim throws the following error:
```
E5113: Error while calling lua chunk: /home/kali/.config/nvim/init.lua:210: attempt to index field 'uv' (a nil value)
stack traceback:
        /home/kali/.config/nvim/init.lua:210: in main chunk
Press ENTER or type command to continue
```

## Desktop

- OS: Kali Linux
- Terminal: GNOME Terminal

## Neovim Version

```
NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1719379426

  fall-back for $VIM: "/usr/share/nvim"

```

**Additional Information:**
Changing `vim.uv` to `vim.loop` resolves the issue:
```lua
local uv = vim.loop

if not uv.fs_stat(lazypath) then
  local lazyrepo = 'https://github.com/folke/lazy.nvim.git'
  local out = vim.fn.system { 'git', 'clone', '--filter=blob:none', '--branch=stable', lazyrepo, lazypath }
  if vim.v.shell_error ~= 0 then
    error('Error cloning lazy.nvim:\n' .. out)
  end
end

vim.opt.rtp:prepend(lazypath)
```

Please let me know if this is the intended usage or if there are any additional configurations required.

LehMichael:
Fixes #1073
This is the suggested way from https://lazy.folke.io/installation and should also work on older nvim versions.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -207,7 +207,7 @@ vim.api.nvim_create_autocmd('TextYankPost', {
 -- [[ Install `lazy.nvim` plugin manager ]]
 --    See `:help lazy.nvim.txt` or https://github.com/folke/lazy.nvim for more info
 local lazypath = vim.fn.stdpath 'data' .. '/lazy/lazy.nvim'
-if not vim.uv.fs_stat(lazypath) then
+if not (vim.uv or vim.loop).fs_stat(lazypath) then
   local lazyrepo = 'https://github.com/folke/lazy.nvim.git'
   local out = vim.fn.system { 'git', 'clone', '--filter=blob:none', '--branch=stable', lazyrepo, lazypath }
   if vim.v.shell_error ~= 0 then


dam9000:
@mike-jl this LGTM.
I'd suggest you remove the *NOTE* block in the PR description, as it makes it look like it's a PR opened by mistake.

LehMichael:
Done! Thanks for the heads up!

VlaDexa:
Hi, sorry for noticing this so late, but I think this is just clutter, because our other config stuff like type hints or even plugins like conform don't support older nvim versions and require 0.10, where vim.loop is deprecated

feoh:
Should I publish a revert?

PR #1103, opened 2024-08-24
Add ripgrep as a telescope dependency

lucaSartore:
I was testing my config, and found out that the telescope `live_grep` builtin search did not work properly.

After a bit of digging I found out that ripgrep is a dependency of telescope, therefore I added it to the config.

Link to telescope documentation: [https://github.com/nvim-telescope/telescope.nvim?tab=readme-ov-file#suggested-dependencies](https://github.com/nvim-telescope/telescope.nvim?tab=readme-ov-file#suggested-dependencies)

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -340,6 +340,7 @@ require('lazy').setup({
     branch = '0.1.x',
     dependencies = {
       'nvim-lua/plenary.nvim',
+      'BurntSushi/ripgrep',
       { -- If encountering errors, see telescope-fzf-native README for installation instructions
         'nvim-telescope/telescope-fzf-native.nvim',
 


VlaDexa:
Ripgrep isn't a plugin and should be installed on the system. See [Install Recipes](https://github.com/nvim-lua/kickstart.nvim?tab=readme-ov-file#install-recipes) in the README

lucaSartore:
Wops… I messed up, sorry for that.

I may suggest adding a comment that reminds of this fact inside the dependencies of telescope, however if you think that that is unnecessary, and would make the config bloated with comments I can understand that.

Let me know if you want me to update this PR with the comment, or close it entirely.

feoh:
Thanks as others have pointed out this may not make sense. Thanks for your contribution but closing for now.

PR #1108, opened 2024-08-25
Enable silent option for default neo-tree plugin keybinding

byrmsh:
When the keybinding is spammed, the command ran for revealing neo-tree is displayed in cmdline and this is especially annoying while used with noice.nvim or similar configuration. Silent option is currently missing to fix this minor issue.


dam9000 (approved):
Nice!

PR #1109, opened 2024-08-26
Add nvim-window-picker for neo-tree plugin

chenlei0x:
neo-tree need  s1n7ax/nvim-window-picker plugin to open file in a picked window


feoh:
Thank you for your contribution but I think people can customize neotree on their own.

PR #1102, opened 2024-08-23
cleanup: refactor which-key configuration for cleaner setup

iton0:
- Moved `which-key` configuration from inline `config` to `opts` for better organization.
- Updated the key mappings setup to use `spec` for defining existing key chains.
- Removed deprecated or unnecessary comments and code.

This change aligns with updated `which-key` configuration practices, improving readability and maintainability as recommended by @VlaDexa in #1068.

PR #1113, opened 2024-08-26
Fix the which-key spec issue caused by recent cleanup

dam9000:
The recent cleanup accidentally broke the leader key specs because the spec block was in the wrong level of braces. That resulted in which-key no longer showing the description of the <leader> key chains such as [S]earch and others.

Fixes the issue introduced with: https://github.com/nvim-lua/kickstart.nvim/pull/1102


feoh:
Great catch. Thank you!

PR #1118, opened 2024-09-01
New config

weatr7:
**NOTE**


VlaDexa:
https://github.com/user-attachments/assets/5cf1b242-c9d2-49b3-bdb8-5e4256bcc68c

I think this PR was for your personal repo

bewuethr:
I'd love to contribute an Actions workflow that checks the PR description for some string that instructs authors to remove it, and if it's still in the description, the PR is automatically closed, and the workflow leaves a comment with an explanation. Would that be useful?

VlaDexa:
Ah, it would be, but sometimes people make legit PR's that include this text in the description. Make a draft PR, so we may discuss this with others there.

rmacklin:
@bewuethr I've been thinking about the same thing. I say go for it. To address @VlaDexa's concern, I think we can add a line to https://github.com/nvim-lua/kickstart.nvim/blob/a22976111e406ec0e4903ae78bf66a1fc0125b8a/.github/pull_request_template.md#L1-L7 that explicitly says something like "If you **did** intend to open a PR for contribution to the original project, please remove this text before opening your PR. Otherwise, it will be automatically closed."

PR #1123, opened 2024-09-03
Add undotree support to Kickstart.nvim

MandavkarPranjal:
This PR integrates Undotree support into the Kickstart.nvim configuration. The Undotree plugin provides a visual representation of the undo history, allowing users to easily navigate and manage their undo actions.  

Key Changes:
- Added configuration for the Undotree plugin to the Kickstart.nvim setup.
- Updated the plugin manager to include Undotree.
- Included basic key mappings for opening and interacting with the Undotree.  

Benefits:
- Enhances the user experience by providing a graphical interface for undo history.
- Improves workflow efficiency by making it easier to visualize and revert changes.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -910,8 +910,9 @@ require('lazy').setup({
 
   { -- Review undo history with a tree view
     'mbbill/undotree',
-    event = 'VeryLazy', -- Set the loading event to 'VeryLazy'
-    vim.keymap.set('n', '<leader>vu', ':UndotreeToggle<CR>', { desc = 'Toggle Undotree' }),
+    keys = {
+      {'<leader>vu', ':UndotreeToggle<CR>', desc = 'Toggle Undotree' }
+    }
   },
 
   -- The following two comments only work if you have downloaded the kickstart repo, not just copy pasted the
--- a/init.lua
+++ b/init.lua
@@ -908,6 +908,12 @@ require('lazy').setup({
     --    - Treesitter + textobjects: https://github.com/nvim-treesitter/nvim-treesitter-textobjects
   },
 
+  { -- Review undo history with a tree view
+    'mbbill/undotree',
+    event = 'VeryLazy', -- Set the loading event to 'VeryLazy'
+    vim.keymap.set('n', '<leader>vu', ':UndotreeToggle<CR>', { desc = 'Toggle Undotree' }),
+  },
+
   -- The following two comments only work if you have downloaded the kickstart repo, not just copy pasted the
   -- init.lua. If you want these files, they are in the repository, so you can just download them and
   -- place them in the correct locations.


init.lua
```diff
@@ -908,6 +908,12 @@ require('lazy').setup({
     --    - Treesitter + textobjects: https://github.com/nvim-treesitter/nvim-treesitter-textobjects
   },
 
+  { -- Review undo history with a tree view
+    'mbbill/undotree',
+    event = 'VeryLazy', -- Set the loading event to 'VeryLazy'
```
iton0: It might be better to remove the event and load the plugin only when using it's keybinding via the [keys](https://lazy.folke.io/spec/lazy_loading#%EF%B8%8F-lazy-key-mappings) spec.

VlaDexa:
Is this really an essential part for the 'starting point' config?

sodoku:
> Is this really an essential part for the 'starting point' config?

For me personally: no

MandavkarPranjal:
> Is this really an essential part for the 'starting point' config? 

I totally get the point! I think of `undotree` as a really handy tool, kind of like how `git` is for version control, but for your undo history. It helps visualize and manage changes, making it easier to track and recover edits. I believe it’s super useful for developers and can really enhance the workflow, which is why I see it as a great addition even for a 'starting config'!

sodoku:
I think the emphasise here is on "essential". In this regard kickstart is different than distributions like lazyvim which could also be a more feature full starting point.
Lsp configuration is part of kickstart as many people see that as a must in an editor _and_ it is not trivial to setup. Linting I personally see as a must, but it is not part of kickstart's init. I would probably never use undotree as _for me_ it is not useful.

What do you think about moving undotree to `lua/kickstart/plugins` like linting? This way it would not be part of the "essential" minimal kickstart config, but make it easier for people to discover undotree.

rmacklin:
Another question is, given that telescope is one of the core/[most prominent](https://github.com/search?q=repo%3Anvim-lua%2Fkickstart.nvim%20telescope&type=code) pieces of kickstart, would https://github.com/debugloop/telescope-undo.nvim perhaps be a better/more seamless fit? What are the pros and cons of leaning into a telescope integration vs an independent vim plugin?

gr8mtyr:
I think these plugins would help new commers understand that neovim doesn't have just a simple undo redo but to be able to pick a whole other undo history from the undo tree. so it would be good if some of these was added, preferably undotree as its the closest to visualize how it works.

shivambegin:
+1 for this , also i suggest to use "ut" as key-map because it's more user friendly for beginners.

feoh:
Anyone is welcome to add this in their own fork. I don't think it makes sense to impose this on everyone. I'm beginning to question whether we should even be including neotree as I've long since switched to oil :)

PR #1131, opened 2024-09-06
feat: update references of tsserver to ts_ls

Cheveniko:
nvim-lspconfig renamed `tsserver` to `ts_ls` (see this [PR](https://github.com/neovim/nvim-lspconfig/pull/3232))

This causes a warning everytime you start nvim because Mason hasn't changed `tsserver` name to `ts_ls`.

This PR fixes it by adding an if statement to check if the server name is `tsserver` and updates it to `ts_ls` in `mason-lspconfig.setup` [(Credits)](https://github.com/neovim/nvim-lspconfig/pull/3232#issuecomment-2331025714)

Also it replaces every reference of `tsserver` to `ts_ls`

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -614,8 +614,8 @@ require('lazy').setup({
         -- Some languages (like typescript) have entire language plugins that can be useful:
         --    https://github.com/pmizio/typescript-tools.nvim
         --
-        -- But for many setups, the LSP (`tsserver`) will work just fine
-        -- tsserver = {},
+        -- But for many setups, the LSP (`ts_ls`) will work just fine
+        -- ts_ls = {},
         --
 
         lua_ls = {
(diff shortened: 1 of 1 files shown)


init.lua
```diff
@@ -614,8 +614,8 @@ require('lazy').setup({
         -- Some languages (like typescript) have entire language plugins that can be useful:
         --    https://github.com/pmizio/typescript-tools.nvim
         --
-        -- But for many setups, the LSP (`tsserver`) will work just fine
-        -- tsserver = {},
+        -- But for many setups, the LSP (`ts_ls`) will work just fine
+        -- ts_ls = {},
```
BraedenKilburn, line 618: Shouldn't this remain `tsserver` until Mason updates to `ts_ls`? Otherwise, we'd be looking for `ts_ls` in Mason and then your fix on line 656 doesn't do anything?

Cheveniko: You're right, I missed that 😵

It should remain `tsserver` in case anyone uncomments it

shivambegin:
this is ready to merge as mason also renamed tsserver to ts_ls

samarth-na:
this should be added as soon as possible

Cheveniko:
Mason updated `tsserver` name to `ts_ls` [(See this issue)](https://github.com/williamboman/mason.nvim/issues/1784#issuecomment-2336761735) so the fix isn't necessary anymore.

I've updated the commit to remove the fix and only update every reference of `tsserver` to `ts_ls`

Issue #1141, opened 2024-09-13
nvim-treesitter config for Ruby on Rails is slow


jrupinski:
## Describe the bug

Current default config  for `nvim-treesitter` is slow on Ruby files (navigation has tiny lag on ~300 line files, probably smaller too)

It seems it's due to changes made for Ruby in [this commit ](https://github.com/nvim-lua/kickstart.nvim/commit/7892c0c354639985b1cf36f11d175201590e267b).

They probably made sense at the time, I don't think they're still required.
After reverting the changes, it seems to work as snappy as expected.
To be exact - `additional_vim_regex_highlighting = { 'ruby' }` seems to be the one causing slowdowns.

## To Reproduce

1. Install a fresh copy of `kickstart.nvim`
2. Open a `.rb` file, navigate around it using `hjkl` (should be couple hundred ms late)
3. open `init.lua`, revert the changes from [commit mentioned ](https://github.com/nvim-lua/kickstart.nvim/commit/7892c0c354639985b1cf36f11d175201590e267b), so `nvim-treesitter` config looks like this:
```
    opts = {
     (...)
      highlight = { enable = true },
      indent = { enable = true },
    },
```
4. Open the same `.rb` file, it should work much snappier.

## Desktop

- OS: `macOS 14.6.1 (23G93)`
- Terminal: `Kitty 0.35.2`

## Neovim Version
NVIM v0.10.1                                                                                                                                                                                        
Build type: Release                                                                                                                                                                                 
LuaJIT 2.1.1725453128

feoh:
This has nothing to do with kickstart.nvim - please file a bug with [treesitter](https://github.com/tree-sitter/tree-sitter).

rmacklin:
@feoh I think they're proposing that line should be removed from kickstart.nvim's config

feoh:
@rmacklin Thanks for pointing that out. Re-opening.

Now who's up for generating a pull request I can review? :) Issues are fine but diffs are spiffier :P

feoh:
@jrupinski Please just file a pull request and I'll merge it. This makes sense to me.

feoh:
Actually giving this a careful read, I'm unsure whether or not this request actually makes sense.

If I understand correctly, the line in question enables fuller syntax highlighting for Ruby regex syntax, at the expense of some startup time.

[...]

Issue #747, opened 2024-03-12
Ruby messes with indentation


ChillerDragon:
## Describe the bug

Kickstart messes with the indentation. It introduces wrong indentation without user confirmation.
This does not happen when doing ``nvim -u NONE``
I tried deleting lspconfig or auto formatter from init.lua and also activating the ruby_ls ruby-lsp from shopify. Nothing worked ._.

## To Reproduce

```
# test.rb

def foo
  # this should be indented
  # type in a.a in here or "anything".size and it will indent to the left
end
```

https://github.com/nvim-lua/kickstart.nvim/assets/20344300/3c4400ce-a1fe-4391-8aba-3f221d9892ab




## Desktop
- OS: linux
- Terminal: gnome

## Neovim Version

```
NVIM v0.9.5                                                                                                                                                                                                        
Build type: Release                                                                                                                                                                                                
LuaJIT 2.1.1692716794                                                                                                                                                                                              
                                                                                                                                                                                                                   
   system vimrc file: "$VIM/sysinit.vim"                                                                                                                                                                           
  fall-back for $VIM: "/__w/neovim/neovim/build/nvim.AppDir/usr/share/nvim"  
```

## Kickstart version

cb1f16b8cad0c06eac841a227825d9f23b4acba0

Can reproduce with unedited and also with auto formatter and lsp stripped out.

tjdevries:
I think this is probably an issue with tree-sitter indent, my guess is you could turn that off or disable it for ruby only. I'm not sure if it makes sense to change the base config for this though (hopefully someday it gets fixed upstream in tree-sitter)

calebdw:
You need to make sure you have the `ruby` grammar installed if you want to use nvim-treesitter indents with ruby

ChillerDragon:
> You need to make sure you have the `ruby` grammar installed if you want to use nvim-treesitter indents with ruby

How do I install that? Isn't the idea that kickstart does auto install what it needs?

The problem here is that it works fine in vanilla neovim and kickstart out of the box breaks it.

ChillerDragon:
> I think this is probably an issue with tree-sitter indent, my guess is you could turn that off or disable it for ruby only.

Indeed. Turning that off fixed it. But now I have to indent everything manually on line breaks. That is also worse than vanilla neovim. Interestingly it works on my mac but on debian it does not. Both systems have ruby installed via rbenv.

tjdevries:
[...]

PR #1164, opened 2024-09-26
Fix file navigation delay in Ruby caused by ts indent

jrupinski:
Related to / Fixes issue https://github.com/nvim-lua/kickstart.nvim/issues/1141.

Caused by `additional_vim_regex_highlighting = { 'ruby' }` line for Ruby in kickstart ts config that was introduced in [this commit ](https://github.com/nvim-lua/kickstart.nvim/commit/7892c0c354639985b1cf36f11d175201590e267b).

The problem this solved at the time seems redundant now (most likely fixed upstream in `treesitter`), and now introduces a small delay when navigating in files instead.

~~I reverted the commit - restoring default ts indenting for it, in turn making config simpler.~~
Edit: Only ts regex highlighting was removed to avoid regressions - thanks @ChillerDragon.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -892,7 +892,7 @@ require('lazy').setup({
       -- Autoinstall languages that are not installed
       auto_install = true,
       highlight = { enable = true },
-      indent = { enable = true },
+      indent = { enable = true, disable = { 'ruby' } },
     },
     -- There are additional nvim-treesitter modules that you can use to interact
     -- with nvim-treesitter. You should go explore a few and see what interests you:
--- a/init.lua
+++ b/init.lua
@@ -891,14 +891,8 @@ require('lazy').setup({
       ensure_installed = { 'bash', 'c', 'diff', 'html', 'lua', 'luadoc', 'markdown', 'markdown_inline', 'query', 'vim', 'vimdoc' },
       -- Autoinstall languages that are not installed
       auto_install = true,
-      highlight = {
-        enable = true,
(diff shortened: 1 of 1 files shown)


DanWlker:
Do we need to set additional_vim_regex_highlighting to false?

(Edit: nvm I checked the default value is false)

ChillerDragon:
Does this cause a regression of https://github.com/nvim-lua/kickstart.nvim/issues/747?

jrupinski:
@ChillerDragon I was not aware of this - it does indeed. I pushed a change to disable ts indenting for Ruby again to fix this regression. Thanks!

Removal of `additional_vim_regex_highlighting = { 'ruby' }` does not seem to regress anything AFAIK, and doesn't cause other indent issues like this one: https://github.com/nvim-treesitter/nvim-treesitter/issues/6114

Also - seems nvim-treesitter has highlight rules of their own:
https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/ruby/highlights.scm

ChillerDragon:
nice thanks

jrupinski:
Update: I did some learning and made a nvim setup from scratch, then went back and re-did kickstart and applied my changes properly.
Turns out this change is not the cause of slowdown, most likely it was incorrectly setup LSP
(as I had timeout issues with it as well). No longer an issue, and it's smooth sailing now.

I don't believe this removal is required anymore, and since it's now slowing down the setup considerably I think it may as well stay.  👍

PR #1193, opened 2024-10-17
Set breakpoint icons and their highlight colors

glmlm:
Set breakpoint icons and their highlight colors in `lua/kickstart/plugins/debug.lua`; see [Here](https://github.com/mfussenegger/nvim-dap/discussions/355) for details


glmlm:
My apologies, I made an error in the operation and deleted the repository. I will submit a new request later.

PR #1194, opened 2024-10-18
Set breakpoint icons and their highlight colors

glmlm:
Set breakpoint icons and their highlight colors in `lua/kickstart/plugins/debug.lua`; see [Here](https://github.com/mfussenegger/nvim-dap/discussions/355) for details

### DIFF
--- a/lua/kickstart/plugins/debug.lua
+++ b/lua/kickstart/plugins/debug.lua
@@ -89,6 +89,19 @@ return {
       },
     }
 
+    -- Change breakpoint icons
+    -- For more information, see https://github.com/mfussenegger/nvim-dap/discussions/355
+    vim.api.nvim_set_hl(0, 'Break', { fg = '#e51400' })
+    vim.api.nvim_set_hl(0, 'Stop', { fg = '#ffcc00' })
+    local breakpoint_icons = vim.g.have_nerd_font
+        and { Breakpoint = '', BreakpointCondition = '', BreakpointRejected = '', LogPoint = '', Stopped = '' }
+      or { Breakpoint = '●', BreakpointCondition = '⊜', BreakpointRejected = '⊘', LogPoint = '◆', Stopped = '⭔' }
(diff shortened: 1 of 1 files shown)


iton0:
I think this should be a overall comment that users can uncomment if they want this change.

lua/kickstart/plugins/debug.lua
```diff
@@ -89,6 +89,19 @@ return {
       },
     }
 
+    -- Change breakpoint icons
+    -- For more information, see https://github.com/mfussenegger/nvim-dap/discussions/355
+    vim.api.nvim_set_hl(0, 'Break', { fg = '#e51400' })
+    vim.api.nvim_set_hl(0, 'Stop', { fg = '#ffcc00' })
+    local breakpoint_icons = vim.g.have_nerd_font
+        and { Breakpoint = '', BreakpointCondition = '', BreakpointRejected = '', LogPoint = '', Stopped = '' }
```
iton0: I think it be better to just have one table of icons

glmlm: Thank you for reviewing the code. With all due respect, that might not be possible.
If `vim.g.have_nerd_font` is `false`, the local variable `breakpoint_icons` would no longer exist.

Error that occurs when deleting the line:
```
Cannot assign `boolean|table` to parameter `<T:table>`.
   - `boolean` cannot match `table`
   - Type `boolean` cannot match `table` [param-type-mismatch]
```

feoh:
Hi guys I have no feelings on this.

@iton0 Can you help me understand why you think this should just be optional? 

Do you use/do other people use different conditional break symbols?

iton0:
I appreciate the overall addition, but I wanted to make it optional for caution's sake. This way, users can experience the default debugging setup first and choose to add features if they wish, rather than having to remove anything.

feoh:
Ah I failed to notice it's already commented out. merging!

PR #1195, opened 2024-10-18
Change diagnostic symbols if vim.g.have_nerd_font is true

glmlm:
Diagnostic symbols on sign column is now displayed graphically when the nerd font is enabled.
The reference page is [here](https://github.com/neovim/nvim-lspconfig/wiki/UI-customization#change-diagnostic-symbols-in-the-sign-column-gutter).

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -588,6 +588,15 @@ require('lazy').setup({
         end,
       })
 
+      -- Change diagnostic symbols in the sign column (gutter)
+      if vim.g.have_nerd_font then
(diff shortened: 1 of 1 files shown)


iton0:
I think this should be a overall comment that users can uncomment if they want this change.

glmlm:
That might indeed be the case. I will update soon.

feoh:
Love this. thanks for commenting it out so we can avoid the "You moved my cheese!" crowd :)

PR #1196, opened 2024-10-18
Optionally add statuscol.nvim

glmlm:
The visibility of the default statuscolumn is not very flattering; only one of diagnostics, gitsigns, and breakpoints can be displayed, and foldcolumn displays a lot of unnecessary symbols.
Therefore, I propose to provide optional settings with a plugin that improve its appearance; [statuscol.nvim](https://github.com/luukvbaal/statuscol.nvim).
<details>
<summary>Note: Subsequent PR was moved to #1198 </summary>
I also considered it really annoying to set the foldings every time I open a file, so I have added a plugin that set them automatically like VSCode; nvim-ufo.

I understand that adding them is controversial, especially from a "kickstart" perspective.
Personally, I think it is worth adding them if **only as an option**, as I believe it would be inconvenient for VSCode users to be without them when they first touch Neovim.
</details>


iton0:
To avoid overwhelming new users, it's important to keep the setup simple and compact. While additional plugins can enhance functionality, they may complicate the initial experience. If we add these plugins, we should also consider the needs of users switching from other IDEs. While these features could be useful, I believe it's better to keep them out of the repository and let users discover them independently.

glmlm:
@iton0
> it's important to keep the setup simple and compact

Obviously, I understand the importance of keeping it to a minimum. Before adding a plug-in, even if it is an optional one, its necessity should be carefully considered.
Based on that line of thinking, it might have been better to split this pull request into at least two parts. I personally think it would be more user-friendly to have automatic folding, but there might be quite a few people who don't need it.
However, I think there is a little more room for further consideration on statuscol.nvim, in my opinion. It is sort of counter-intuitive that only one of the diagnostic and gitsigns symbols are displayed, even though they are configured to be displayed in init.lua.

I would appreciate your candid feedback.

iton0:
> Based on that line of thinking, it might have been better to split this pull request into at least two parts.

I agree that splitting the PR may be a better approach. 

> It is sort of counter-intuitive that only one of the diagnostic and gitsigns symbols are displayed,

That is true. I would read the docs on the [statuscolumn](https://neovim.io/doc/user/options.html#'statuscolumn') because you may be able to configure it in such a way that a plugin is not needed. For example with the issue that you pointed out, you can increase the width of the signcolumn to display more signs.

glmlm:
> may be able to configure it in such a way that a plugin is not needed

It would be appreciated if this problem could be solved without using a plugin, but it appears that it is not that simple.
vim.opt.signcolumn = 'yes:2' simply increases the upper limit of the display, and if there are more than two diagnostics, the symbol of gitsigns will no longer be displayed.
There is also a problem with the method of displaying foldings neatly without a plugin. It is mentioned [here](https://github.com/kevinhwang91/nvim-ufo/issues/4#issuecomment-2241159750), but I don't think it is a smart way to go about it because it is to increase the display limit as above.

cmdrrobin:
I agree with @iton0, this may overwhelm new users and I think this plugin is more a personal taste to use than for everybody. I advise to keep kickstart.nvim as simple as possible, with a minimum of plugins.

feoh:
Hi. Thank you for your contribution, feel free to make this change in your own fork, but as the community has pointed out this change probably isn't appropriate for kickstart.

Closing.

PR #1198, opened 2024-10-19
Add nvim-ufo as an option

glmlm:
It is really annoying for me to set the foldings every time I open a file, so I have looked for a plugin that set them automatically like VSCode; [nvim-ufo](https://github.com/kevinhwang91/nvim-ufo) and added it.

I understand that adding this is controversial, especially from a "kickstart" perspective, but I think it is worth adding them if **only as an option**.

I often use this feature when coding to hide parts of the code that I don't need to see, and I personally believe it would be inconvenient for VSCode users to be without them when they first touch Neovim. Howerver, I think the opinion that there is no need to bother putting it in lua/kickstart/plugins if there seems to be no demand for it from many users should also be respected.

I would appreciate your thoughts on this.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -922,6 +922,7 @@ require('lazy').setup({
   -- require 'kickstart.plugins.lint',
   -- require 'kickstart.plugins.autopairs',
   -- require 'kickstart.plugins.neo-tree',
+  -- require 'kickstart.plugins.nvim-ufo',
   -- require 'kickstart.plugins.gitsigns', -- adds gitsigns recommend keymaps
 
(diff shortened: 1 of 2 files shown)


feoh:
Thank you for your contribution, but there are many Neovim users who do not wish to clone every VSCode behavior in their Neovim environment.

I'm closing. If others in the community feel very strongly about this please re-open and make a case for why it should be included, along with grid coordinates so you can be the human meat shield when the 90000 trad neovim users pop out of the primordial ooze to whine when it gets merged.

PR #1200, opened 2024-10-20
fix(lsp): remove BufEnter event on linting

cmdrrobin:
When in use, linting will also be triggered when viewing lsp signature help (or lsp hover)

If you enter a hover- or signature help window, markdownlint triggers warninging within the LSP window. This is very annoying and shouldn't be triggered.

Here is a screenshot what happens when entering a LSP window.

image

### DIFF
--- a/lua/kickstart/plugins/lint.lua
+++ b/lua/kickstart/plugins/lint.lua
@@ -44,7 +44,7 @@ return {
       -- Create autocommand which carries out the actual linting
       -- on the specified events.
       local lint_augroup = vim.api.nvim_create_augroup('lint', { clear = true })
-      vim.api.nvim_create_autocmd({ 'BufEnter', 'BufWritePost', 'InsertLeave' }, {
+      vim.api.nvim_create_autocmd({ 'BufWritePost', 'InsertLeave' }, {
         group = lint_augroup,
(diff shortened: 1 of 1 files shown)


ccjmne:
Hmm, that's pretty smart. It's something that's been bothering me as well!
But then I suppose that you won't quite get diagnostics as you purposefully open a buffer whose linting you care for... Not until you enter and leave insert mode or write your buffer to a file, is that correct?

It may be better to identify somehow that the buffer you just entered is actually a pop-up one, and only disable these.

ccjmne:
We could consider skipping the linting for read-only buffers. It works nicely, but may be undesirable in some cases.

cmdrrobin:
> Hmm, that's pretty smart. It's something that's been bothering me as well! But then I suppose that you won't quite get diagnostics as you purposefully open a buffer whose linting you care for... Not until you enter and leave insert mode or write your buffer to a file, is that correct?
> 
Yes that is correct.

> It may be better to identify somehow that the buffer you just entered is actually a pop-up one, and only disable these.
>
That would be much better solution, but I didn't know how.

cmdrrobin:
I will close this PR, in favor for #1202

PR #1203, opened 2024-10-22
fix : fish cannot execute this command

marsevilspirit:
```bash
git clone https://github.com/nvim-lua/kickstart.nvim.git "${XDG_CONFIG_HOME:-$HOME/.config}"/nvim
```

This command cannot be executed in fish. It is recommended to modify it to the following form:

```bash
bash -c 'git clone https://github.com/nvim-lua/kickstart.nvim.git "${XDG_CONFIG_HOME:-$HOME/.config}/nvim"'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ too - it's ignored in the kickstart repo to make maintenance easier, but it's
 <details><summary> Linux and Mac </summary>
 
 ```sh
-git clone https://github.com/nvim-lua/kickstart.nvim.git "${XDG_CONFIG_HOME:-$HOME/.config}"/nvim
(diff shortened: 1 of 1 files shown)


feoh:
I don't love this approach.

Is there a fish equivalent to the $XDG_HOME variable expansion that we could provide as an alternative "If Using fish shell"  instruction in the README?

feoh:
The more I think about this, the more I think this isn't actually appropriate fir kickstart.

What about Nushell? Elvish? tcsh? csh?

Shells are legion.  bash/zsh represent a reasonable mainline target to support.

Please, folks in the community please reopen if you feel strongly otherwise.

PR #1210, opened 2024-10-27
Simplify extended plugin comment.

winder:
Simplify and clarify the extended plugin configuration comment.

This change could also simply remove "two", but I found most of the original comment to be unnecessary.


feoh:
thank you for your contribution, but respectfully, you don't have to answer the umpteen issues per week from people who watched TJ's old video and just copied in the init.lua and watched it blow up :(

I'm closing this. Someone please lmk if you feel very strongly I'm in the wrong.

PR #1216, opened 2024-11-01
Fix nvim-dap not lazy loading

kaezrr:
Debugging related plugins should "lazy-load" only when debugging keys are pressed, but the default configuration kickstart provides prevents it from lazy-loading.

The `keys` property had local variables `dap` and `dap-ui` that used `require` and prevented all DAP related plugins from lazy-loading. Fixed this by changing keys to a table and substituting the local variables with a anonymous function.

nvim-dap and all dap related plugins now load only when any debugging related keybinds are pressed. 
Reduces neovim start-up time by around 100ms (depending on your hardware).

### DIFF
--- a/lua/kickstart/plugins/debug.lua
+++ b/lua/kickstart/plugins/debug.lua
@@ -24,28 +24,59 @@ return {
     -- Add your own debuggers here
     'leoluz/nvim-dap-go',
   },
-  keys = function(_, keys)
-    local dap = require 'dap'
-    local dapui = require 'dapui'
-    return {
-      -- Basic debugging keymaps, feel free to change to your liking!
-      { '<F5>', dap.continue, desc = 'Debug: Start/Continue' },
-      { '<F1>', dap.step_into, desc = 'Debug: Step Into' },
-      { '<F2>', dap.step_over, desc = 'Debug: Step Over' },
-      { '<F3>', dap.step_out, desc = 'Debug: Step Out' },
-      { '<leader>b', dap.toggle_breakpoint, desc = 'Debug: Toggle Breakpoint' },
-      {
-        '<leader>B',
-        function()
-          dap.set_breakpoint(vim.fn.input 'Breakpoint condition: ')
-        end,
-        desc = 'Debug: Set Breakpoint',
-      },
-      -- Toggle to see last session result. Without this, you can't see session output in case of unhandled exception.
-      { '<F7>', dapui.toggle, desc = 'Debug: See last session result.' },
-      unpack(keys),
-    }
-  end,
+  keys = {
+    -- Basic debugging keymaps, feel free to change to your liking!
+    {
+      '<F5>',
+      function()
+        require('dap').continue()
+      end,
+      desc = 'Debug: Start/Continue',
+    },
+    {
+      '<F1>',
+      function()
+        require('dap').step_into()
+      end,
+      desc = 'Debug: Step Into',
+    },
+    {
+      '<F2>',
+      function()
+        require('dap').step_over()
+      end,
+      desc = 'Debug: Step Over',
+    },
+    {
+      '<F3>',
+      function()
+        require('dap').step_out()
+      end,
+      desc = 'Debug: Step Out',
+    },
+    {
+      '<leader>b',
+      function()
+        require('dap').toggle_breakpoint()
+      end,
+      desc = 'Code Debug: Toggle Breakpoint',
+    },
+    {
+      '<leader>B',
+      function()
+        require('dap').set_breakpoint(vim.fn.input 'Breakpoint condition: ')
+      end,
+      desc = 'Code Debug: Set Breakpoint',
+    },
+    -- Toggle to see last session result. Without this, you can't see session output in case of unhandled exception.
+    {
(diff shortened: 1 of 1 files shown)


lua/kickstart/plugins/debug.lua
```diff
@@ -24,28 +24,59 @@ return {
...
+        require('dap').step_out()
+      end,
+      desc = 'Debug: Step Out',
+    },
+    {
+      '<leader>b',
+      function()
+        require('dap').toggle_breakpoint()
+      end,
+      desc = 'Code Debug: Toggle Breakpoint',
```
iton0: for the toggling of regular and conditional breakpoint can you update it such that they both start with "Debug:" instead of "Code Debug:". 
This just makes sure that this is consistent with the other debugging keymaps. 
other than that lgtm

cmdrrobin (commented):
Good point! You could rewrite it into more readable and without repeating the require() statements.

```lua
  keys = function()
    local dap = require('dap')
    local dapui = require('dapui')
    local keys = {
      -- Basic debugging keymaps, feel free to change to your liking!
      { '<F5>', dap.continue, desc = 'Debug: Start/Continue' },
      { '<F1>', dap.step_into, desc = 'Debug: Step Into' },
      { '<F2>', dap.step_over, desc = 'Debug: Step Over' },
      { '<F3>', dap.step_out, desc = 'Debug: Step Out' },
      { '<leader>b', dap.toggle_breakpoint, desc = 'Debug: Toggle Breakpoint' },
      {
        '<leader>B',
        function()
          dap.set_breakpoint(vim.fn.input('Breakpoint condition: '))
        end,
        desc = 'Code Debug: Set Breakpoint',
      },
      -- Toggle to see last session result. Without this, you can't see session output in case of unhandled exception.
      { '<F7>', dapui.toggle, desc = 'Debug: See last session result.' },
    }

    return keys
  end,
```

(Sorry, for the edits. I should have tested it first!)

kaezrr:
@rgruyters No actually, if you have those local variables `dap` and `dap-ui`, it is preventing lazy from somehow evaluating what are the keys for `nvim-dap`  without loading it first, that's why I replaced all the local variables with `require`s even tho it is kind of repetitive. 

Check if your code lazy loads `nvim-dap` properly (i.e it shouldn't load at all unless you press any of the debug keybinds), it doesn't on my machine. If it does on yours then I will assume there maybe some issue with my configuration and close this pull request

cmdrrobin:
@kaezrr looks like you are correct! Didn't know this behavior.

iton0 (changes requested):

lua/kickstart/plugins/debug.lua
kaezrr: Done 👍

lua/kickstart/plugins/debug.lua
```diff
@@ -24,28 +24,59 @@ return {
...
+      end,
+      desc = 'Debug: Step Out',
+    },
+    {
+      '<leader>b',
+      function()
+        require('dap').toggle_breakpoint()
+      end,
+      desc = 'Debug: Toggle Breakpoint',
+    },
```
glmlm, line 63: I would like to suggest that we try this approach:
```
    { '<F5>', <cmd>DapContinue<CR>, desc = 'Debug: Start/Continue' },
    { '<F1>', <cmd>DapStepInto<CR>, desc = 'Debug: Step Into' },
    { '<F2>', <cmd>DapStepOver<CR>, desc = 'Debug: Step Over' },
    { '<F3>', <cmd>DapStepOut<CR>, desc = 'Debug: Step Out' },
    { '<leader>b', <cmd>DapToggleBreakpoint<CR>, desc = 'Debug: Toggle Breakpoint' },
```

glmlm (commented):
@rgruyters
>could rewrite it into more readable

From this point of view, the following approach might be useful.

lua/kickstart/plugins/debug.lua
kaezrr: Ok I have tested this and this works ok, and lazy loads nvim-dap but it is missing the `<F7>` and `<leader>B` keybinds and I dont know the commands for those. (dap-ui toggle and set breakpoint condition respectively)

glmlm: Unfortunately, there does not seem to be a user command defined for them at this time.
There are several ways to do like:
* Create a user command
```
vim.api.nvim_create_user_command('DapUiToggle', function()
  require('dapui').toggle()
end, { nargs = 0 })
return {
  ...
    { '<F7>', '<cmd>DapUiToggle<CR>', desc = 'Debug: See last session result.' },
  ...
}
```
* Write as a command of lua function
```
return {
  ...
    { '<F7>', '<cmd>lua require("dapui").toggle()<CR>', desc = 'Debug: See last session result.' },
  ...
}
```
* Keep the status quo

I'm not going to be involved in the decision which you choose, and there might well be a better way of doing it.
In any case, it would be best to leave the final decision to the community.

feoh:
Thanks for this and thanks to everyone for the good back and forth around the details! Merging.

PR #1227, opened 2024-11-08
fix: which-key comment typo

mihasket:
Found a typo in `init.lua` when I was reading the notes for which-key.nvim.

Replaced `whick-key.nvim` with `which-key.nvim`.


feoh:
Ah, the dread whick key! :) Thanks for the fix.

PR #1232, opened 2024-11-14
Prepare for upcoming deprecation of configuring diagnostic-signs using sign_define()

glmlm:
Since Neovim 0.10, it seems that configuring diagnostic-signs using sign_define() has been [deprecated](https://neovim.io/doc/user/deprecated.html#_checkhealth).
There is another way to set them as follows:
```
if vim.g.have_nerd_font then
  vim.diagnostic.config {
    signs = {
      text = {
        [vim.diagnostic.severity.ERROR] = '',
        [vim.diagnostic.severity.WARN] = '',
        [vim.diagnostic.severity.INFO] = '',
        [vim.diagnostic.severity.HINT] = '',
      },
    },
  }
end
```
In the committed one, I simply focused on fewer lines and readability(after all, this section is commented out).

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -590,11 +590,12 @@ require('lazy').setup({
 
       -- Change diagnostic symbols in the sign column (gutter)
       -- if vim.g.have_nerd_font then
-      --   local signs = { Error = '', Warn = '', Hint = '', Info = '' }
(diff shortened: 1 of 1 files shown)


feoh:
Thanks for this! Merging.

Issue #1253, opened 2024-12-05
Highlighting the word under the cursor


denius:
Hi,

I'm confused by this part of the config:
https://github.com/nvim-lua/kickstart.nvim/blob/8d1ef972bc32faa86fee21a57f9033b41f612ebb/init.lua#L555

This line and the following lines define functions for highlighting words under the cursor. But I know that this functionality already exists in the plugin https://github.com/RRethy/vim-illuminate.

Maybe it's worth defining this functionality in the additional plugin section `lua/kickstart/plugins` with the corresponding config?

I'm using a configuration that exactly matches the current functionality of the `init.lua` configuration; this configuration is taken entirely from the plugin's page:

```lua
  -- highlight (underline) the word under the cursor
  -- https://github.com/RRethy/vim-illuminate
  {
    'RRethy/vim-illuminate',
    opts = {},
    config = function()
      require('illuminate').configure {
        -- providers: provider used to get references in the buffer, ordered by priority
        providers = {
          'lsp',
          'treesitter',
          'regex',
        },
        -- delay: delay in milliseconds
        delay = 120,
        -- filetype_overrides: filetype specific overrides.
        -- The keys are strings to represent the filetype while the values are tables that
        -- supports the same keys passed to .configure except for filetypes_denylist and filetypes_allowlist
        filetype_overrides = {},
        -- filetypes_denylist: filetypes to not illuminate, this overrides filetypes_allowlist
        filetypes_denylist = {
          'dirvish',
          'fugitive',
          'alpha',
          'NvimTree',
          'lazy',
          'neogitstatus',
          'Trouble',
          'lir',
          'Outline',
          'spectre_panel',
          'toggleterm',
          'DressingSelect',
          'TelescopePrompt',
        },
        -- filetypes_allowlist: filetypes to illuminate, this is overridden by filetypes_denylist
        filetypes_allowlist = {},
        -- modes_denylist: modes to not illuminate, this overrides modes_allowlist
        modes_denylist = {},
        -- modes_allowlist: modes to illuminate, this is overridden by modes_denylist
        modes_allowlist = {},
        -- providers_regex_syntax_denylist: syntax to not illuminate, this overrides providers_regex_syntax_allowlist
        -- Only applies to the 'regex' provider
        -- Use :echom synIDattr(synIDtrans(synID(line('.'), col('.'), 1)), 'name')
        providers_regex_syntax_denylist = {},
[...]

PR #1265, opened 2024-12-13
Replace custom highlighting with plugin RRethy/vim-illuminate

denius:
This is a PR for #1253, it replaces the word highlighting functions with a plugin that has more customization options.

The main reason for this PR is that the highlighting functionality in kickstart.nvim is incompatible with the use of highlighting plugins, as there is no way to disable the built-in highlighting other than commenting it out entirely.

The configuration is completely taken from the plugin page [RRethy/vim-illuminate](https://github.com/RRethy/vim-illuminate#configuration).

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -664,7 +664,7 @@ require('lazy').setup({
         filetype_overrides = {},
         -- filetypes_denylist: filetypes to not illuminate, this overrides filetypes_allowlist
         filetypes_denylist = {
-        'dirbuf',
+          'dirbuf',
           'dirvish',
           'fugitive',
         },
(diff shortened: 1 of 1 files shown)


iton0:
I think introducing a plugin for cursor word highlighting doesn't fit in with kickstart. However I think given the incompatibility issue, it would be beneficial to update the comments for this code. This way new users can be made aware and either keep the default or comment out and add whichever highlighting plugins in their own configs.

feoh:
@denius Thank you for your PR. I know I suggested it, but I think @iton0 is right and the way to go here is to just suggest that users have the option to use this in comments.

denius:
I realized that it is better to do without an extra plugin if possible.

PR #1286, opened 2024-12-23
chore(gitsigns): Remove visual mappings from select mode

NextMerge:
Using v makes mappings work in visual and select mode. Using x instead of v makes mappings work only in visual mode which is in my opinion correct. Select mode is mostly used with snippets and current mappings might interfere with them.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -319,7 +319,7 @@ require('lazy').setup({
         { '<leader>s', group = '[S]earch' },
         { '<leader>w', group = '[W]orkspace' },
         { '<leader>t', group = '[T]oggle' },
-        { '<leader>h', group = 'Git [H]unk', mode = { 'n', 'v' } },
+        { '<leader>h', group = 'Git [H]unk', mode = { 'n', 'x' } },
       },
     },
   },
--- a/lua/kickstart/plugins/gitsigns.lua
+++ b/lua/kickstart/plugins/gitsigns.lua
@@ -34,10 +34,10 @@ return {
 
         -- Actions
         -- visual mode
-        map('v', '<leader>hs', function()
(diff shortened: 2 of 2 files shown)


feoh:
Hi!

What does this PR accomplish and why is this desirable? Could you please add some flavor text so I can understand what I'm merging?

kuator:
> Hi!
> 
> What does this PR accomplish and why is this desirable? Could you please add some flavor text so I can understand what I'm merging?

Using `v` makes mappings work in visual and select mode. Using `x` instead of `v` makes mappings work only in visual mode which is in my opinion correct. Select mode is mostly used with snippets and current mappings might interfere with them.

feoh:
@iton0 @dam9000 @benfleis Hi all I have zero opinions on this. What do you think?

Note that in order to change a keymap like this I'd want to see pretty strong feelings from most folks, otherwise we can leave things as they are.

Thanks and apologies for picking on you all. I REALLY appreciate the help!!!!!!

iton0:
Follow the keymap setup recommended by gitsigs.nvim. If no recommendation is provided, keep the default settings.

feoh:
Thanks all!

Sorry @NextMerge I'm afraid your PR isn't a good fit for this project at this time. Feel free to do this in your fork!

NextMerge:
For some reason I wasn't getting notifications from this PR until just now.

I opened this for the reasons @kuator mentioned. Depending on which leader key you set those mappings can't be used in select mode. The leader key set right now in the config is `<space>`, and attempting to use the gitsigns binds in select mode just results in inserting a space.

With that said I agree with @iton0. I'll see if gitsigns would want to update their recommended maps first.

PR #1287, opened 2024-12-24
Sync vim diagnostic local list with LSP.

mazdabat:
This update ensures synchronization between the local Vim diagnostic list and the LSP.

The issue arose with the _clangd_ LSP, where the local diagnostic list retained syntax errors reported by _clangd_ even after they were resolved and _clangd_ no longer reported them.

While it's unclear if this problem exists for other programming languages, this PR effectively addresses the issue for _clangd_ when working with C/C++.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -165,7 +165,9 @@ vim.opt.scrolloff = 10
 vim.keymap.set('n', '<Esc>', '<cmd>nohlsearch<CR>')
 
 -- Diagnostic keymaps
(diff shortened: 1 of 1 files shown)


iton0:
The behavior is working as expected. Location list does not automatically update when errors are fixed; its like a static snapshot of the diagnostics of the file. Additionally, the open option for vim.diagnostic.setloclist is true by default.

feoh:
Hi. I'm with @iton0 here, but if you think we're in error, please make a case for why and re-open.

Issue #1285, opened 2024-12-23
[Not a bug] Proposal to set up GitHub Discussions


glmlm:
I have been looking at the Issue Tracker for this project over the past few months and there seems to be a certain segment of the community that seems to think that it is a free support forum for users.
To maintain the quality of the Issue Tracker, it would be preferable to direct these kinds of issues to the forum and allow users to resolve them among themselves.
I understand the sensitivity of this matter, but I would appreciate your input.

iton0:
**TL;DR**: While I agree with the idea of adding Discussions, I think the core issue is that new users aren't reading documentation. Two areas to address: improving the issue template to prompt users to review docs, and increasing awareness of Kickstart documentation.

---

I agree with the sentiment, but I have hesitation and two concerns that might be contributing to the issue:

1. **Improve the issue template**: Add a checklist to confirm if users have read the docs. This would encourage research and help collaborators identify those who’ve put in the effort to read the issue template.
  
2. **Review Kickstart documentation**: The issue isn’t necessarily poor documentation but lack of exposure. We need to find ways to increase awareness of the Kickstart documentation.

By addressing these points, we can better evaluate whether adding Discussions is the right approach.

feoh:
@glmlm I know where you're coming from here. I've often thought that having Discussions where I could ask questions about a project could be helpful to me as an end user.

But try on a different pair of shoes for a moment if you will. I am the only one with merge privs still paying attention with any regularity here, and we have a few regular contributors who help me understand proposed changes where I feel out of my depth.

First, I don't have the power to do what you're asking, you'd need to petition the repo owners, who don't have the time to participate here much..

Anyway, I'm not necessarily averse to this if others feel it's the right fit for this project as well, and if we can get the Powers that Be to enable them.

I just know I don't have any more time to contribute to this particular project than I spend already :)

feoh:
Given what I just wrote, I'm going to close this for now. Conditions that would make re-opening make sense are:

- Broad interest in the community
- Finding a way to get folks with the power to enable this to do so

Until then, I'm closing this. Thanks for understanding.

iton0:
@feoh I think it may be appropriate to take another look at this and get @tjdevries input about adding Discussions as @glmlm suggested. Particularly, I think doing an approach similar to [ghostty](https://github.com/ghostty-org/ghostty/issues) would be more effective compared to the checklist that I suggested.

feoh:
[...]

PR #1288, opened 2024-12-27
chore: add pre-issue requirements

iton0:
Based on #1285


feoh:
Love this! Thanks so much :)

Now if we could just get people to ACTUALLY READ IT!!! :)

Issue #1297, opened 2025-01-02
mason setup is called twice


tomasgareau:
When setting up `nvim-lspconfig`, we tell Lazy that `mason.nvim` is a dependency and configure it right away (#865 & linked issues touch on why):

https://github.com/nvim-lua/kickstart.nvim/blob/a8f539562a8c5d822dd5c0ca1803d963c60ad544/init.lua#L456-L460

But later in `nvim-lspconfig`'s `config` function, we explicitly setup `mason` again:

https://github.com/nvim-lua/kickstart.nvim/blob/a8f539562a8c5d822dd5c0ca1803d963c60ad544/init.lua#L648-L654

These two setup calls can interfere with each other. For example, if you tweaked the second setup call to _append_ Mason's bin path (e.g., to allow overriding Mason binaries with system binaries):

```lua
require("mason").setup({
  PATH = "append"
})
```

the _first_ setup call will prepend the Mason bin path (the default behaviour) and your change will append the bin path. Your path then ends up being:

```
/path/to/mason/bin:...:/rest/of/your/path:...:/path/to/mason/bin
```

Oops haha.

I _think_ we can just bin the second `setup` call and put all the mason configuration in the `dependencies` block:

```diff

  {
    -- Main LSP Configuration
    'neovim/nvim-lspconfig',
    dependencies = {
      -- Automatically install LSPs and related tools to stdpath for Neovim
-      { 'williamboman/mason.nvim', config = true }, -- NOTE: Must be loaded before dependants
+      {
+        "williamboman/mason.nvim",
+        opts = {
+	  -- Mason must be loaded before dependants so we need to configure it here.
+	  -- You can override its default settings by adding them to this table, e.g.:
+	  -- PATH = "append" -- append Mason's bin path so system packages take precedence
+	  PATH = "append",
+        },
+      },
      'williamboman/mason-lspconfig.nvim',
      'WhoIsSethDaniel/mason-tool-installer.nvim',

      -- etc
    },
    config = function()
      -- etc...
    
      -- Ensure the servers and tools above are installed
      --  To check the current status of installed tools and/or manually install
      --  other tools, you can run
      --    :Mason
      --
      --  You can press `g?` for help in this menu.
-      require('mason').setup()
```

(but I'm not sure if e.g., the second setup call is required for auto-installation _after_ the other config). Happy to toss up a PR if the above makes sense!

iton0:
[...]

PR #1298, opened 2025-01-03
fix: prevent mason setup from being run twice

tomasgareau:
Addresses https://github.com/nvim-lua/kickstart.nvim/issues/1297

Currently, we're calling `require('mason').setup(...)` twice:
* once when setting it as a dependency of `nvim-lspconfig` (since we set `config = true`)
* once in the `config` function we define for `nvim-lspconfig`

Calling setup twice can cause issues with, e.g., setting the `PATH` option: you might append Mason's bin dir in one setup call and prepend it in the other.

We've kept the setup of `mason` in the `nvim-lspconfig` dependencies table since leaving it to the `config` function caused some plugin-loading-order related issues in the past. See:
* https://github.com/nvim-lua/kickstart.nvim/pull/210
* https://github.com/nvim-lua/kickstart.nvim/issues/554
* https://github.com/nvim-lua/kickstart.nvim/pull/555
* https://github.com/nvim-lua/kickstart.nvim/pull/865

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -457,7 +457,16 @@ require('lazy').setup({
     'neovim/nvim-lspconfig',
     dependencies = {
       -- Automatically install LSPs and related tools to stdpath for Neovim
-      { 'williamboman/mason.nvim', config = true }, -- NOTE: Must be loaded before dependants
+      {
+        'williamboman/mason.nvim',
+        -- Mason must be loaded before its dependents so we'll define `opts`
+        -- here: this will make `lazy` automatically run
+        -- `require('mason').setup(opts)` when `nvim-lspconfig` loads.
+        opts = {
+          -- Add overrides for mason's default settings here if needed! e.g.:
+          -- log_level = vim.log.levels.DEBUG,
+        },
+      },
       'williamboman/mason-lspconfig.nvim',
       'WhoIsSethDaniel/mason-tool-installer.nvim',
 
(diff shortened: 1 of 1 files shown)


init.lua
```diff
@@ -457,7 +457,16 @@ require('lazy').setup({
     'neovim/nvim-lspconfig',
     dependencies = {
       -- Automatically install LSPs and related tools to stdpath for Neovim
-      { 'williamboman/mason.nvim', config = true }, -- NOTE: Must be loaded before dependants
+      {
+        'williamboman/mason.nvim',
+        -- Mason must be loaded before its dependents so we'll define `opts`
+        -- here: this will make `lazy` automatically run
+        -- `require('mason').setup(opts)` when `nvim-lspconfig` loads.
```
iton0: It might be better to adapt the NOTE from the fidget plugin here.

init.lua
```diff
@@ -457,7 +457,16 @@ require('lazy').setup({
...
+        opts = {
+          -- Add overrides for mason's default settings here if needed! e.g.:
```
iton0: I would remove this comment and just make it `opts = {},`

iton0 (approved):
LGTM

feoh:
Thank you for this @tomasgareau you rock!

And thank you @iton0 for the review! It looked reasonable to me but I don't really have time to dive down the rabbit hole and understand its goodness :)

PR #1299, opened 2025-01-06
move lines up or down

PhilBrk8:
Added keymap to move lines up or down with Alt+k / Alt+j

**NOTE**
I really like this feature in vs code and I think it is quite helpful and simplifies my life.
I hope that others will benefit as well from it if it is a default in the kickstarter.

Kind regards, and thank you for the kickstarter and advent of neovim series - helped me a ton.


iton0:
I like the idea, but I think it's best not to add more keymaps. If a user already has these keymaps for different functionality, it could lead to conflicts. If you're set on including this feature, I’d recommend adding the keymaps as a comment instead to avoid any potential issues.

PhilBrk8:
Thanks for the advice.
I used the kickstarter as starting point for a fresh nvim, so I didn't think about that.

feoh:
Hi. Thanks for this contribution but I don't really think this is a good fit for every Neovim user.

PR #1301, opened 2025-01-07
Remove the additional_vim_regex_highlighting clause which is discouraged.

feoh:
As per #1141 

Remove the additional_vim_regex_highlighting clause which is discouraged.

https://www.reddit.com/r/neovim/comments/yxjrkr/treesitter_syntax_highlighting_too_slow_on_large/

[I am creating this PR against the main nvim-lua/kickstart.nvim because I maintain my own incompatible fork.]

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -912,9 +912,10 @@ require('lazy').setup({
         -- Some languages depend on vim's regex highlighting system (such as Ruby) for indent rules.
         --  If you are experiencing weird indenting issues, add the language to
         --  the list of additional_vim_regex_highlighting and disabled languages for indent.
-        additional_vim_regex_highlighting = { 'ruby' },
+        -- additional_vim_regex_highlighting = { 'ruby' },
       },
-      indent = { enable = true, disable = { 'ruby' } },
+      -- indent = { enable = true, disable = { 'ruby' } },
+      indent = { enable = true },
     },
     -- There are additional nvim-treesitter modules that you can use to interact
     -- with nvim-treesitter. You should go explore a few and see what interests you:


JaredSharplin:
Hi @feoh , thanks for making this PR.

I'm just a user of `kickstart` and don't have much knowledge of treesitter. All I can say is anecdotally having commented out `additional_vim_regex_highlighting` its noticeably reduced lag for me with vim motions, etc.

I also haven't personally experienced the issue in #747 with indenting either.

BALAJI24092001 (approved):
It's better to include it as a choice rather than a mandatory inclusion. Doesn't change much and works faster with out it.

Thanks for the PR, I just checked it out and it seems to work better now.

bewuethr:
Doing this change breaks indentation in my Ruby codebase and results in error messages. For example, here

```ruby
module Pack
  class Index
    HEADER_SIZE = 8
    FANOUT_SIZE = 1024

    OID_LAYER = 2
    CRC_LAYER = 3
    OFS_LAYER = 4
    EXT_LAYER = 5

    def initialize(input)
      @input = input
      load_fanout_table
    end

    def oid_offset(oid)
    # <-- cursor goes here

    private def load_fanout_table
      @input.seek(HEADER_SIZE)
      @fanout = @input.read(FANOUT_SIZE).unpack("N256")
    end
  end
end
```

the method body is not indented if I don't add additional regex highlighting for Ruby and disable treesitter indent for Ruby.

Edit: probably red herring... a recent change in the Ruby LSP introduced what I'm seeing: https://github.com/Shopify/ruby-lsp/issues/3019

Sooo, I have no opinion on this PR 😬

JaredSharplin:
Out of curiousity I tried to replicate your issue @bewuethr . Just saw your edit though so all good, I'm not using `ruby-lsp` so that might be it:

https://github.com/user-attachments/assets/4dea0e46-29b3-4206-a30c-8c8c49096211

feoh:
@bewuethr That's what I was afraid of. Thanks for the report. We're gonna leave it as-is.Peeps who want better perf and aren't using the LSP can comment it out :)

bewuethr:
Sorry, to be clear, I think I was affected by an unrelated issue from ruby-lsp, but I'll report back when they have fixed that issue!

feoh:
So we SHOULD merge this change?

Please approve the PR if you think this change is OK.

danarnold:
I don't think this is related to the ruby-lsp issue. I've used nvim-kickstart and ruby-lsp for a while now (since before the release of ruby-lsp that caused the issue linked above), and although I've noticed the slowness that's caused by the kickstart lines in question and I've tried disabling them, I've had to keep them enabled because otherwise it breaks indentation in Ruby files.

bewuethr:
> So we SHOULD merge this change?

I think this happened:

- I applied the change in this PR to test it out
- At the same time, I was using an updated, buggy ruby-lsp version for the first time
- Indentation wasn't working correctly for me (see code block above)
- I got weird LSP errors

I think the LSP errors are due to the ruby-lsp bug; they go away after downgrading. The indentation seems to break with the change from *this* PR, so based on that, I would recommend not merging it.

On the other hand, @JaredSharplin was not able to reproduce my problem, so now I'm not sure 🤷🏻‍♂️

If I have time, I'll try combinations of applying this change while using ruby-lsp and not using it, to see if ruby-lsp affects indentation.

feoh:
Thanks all. I tried to be helpful here but I'm getting lots of mixed signals so I'll close this again :)

danarnold:
@bewuethr Now that the error in ruby-lsp has been fixed in a newly released version, I made the changes in this PR and I was able to reproduce the indenting error exactly as you described it. I'm also not sure why @JaredSharplin wasn't able to reproduce it, but yeah, it definitely wasn't related to the issue with ruby-lsp.

JaredSharplin:
Apologies for the hassle. If nobody else has experienced the issue with navigation lag then its best to leave it. I more just wanted to leave a comment on the original issue in case others were experiencing the same problem, which doesn't seem to be the case.

For reference I'm using `sorbet-lsp` and have kept `indent = { enable = true, disable = { 'ruby' } }` as is: https://github.com/JaredSharplin/kickstart-modular.nvim/blob/master/lua/kickstart/plugins/treesitter.lua

Appreciate the assistance 🙏  @feoh

feoh:
No worries and no hassle :)

PR #1303, opened 2025-01-09
lazydev.nvim: Use luals 3rd library for luv

diorman:
lazydev.nvim: Follows installation instructions as updated in https://github.com/folke/lazydev.nvim/commit/f125453e2b2c615eeeba96f321590b969869e27c which:

- Changes path for loading luvit types
- Removes the dependency on https://github.com/Bilal2453/luvit-meta

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -447,11 +447,10 @@ require('lazy').setup({
     opts = {
       library = {
         -- Load luvit types when the `vim.uv` word is found
-        { path = 'luvit-meta/library', words = { 'vim%.uv' } },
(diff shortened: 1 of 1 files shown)


iton0 (approved):
LGTM

Bilal2453:
Do note, the built-in luv definitions are deprecated and will be removed from the language server, they are migrating to the addon manager, see https://github.com/LuaLS/lua-language-server/discussions/1950.

Though they still haven't been migrated, not sure when will that happen. Likely when the addon manager is added to the actual language server binary.

PR #1307, opened 2025-01-13
chore: remove redundant comment

nhld:
This is a duplicate comment as it's already mentioned in the mason part.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -464,7 +464,6 @@ require('lazy').setup({
       'WhoIsSethDaniel/mason-tool-installer.nvim',
 
       -- Useful status updates for LSP.
-      -- NOTE: `opts = {}` is the same as calling `require('fidget').setup({})`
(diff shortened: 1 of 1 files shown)


BALAJI24092001 (changes requested):
For anyone who's starting Neovim with Kickstart might consider this comment useful. I didn't understand when I started, and this helped me.

nhld:
@BALAJI24092001 

The recent PR #1298 added a note with the same concept, as the mason config is now using `opts = {}` - this is now the first occurrence of using opts with an empty table in `init.lua`. So the fidget's note seems kinda redundant no?

https://github.com/nvim-lua/kickstart.nvim/blob/db4867acb939b548cf3c60641e1661783551248d/init.lua#L459-L469

BALAJI24092001 (approved):

PR #1316, opened 2025-01-19
docs: clarify using opts = {} vs config = function() ... require('plu…

bleacheda:
The current documentation mentioning that using "require" is equivalent to using "opts" without detailing the use in the "config = function()" block seems inaccurate.

Lower in the configuration the "config = function()" block is used without clarifying why it is needed and what it does.

This clarification may help new users understand the difference between the two, or how and where to place the "require" statement in order for it to be equivalent to using "opts".

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -250,7 +250,7 @@ require('lazy').setup({
   --
   -- Here is a more advanced example where we pass configuration
   -- options to `gitsigns.nvim`.
-  --  
+  --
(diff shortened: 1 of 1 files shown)


BALAJI24092001 (approved):
The added documentation seems helpful and will be easy for new users to understand and work with the setup better.

feoh:
Thank you. Great work!

Issue #1319, opened 2025-01-20
Deprecated Functions in Gitsigns


ioneyed:
Updating to the latest gitsigns there are some functions that were deprecated.

Specifically this commit - https://github.com/lewis6991/gitsigns.nvim/commit/8b74e560f7cba19b45b7d72a3cf8fb769316d259

>  - Deprecated config.show_deleted and toggle_deleted()
>  - Use preview_hunk_inline()
>  - Deprecated undo_stage_hunk()
>  - Use stage_hunk() on staged lines
>  - Updated keymap suggestion in README.md
>  - Removed hidden option config._inline2

Changes to the initial config may be in order

Deprecated Functions

https://github.com/nvim-lua/kickstart.nvim/blob/5bdde24dfb353d365d908c5dd700f412ed2ffb17/lua/kickstart/plugins/gitsigns.lua#L47

Suggested by lsp is to replace this with `gitsigns.stage_hunk()` on staged signs however the gitsigns plugin just removed this mapping entirely.

https://github.com/nvim-lua/kickstart.nvim/blob/5bdde24dfb353d365d908c5dd700f412ed2ffb17/lua/kickstart/plugins/gitsigns.lua#L57

Suggested by lsp is to reeplace this with `gitsigns.preview_hunk_inline()`

feoh:
Thanks for the PR! Closing as complete.

PR #1321, opened 2025-01-21
fix (#1319): gitsigns deprecated functions

ErlanRG:
- This commit change two functions that are marked as deprecated now: `gitsigns.stage_hunk` and `gitsigns.toggle_deleted`

### DIFF
--- a/lua/kickstart/plugins/gitsigns.lua
+++ b/lua/kickstart/plugins/gitsigns.lua
@@ -44,7 +44,7 @@ return {
         map('n', '<leader>hs', gitsigns.stage_hunk, { desc = 'git [s]tage hunk' })
         map('n', '<leader>hr', gitsigns.reset_hunk, { desc = 'git [r]eset hunk' })
         map('n', '<leader>hS', gitsigns.stage_buffer, { desc = 'git [S]tage buffer' })
-        map('n', '<leader>hu', gitsigns.undo_stage_hunk, { desc = 'git [u]ndo stage hunk' })
+        map('n', '<leader>hu', gitsigns.stage_hunk, { desc = 'git [u]ndo stage hunk' })
         map('n', '<leader>hR', gitsigns.reset_buffer, { desc = 'git [R]eset buffer' })
         map('n', '<leader>hp', gitsigns.preview_hunk, { desc = 'git [p]review hunk' })
         map('n', '<leader>hb', gitsigns.blame_line, { desc = 'git [b]lame line' })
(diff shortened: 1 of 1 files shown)


BALAJI24092001 (approved):

feoh:
My minimal testing saw no problems. Thanks for the contribution! Merging.

b4shful:
@ErlanRG @feoh
Can you take a closer look at this and see if it's definitely correct? Particularly the `desc` parameters.

```diff
--- a/lua/kickstart/plugins/gitsigns.lua
+++ b/lua/kickstart/plugins/gitsigns.lua
@@ -44,7 +44,7 @@
-        map('n', '<leader>hu', gitsigns.undo_stage_hunk, { desc = 'git [u]ndo stage hunk' })
+        map('n', '<leader>hu', gitsigns.stage_hunk, { desc = 'git [u]ndo stage hunk' })
```
You have changed undo_stage_hunk to stage_hunk, should the description be changed too? 

Actually looking into it a bit, I think stage hunk is already mapped, so perhaps if it undoes/redoes using the same function, we could just have one keybind for this? Something like "git stage/unstage hunk".

```diff
@@ -54,7 +54,7 @@
-        map('n', '<leader>tD', gitsigns.toggle_deleted, { desc = '[T]oggle git show [D]eleted' })
+        map('n', '<leader>tD', gitsigns.preview_hunk_inline, { desc = '[T]oggle git show [D]eleted' })
```
You have changed toggle_deleted to preview_hunk_inline. I'm not sure if this is supposed to be a functional replacement, or just replacing it with a new mapping due to the old command being deprecated.

But either way, unless preview_hunk_inline happens to toggle show deleted, then the description should really be changed too.

dam9000:
@b4shful I agree this does not seem to be a completely correct change, perhaps it would be worth checking if the obsoleted functions have an actual replacement, or if this functionality should be removed all together.

feoh:
Thanks I guess I merged too quickly.

Would anyone be willing to spin up a new PR fixing the desc problems you see?

b4shful:
> Thanks I guess I merged too quickly.
> 
> Would anyone be willing to spin up a new PR fixing the desc problems you 

Lots going on at university right now, haven't had time to do anything unfortunately. Hence the request below which otherwise I would do myself.

Could somebody spin this out into an issue so that it doesn't get lost in here as comments on a PR? 

(yes, I did just open an issue on something unrelated, but I haven't opened one for this because I don't actually use gitsigns and would have to research the plugin and what the different functions are supposed to do before I knew what to even write haha)

PR #1335, opened 2025-02-01
feat(diagnostic): add diagnostic config

gelocraft:
This PR adds diagnostic configuration for `kickstart.nvim`:
- diagnostic signs
- diagnostic virtual_text
- diagnostic severity_sort
- diagnostic underline
- diagnostic float

See `:help vim.diagnostic.Opts`

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -589,15 +589,34 @@ require('lazy').setup({
         end,
       })
 
-      -- Change diagnostic symbols in the sign column (gutter)
-      -- if vim.g.have_nerd_font then
-      --   local signs = { ERROR = '', WARN = '', INFO = '', HINT = '' }
-      --   local diagnostic_signs = {}
-      --   for type, icon in pairs(signs) do
-      --     diagnostic_signs[vim.diagnostic.severity[type]] = icon
-      --   end
-      --   vim.diagnostic.config { signs = { text = diagnostic_signs } }
-      -- end
+      -- Diagnostic Config
+      -- See :help vim.diagnostic.Opts
+      vim.diagnostic.config {
+        severity_sort = true,
+        float = { border = 'rounded', source = 'if_many' },
+        underline = { severity = vim.diagnostic.severity.ERROR },
+        signs = vim.g.have_nerd_font and {
+          text = {
+            [vim.diagnostic.severity.ERROR] = '󰅚 ',
+            [vim.diagnostic.severity.WARN] = '󰀪 ',
+            [vim.diagnostic.severity.INFO] = '󰋽 ',
+            [vim.diagnostic.severity.HINT] = '󰌶 ',
+          },
+        } or {},
+        virtual_text = {
+          source = 'if_many',
+          spacing = 2,
+          format = function(diagnostic)
+            local diagnostic_message = {
+              [vim.diagnostic.severity.ERROR] = diagnostic.message,
+              [vim.diagnostic.severity.WARN] = diagnostic.message,
+              [vim.diagnostic.severity.INFO] = diagnostic.message,
+              [vim.diagnostic.severity.HINT] = diagnostic.message,
+            }
+            return diagnostic_message[diagnostic.severity]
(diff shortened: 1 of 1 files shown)


gelocraft:
Do you like my code Mr. @tjdevries ?

feoh:
Hi!

Thanks for your contribution.

I'm having some real trouble understanding what visual difference this change makes, and whether or not the wider community would welcome it.

Could you attach screenshots of how your change makes things different please?

@tjdevries is still officially here, but only comes floating down from his OCaml cloud once a year or so to bless us with an awesome rewrite and then vanish again :)

(Love ya @tjdevries if you're reading this :)

gelocraft:
> Hi!
> 
> Thanks for your contribution.

Re: Hi thank you for noticing my PR :)

> Could you attach screenshots of how your change makes things different please?

Re: this is how the diagnostic signs would look like when `vim.g.have_nerd_font = true`

Screenshot_20250217-213809

Re: diagnostic signs when `vim.g.have_nerd_font = false`

Screenshot_20250217-213853

this PR adds new configuration launch point for **diagnostics** from **0.10 deprecation notice**: 
> [!Warning]
> Configuring `diagnostic-signs` using `:sign-define` or `sign_define()`. 
> Use the "signs" key of `vim.diagnostic.config()` instead.

gelocraft:
See `:help deprecated-0.10` and `:help vim.diagnostic.Opts`

gelocraft:
> I'm having some real trouble understanding what visual difference this change makes, and whether or not the wider community would welcome it.

you can checkout this PR locally and `export NVIM_APPNAME=kickstart.nvim` for a quick review on your local machine. :)

feoh:
Thank you. Love this!!!

ronandalton:
@gelocraft Hi, I was just wondering about the purpose of these lines in your pull request:
```lua
format = function(diagnostic)
  local diagnostic_message = {
    [vim.diagnostic.severity.ERROR] = diagnostic.message,
    [vim.diagnostic.severity.WARN] = diagnostic.message,
    [vim.diagnostic.severity.INFO] = diagnostic.message,
    [vim.diagnostic.severity.HINT] = diagnostic.message,
  }
  return diagnostic_message[diagnostic.severity]
end,
```

I'm pretty sure the array lookup here is unnecessary as each value is the same (all `diagnostic.message`). You could just do `return diagnostic.message` without defining the array, but even then I think that's unnecessary as the default nvim behavior seems to just print `diagnostic.message` anyway.

Would it be reasonable to make a follow-up pull request to remove these lines or is there a reason they should stay?

gelocraft:
> @gelocraft Hi, I was just wondering about the purpose of these lines in your pull request:
> 
> ```lua
> format = function(diagnostic)
>   local diagnostic_message = {
>     [vim.diagnostic.severity.ERROR] = diagnostic.message,
>     [vim.diagnostic.severity.WARN] = diagnostic.message,
>     [vim.diagnostic.severity.INFO] = diagnostic.message,
>     [vim.diagnostic.severity.HINT] = diagnostic.message,
>   }
>   return diagnostic_message[diagnostic.severity]
> end,
> ```
> 
> I'm pretty sure the array lookup here is unnecessary as each value is the same (all `diagnostic.message`). You could just do `return diagnostic.message` without defining the array, but even then I think that's unnecessary as the default nvim behavior seems to just print `diagnostic.message` anyway.
> 
> Would it be reasonable to make a follow-up pull request to remove these lines or is there a reason they should stay?

it allows user to customize the diagnostic message for different diagnostic severity.

example, adding an icon before the diagnostic message is displayed. but u can just return the diagnostic message if u r not planning to have different formats for each severity.

ronandalton:
I see. But the user could find that out themselves by reading `:help vim.diagnostic.Opts`. This code just seems unnecessary to me (especially since it isn't doing anything). I also wonder how likely the user is to want to change the diagnostic message formatting anyway.

PR #1353, opened 2025-02-15
Remove wildly inaccurate OS specific install steps and point at the excellent offocial docs intead!

feoh:
The old platform / OS specificd install instructions were wildly incorrect along a number of axes. The official Neovim installation docs are excellent now. Leverage those instead.

(Note: As I've mentioned in the past, I'm generating a PR from this repo because I already run my own personal fork that has diverged a ton from master and can't generate PRs from there without a hand picked nightmare :)


VlaDexa:
We have some extra dependencies, like `fd` or `ripgrep`, on top of just Neovim. Maybe we can leave instructions for them and delete only neovim specific instructions?

feoh:
@VlaDexa I like that idea but it's unclear to me how to thread that needle.

Would you consider proposing an alternative PR? You can even use the Github web UI to create it! :)

feoh:
This problem really needs a fix but this isn't the right solution as it throws out too much.

Issue #1305, opened 2025-01-12
Warning showing when requiring the ('mason.lspconfig').setup


nehamand:
## Before Reporting an Issue
- I have read the kickstart.nvim README.md.
- I have read the appropiate plugin's documentation.
- I have searched that this issue has not been reported before.

- [x ] **By checking this, I confirm that the above steps are completed. I understand leaving this unchecked will result in this report being closed immediately.**

## Describe the bug

On line 669 on the init.lua file, there is a warning stating that the mason.lspconfig setup is missing required fields in type `MasonLspconfigSettings`: `ensure_installed`, `automatic_installation` [missing-fields]. 

## To Reproduce

1. Just clone the repo and open it in neovim, go the line 669 and you will see the lua-lsp pointing to the warning. 

## Desktop

- OS: Mac Os Sequoia 15.1
- Terminal: Warp

## Neovim Version

```
0.10.3
```
<img width="1356" alt="Captura de Tela 2025-01-12 às 00 24 06" src="https://github.com/user-attachments/assets/367439f4-cb59-48c7-8f05-9b785611ba4b" />

BALAJI24092001:
I have the same issue, the LSP for lua seems to generate those warnings. There won't be any problem with the highlighted warnings. But I wouldn't want them to pop on screen either. I added the ensure_installed = {} which removed one warning, but the other one I didn't. It should probably work. Add them and leave them blank.

nhld:
Hi,
You might want to uncomment this line:
https://github.com/nvim-lua/kickstart.nvim/blob/db4867acb939b548cf3c60641e1661783551248d/init.lua#L639-L645
Hope this helps!

feoh:
@nhld Should we consider uncommenting this by default?

I feel like we get a constant stream of these issues, all well meant, and I don't blame folks for filing them, but they're kinda noise.

What's the down side if any?

BALAJI24092001:
I suppose uncommenting is better. It could be confusing for beginners to see unnecessary warnings. It doesn't seem to have any downside so far for me.

feoh:
No, the warnings are essentially meaningless, but they cause well meaning folks to file bug after bug after bug :)

Let me ask the same question from a different angle: Does having these warnings benefit anyone ever?

BALAJI24092001:
They do. 
[...]

PR #1354, opened 2025-02-15
Propsed fix for init.lua warnings without disabling diagnostics

feoh:
Proposed fix as per https://github.com/nvim-lua/kickstart.nvim/issues/1305#issuecomment-2657770325

FOLKS SHOULD TEST THIS THOROUGHLY. I'm not having any issues but I'd love more feedback please.

**NOTE**

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -675,6 +675,8 @@ require('lazy').setup({
       require('mason-tool-installer').setup { ensure_installed = ensure_installed }
 
       require('mason-lspconfig').setup {
+        ensure_installed = {}, -- explicitly set to an empty table (Kickstart populates installs via mason-tool-installer)
+        automatic_installation = false,
         handlers = {
           function(server_name)
             local server = servers[server_name] or {}


Jessy616:
8853a60e-7dde-46d9-bd1d-0441516872fd

Jessy616:
IMG_9606

Jessy616:
Hi Good evening James, how are you dokng ?

init.lua
```diff
@@ -675,6 +675,8 @@ require('lazy').setup({
       require('mason-tool-installer').setup { ensure_installed = ensure_installed }
 
       require('mason-lspconfig').setup {
```
jrentlez, line 677: `setup_handlers()` could also be used instead of `setup()`. See `:h mason-lspconfig.setup_handlers()`

feoh: Thank you I'll look into that.

Do you have any thoughts on the key point from the issue about the concerns that turning off auto-install might break something?

jrentlez: AFAIK, mason-lspconfig just [merges the user-config with a table containing default values](https://github.com/williamboman/mason-lspconfig.nvim/blob/1a31f824b9cd5bc6f342fc29e9a53b60d74af245/lua/mason-lspconfig/settings.lua#L30). For `automatic_installation` that default value seems to be `false`. So auto-install is already implicitly disabled.

feoh:
Lots of good commentary here, and thanks for all of it!

Unless I hear "Don't merge this. You'll break the world!" I'm going to merge tomorrow, because people keep opening bugs against these warnings and it really would be great to silence them :)

Jessy616:
Hi

PR #1356, opened 2025-02-16
fix: arguments for the `vim.lsp.Client.supports_method()` method

jrentlez:
This PR fixes the calls to `vim.lsp.Client.supports_method()`. The `:` automatically passes `client` as the first parameter.

**NOTE**

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -564,7 +564,7 @@ require('lazy').setup({
           --
           -- When you move your cursor, the highlights will be cleared (the second autocommand).
           local client = vim.lsp.get_client_by_id(event.data.client_id)
-          if client and client.supports_method(vim.lsp.protocol.Methods.textDocument_documentHighlight) then
+          if client and client:supports_method(vim.lsp.protocol.Methods.textDocument_documentHighlight) then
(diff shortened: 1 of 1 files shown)


feoh:
Awesome thanks for the fix! Merging.

glaulher:
for me it gave a typing alert and null. I needed to correct it as follows:


```
 if client ~= nil and client.supports_method ~= nil and client:supports_method(vim.lsp.protocol.Methods.textDocument_documentHighlight) then
            local highlight_augroup = vim.api.nvim_create_augroup('kickstart-lsp-highlight', { clear = false })
            vim.api.nvim_create_autocmd({ 'CursorHold', 'CursorHoldI' }, {
              buffer = event.buf,
              group = highlight_augroup,
              callback = vim.lsp.buf.document_highlight,
            })


```
here too:

```
if client ~= nil and client.supports_method ~= nil and client:supports_method(vim.lsp.protocol.Methods.textDocument_inlayHint) then
            map('<leader>th', function()
              vim.lsp.inlay_hint.enable(not vim.lsp.inlay_hint.is_enabled { bufnr = event.buf })
            end, '[T]oggle Inlay [H]ints')
          end
        end,
      })


```

Issue #1331, opened 2025-01-27
Use blink.cmp over nvim-cmp?


makew0rld:
Thanks for creating Kickstart, it's been a great way for me to start using Neovim for the first time.

One of the only things I've noticed missing is getting function signatures to appear when I'm typing the arguments of a function. I was eventually able to solve this using mini.completions, but when I posted about this [on Reddit](https://old.reddit.com/r/neovim/comments/1i8zeo8/i_finally_got_function_signatures_working/)  people suggested using [blink.cmp](https://github.com/Saghen/blink.cmp) which I hadn't heard of before.

Do you have any thoughts on blink and using it in Kickstart?

feoh:
My thoughts are that blink-cmp looks interesting but that I do not understand Neovim completion well enough to undertake the transition.

Pull requests always welcome!

bluz71:
Blink ships a Rust compiled library whilst nvim-cmp is pure Lua. Pure Lua plugins will have lower maintenance overhead in the long run.

nvim-cmp used to have poor performance with Tailwind LSP, however a recent commit resolved that issue.

I compared completion performance between the two recently and could not see any obvious difference.

I suggest be conservative for now.

iSolate77:
Agree with sticking to nvim-cmp. While I've used blink for the past few weeks. As a kickstarter kit, it's better to have a more stable plugin like nvim-cmp, instead of the actively developed blink, which could introduce breaking changes.

sarmong:
Agree with sticking to nvim-cmp, but keeping an eye on blink and switching to it when the stable release is published. Mainly for the reason of it being simpler to use, because nvim-cmp may be tricky to configure even for advanced neovim users.

feoh:
Thanks for the feedback all! Let's collectively keep an eye on things and switch when it makes sense :)

I do like the idea that blink bundles a lot of things we currently have to explicitly configure in by default.

I think long term, shrinking kickstart's code size would be hugely beneficial in the mission of helping folks get started in an understandable way.

rfletchr:
cmp can display signatures, it needs them added as a source.

I'll make a PR

rfletchr:
@makew0rld #1358 

That'll get you going.

I think you should still explore blink. I personally prefer it to cmp.nvim. I do also think that the choice to use a pure lua solution is sound. Introducing compiled dependencies always adds complication. Rust is no exception.

feoh:
[...]

PR #1358, opened 2025-02-16
feat: add basic function signature help

rfletchr:
this change adds basic function signature help to kickstart.

Screenshot From 2025-02-16 17-43-01

this relates to the conversation in #1331

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -767,6 +767,8 @@ require('lazy').setup({
       --  into multiple repos for maintenance purposes.
       'hrsh7th/cmp-nvim-lsp',
       'hrsh7th/cmp-path',
+      'hrsh7th/cmp-path',
+      'htsh7th/cmp-nvim-lsp-signature-help',
     },
     config = function()
       -- See `:help cmp`
@@ -843,6 +845,7 @@ require('lazy').setup({
           { name = 'nvim_lsp' },
           { name = 'luasnip' },
           { name = 'path' },
+          { name = 'nvin_lsp_signature_help' },
         },
       }
     end,
--- a/init.lua
+++ b/init.lua
@@ -845,7 +845,7 @@ require('lazy').setup({
           { name = 'nvim_lsp' },
           { name = 'luasnip' },
           { name = 'path' },
-          { name = 'nvin_lsp_signature_help' },
+          { name = 'nvim_lsp_signature_help' },
         },
       }
     end,


init.lua
```diff
@@ -843,6 +845,7 @@ require('lazy').setup({
           { name = 'nvim_lsp' },
           { name = 'luasnip' },
           { name = 'path' },
+          { name = 'nvin_lsp_signature_help' },
```
makew0rld:
```suggestion
          { name = 'nvim_lsp_signature_help' },
```

arerossehaug:
Alternatively there could be a keybind like:
```lua
vim.keymap.set("i", "<C-h>", function()
    vim.lsp.buf.signature_help() 
    end, opts)

```
Rather than bringing in a new plugin and extending cmp even more when the lsp api can help here

rfletchr:
> Alternatively there could be a keybind like:
> 
> ```lua
> vim.keymap.set("i", "<C-h>", function()
>     vim.lsp.buf.signature_help() 
>     end, opts)
> ```
> 
> Rather than bringing in a new plugin and extending cmp even more when the lsp api can help here

this feels like a matter of taste. I don't really have an opinion either way. I just saw someone requesting this was added so I made the PR.

init.lua
rfletchr: > 

how the heck. well thats embarrassing. thanks for pointing it out

feoh:
I don't see a problem with this addition. It's just another nvim-cmp extension-let, and signature help feels like a great addition for new users.

That said, this is a community repo, so if others have strong feelings against I'll give you all a bit to speak up. Otherwise I'll merge this.

Thanks for pitching in everyone I *really* appreciate it!

rfletchr:
> I don't see a problem with this addition. It's just another nvim-cmp extension-let, and signature help feels like a great addition for new users.
> 
> That said, this is a community repo, so if others have strong feelings against I'll give you all a bit to speak up. Otherwise I'll merge this.
> 
> Thanks for pitching in everyone I _really_ appreciate it!

Sounds Good, coordinating this sort of thing can be allot of work, and having a guide for contributors on what is and isn't a suitable PR might really help the project.

I've worked on a few open projects and its good to have something to refer to, it helps with subjective choices. It helps to reduce bike-shedding and solves allot of subjective issues where it comes down to preferences.

PR #1360, opened 2025-02-17
perf: load tokyonight.nvim in the intended way

### LINKED ISSUE #1357: Tokyonight should be initialized as intended
## Incorrect Initialization of `tokyonight.nvim`

I have noticed that `tokyonight.nvim` is being initialized in a way that is not aligned with the recommended approach by [@folke](https://github.com/folke).

### Expected Initialization

[According to the official README](https://github.com/folke/tokyonight.nvim?tab=readme-ov-file#-installation), the proper way to install and initialize the plugin is:

```lua
{
  "folke/tokyonight.nvim",
  lazy = false,
  priority = 1000,
  opts = {},
}
```

This does **not** set the colorscheme; it only initializes the plugin. However, it is important to call `setup()` before configuring the theme. This is because `tokyonight.nvim` applies only the necessary highlight groups based on the installed plugins, which helps optimize startup time.

### Correct Way to Apply the Colorscheme

To properly apply the colorscheme, the configuration should look like this:

```lua
{
  "folke/tokyonight.nvim",
  lazy = false,
  priority = 1000,
  config = function()
    require("tokyonight").setup()
    vim.cmd.colorscheme("tokyonight")
  end
}
```

### Disabling Italics in Comments (as Kickstart Does)

If you want to disable italics in comments, similar to what Kickstart currently does, use:

```lua
{
  "folke/tokyonight.nvim",
  lazy = false,
  priority = 1000,
  config = function()
    require("tokyonight").setup({
      styles = {
        comments = { italic = false },
      },
    })
    vim.cmd.colorscheme("tokyonight")
  end
}
```

### Alternative Method (Less Documented in `lazy.nvim`)

Although not well-documented in `lazy.nvim`, you can also achieve the same effect with:

```lua
{
  "folke/tokyonight.nvim",
  priority = 1000,
  opts = {
    styles = {
      comments = { italic = false },
    },
  },
  config = function(_, opts)
    require("tokyonight").setup(opts)
    vim.cmd.colorscheme("tokyonight-night")
  end,
}
```

Would it be possible to align the initialization of `tokyonight.nvim` with the recommended approach?

warbacon:
This will cause `tokyonight.nvim` to use its cache and set only the necessary highlighting groups according to the installed plugins, and may improve startup time.

Fixes #1357.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -855,14 +855,18 @@ require('lazy').setup({
     -- If you want to see what colorschemes are already installed, you can use `:Telescope colorscheme`.
     'folke/tokyonight.nvim',
     priority = 1000, -- Make sure to load this before all the other start plugins.
-    init = function()
+    config = function()
+      ---@diagnostic disable-next-line: missing-fields
+      require('tokyonight').setup {
+        styles = {
+          comments = { italic = false }, -- Disable italics in comments
+        },
+      }
+
       -- Load the colorscheme here.
       -- Like many other themes, this one has different styles, and you could load
       -- any other, such as 'tokyonight-storm', 'tokyonight-moon', or 'tokyonight-day'.
       vim.cmd.colorscheme 'tokyonight-night'
-
-      -- You can configure highlights by doing something like:
-      vim.cmd.hi 'Comment gui=none'
     end,
   },
 


init.lua
```diff
@@ -855,14 +855,18 @@ require('lazy').setup({
...
+        },
+      }
+
       -- Load the colorscheme here.
       -- Like many other themes, this one has different styles, and you could load
       -- any other, such as 'tokyonight-storm', 'tokyonight-moon', or 'tokyonight-day'.
       vim.cmd.colorscheme 'tokyonight-night'
-
-      -- You can configure highlights by doing something like:
-      vim.cmd.hi 'Comment gui=none'
```
ro0gr, line 865: I'm unsure what it stood for, but why do you remove this?

warbacon: This prevented comments from being displayed in italics, as that's the default behavior in Tokyonight. I removed that line of code and instead used Tokyonight's built-in method (as shown in the `setup` function above).

feoh (approved):
This looks excellent. Thank you!

PR #1368, opened 2025-02-18
feat(keymap): move windows without `<C-w>`

sand4rt:
**NOTE**

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -189,6 +189,12 @@ vim.keymap.set('n', '<C-l>', '<C-w><C-l>', { desc = 'Move focus to the right win
 vim.keymap.set('n', '<C-j>', '<C-w><C-j>', { desc = 'Move focus to the lower window' })
 vim.keymap.set('n', '<C-k>', '<C-w><C-k>', { desc = 'Move focus to the upper window' })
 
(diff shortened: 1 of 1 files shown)


feoh:
This feels like a highly personalized customization to me. But if we get a ton of "HECK YES WE WANT THIS MERGE IT PLEASE!" I'll be delighted to do so.

sand4rt:
@feoh I added those keymaps because similar ones that allow changing window focus without using `<C-w>` are included in kickstart.nvim. It seems odd to me that window movement still requires `<C-w>` but window focus doesn't: https://github.com/nvim-lua/kickstart.nvim/blob/34e7d29aa7b6e95cf09d62baf4c9082db5b129c0/init.lua#L186-L190

Feel free to close the PR if it's not desired of course.

feoh:
Hey thanks for the explanation. I like the idea!

Anyone have any feelings about this? I'd just love a second "I want this" before I merge.

glaulher:
I liked it, it was more simplified.

VlaDexa:
I won't use these, but I like the idea of including these as a helping aid guiding people to learn about windows and tabs and stuff

feoh:
Thanks everyone for chiming in! Merging!

PR #1387, opened 2025-02-25
Updated colorscheme

ils-davec:
Updated init.lua to utilize the `catppuccin` color scheme instead of the default `tokyonight`.


ils-davec:
Validated and approved.

PR #1395, opened 2025-03-03
Possible unintended behavior in Conform's opts format_on_save 

RulentWave:
This pull request changes Conform's format_on_save lambda so that buffers that match disable_filetypes return nil instead of a table with the lsp_format  entry equaling "never". This allows you to enable a formatter for langages in the disable_filetypes table to have a formatter that can be run manually with Leader-f but doesn't enable format_on_save for them. Before this change, if you add a formatter for a language such as c, like clang-format, it enables format_on_save since the formatter is set explicitly. I think this probably isn't the intended behavior but in case it is, feel free to reject this PR.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -744,16 +744,14 @@ require('lazy').setup({
         -- have a well standardized coding style. You can add additional
         -- languages here or re-enable it for the disabled ones.
         local disable_filetypes = { c = true, cpp = true }
-        local lsp_format_opt
         if disable_filetypes[vim.bo[bufnr].filetype] then
-          lsp_format_opt = 'never'
+          return nil
         else
-          lsp_format_opt = 'fallback'
+          return {
+            timeout_ms = 500,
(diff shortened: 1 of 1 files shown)


feoh:
@RulentWave Thanks for the contribution!

I don't understand the surrounding code well enough to have feelings on this. Thanks for the thumbs up @jiixyj.

Does anyone else have feelings on this yay or nay?

I'll give it some more time for the folks who usually are kind enough to come back and correct my mistakes after the fact to chime in :)

feoh:
OK. 

It's been 2 weeks.

I'm sure if I merge this now a bunch of helpful people who only check in here once in a while will tell me what a buffoon I am and file a corrective fix (Thank you people! Happy to be the buffoon that gets the help to fix :).

So I'll merge this in a few days, possibly this weekend. Thanks for hanging tough!

PR #1414, opened 2025-03-20
fix: use correct github abmonition syntax

saccarosium:


feoh:
Looks like this is based around the syntax laid out  [here](https://github.com/orgs/community/discussions/16925).

Dig it. Merging!

PR #1427, opened 2025-03-29
Change LSP Keybindings to Match the Default `gr` Bindings Introduced in Neovim 0.11

theopn:
```
                                          *grr* *gra* *grn* *gri* *i_CTRL-S*
Some keymaps are created unconditionally when Nvim starts:
- "grn" is mapped in Normal mode to |vim.lsp.buf.rename()|
- "gra" is mapped in Normal and Visual mode to |vim.lsp.buf.code_action()|
- "grr" is mapped in Normal mode to |vim.lsp.buf.references()|
- "gri" is mapped in Normal mode to |vim.lsp.buf.implementation()|
- "gO" is mapped in Normal mode to |vim.lsp.buf.document_symbol()|
- CTRL-S is mapped in Insert mode to |vim.lsp.buf.signature_help()|
```

Neovim 0.11 introduced many default LSP keybindings prefixed with `gr`, and this PR changes the Telescope keybindings defined in `LspAttach` autocmds to match the default ones.

I think changing other keybindings (`<leader>D` for type definition and `<leader>ws` for workspace symbols) to have `g` prefix should also be considered.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -337,7 +337,6 @@ require('lazy').setup({
       -- Document existing key chains
       spec = {
         { '<leader>s', group = '[S]earch' },
-        { '<leader>w', group = '[W]orkspace' },
         { '<leader>t', group = '[T]oggle' },
         { '<leader>h', group = 'Git [H]unk', mode = { 'n', 'v' } },
       },
@@ -535,7 +534,7 @@ require('lazy').setup({
 
           -- Execute a code action, usually your cursor needs to be on top of an error
           -- or a suggestion from your LSP for this to activate.
-          map('gra', vim.lsp.buf.code_action, 'Code [A]ction', { 'n', 'x' })
+          map('gra', vim.lsp.buf.code_action, '[G]oto Code [A]ction', { 'n', 'x' })
 
           -- Find references for the word under your cursor.
           map('grr', require('telescope.builtin').lsp_references, '[G]oto [R]eferences')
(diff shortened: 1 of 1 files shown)


arerossehaug:
Any reason to choose gca over gra for code actions?
While the [G]oto [C]ode [A]ction makes more sense than [G]oto [R]? [A]ction, the gra is a smoother key combination in my opinion

Semi related, but should it be considered that gd (goto definition) should become grd?

theopn:
@arerossehaug 😅 just a typo, I amended it to be `gra`. My idea was to follow the default keybindings provided by Neovim to minimize the confusion.

I have also added a commit for changing the other keybindings, prefixing `gr` for "Goto" and `g` for opening a list of symbols:

- `gd` -> `grd` for goto definition
- `gD` -> `grD` for goto declaration
- `<leader>D` -> `grt` for type definition
- `<leader>w` -> `gW` for workspace symbols

feoh:
I have no feelings about this, but making changes to key bindings always ends up with a million angry WHY DID YOU MOVE MY CHEESE? comments.

Thoughts from everyone?

theopn:
@feoh I agree that changing keybindings can cause confusion, but I have two arguments for this particular changes:

1. some existing keybindings (e.g., `gr` for LSP references) conflicts with the new default keybindings in 0.11, so we need to either delete the default keybindings or change our keybindings; see #1436
2. these are the default keybindings that Neovim will use moving forward, and considering the demographics of people using Kickstart (lots of beginners and new to configuring Neovim) and how they use Kickstart (use to build their own config), I think it is important to let them know that default keybindings for such and such actions exist.

Thoughts?

chrstief:
If the kickstart project was started today, they would go with the default nvim keybinds. I’m in support of this change, instead of sticking to non default keybinds for historic reasons.

feoh:
Two people voting yay is enough for me  :) Thanks all! merging.

5brian:
Sorry to necro this, would a pr to add the few leader key bindings back in comments be accepted? 

For example:

```

-- If you prefer the legacy <leader> keymaps,
-- you can uncomment the following lines

```

PR #1426, opened 2025-03-28
feat: switch nvim-cmp for blink.cmp

saghen:
Opening as draft since I have yet to verify everything works as expected

Closes #1331

PR #1433, opened 2025-03-31
feat: native LSP completion

cmdrrobin:
This is my attempt to replace nvim-cmp with the `vim.lsp.completion` that came with Neovim 0.11 release.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -161,6 +161,10 @@ vim.opt.scrolloff = 10
 -- See `:help 'confirm'`
 vim.opt.confirm = true
 
+-- set options for insert completion (ins-completion)
+-- See `:help completeopt`
+vim.o.completeopt = 'noselect,menu,menuone,noinsert,popup'
+
 -- [[ Basic Keymaps ]]
 --  See `:help vim.keymap.set()`
 
@@ -486,8 +490,8 @@ require('lazy').setup({
       -- Useful status updates for LSP.
       { 'j-hui/fidget.nvim', opts = {} },
 
-      -- Allows extra capabilities provided by nvim-cmp
-      -- 'hrsh7th/cmp-nvim-lsp',
+      -- Show pictograms for completion
+      { 'onsails/lspkind.nvim', enabled = vim.g.have_nerd_font },
     },
     config = function()
       -- Brief aside: **What is LSP?**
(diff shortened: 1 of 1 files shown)


rfletchr:
you beat me to it!

I added this little hacky function to trigger on `InsertCharPre` to enable automatic signature help. 

I was going to change it to be setup on buffer attachment.

```lua
local function trigger_signature_if_applicable()
	local params = vim.lsp.util.make_position_params()
	for _, client in pairs(vim.lsp.get_clients()) do
		if client.server_capabilities.signatureHelpProvider then
			client.request("textDocument/signatureHelp", params, function(err, result, ctx)
				if not err and result and result.signatures and #result.signatures > 0 then
					vim.lsp.buf.signature_help()
				end
			end, 0)
			break
		end
	end
end

```

image

feoh:
Hi! I tested this and it seems to work great.

Thoughts:
- I love how FAST it is! Definitely even faster than blink much though <3 that plugin so much :)
- You only get LSP completions - is there a mechanism to add completions from different providers like Copilot?

(Guessing most 'go native/default' folks will loathe the idea of Copilot or other AI assistants but RealTalk many of us use them as a way of life nowadays, for better or worse :)

Thanks so much for this. It's really great to have a good example of both styles so we can figure out what works best for our newbie Neovimmers :)

cmdrrobin:
For the record, I'm not the native guy, I just try to contribute to the project. ;)

As for your question, if it is possible to use Copilot with native completion? I really don't know. I don't use Copilot nor know if this capable. 

The native completion is part of `vim.lsp`, so i don't know if we could integrate other plugins with native completion. 

As for finalizing the PR, I have quite busy lately, and yet not had the time to dive into it.

rfletchr:
> Hi! I tested this and it seems to work great.
> 
> Thoughts:
> 
> * I love how FAST it is! Definitely even faster than blink much though <3 that plugin so much :)
> * You only get LSP completions - is there a mechanism to add completions from different providers like Copilot?
> 
> (Guessing most 'go native/default' folks will loathe the idea of Copilot or other AI assistants but RealTalk many of us use them as a way of life nowadays, for better or worse :)
> 
> Thanks so much for this. It's really great to have a good example of both styles so we can figure out what works best for our newbie Neovimmers :)

I believe its using the old `help: completion-functions` system. If that's the case you can probably grab that function and wrap it with what ever you want. (citation needed ;))

I'm a 'native/default' enjoyer and I don't really care what you use so long as it works for you, it's definitely something people expect and is a worth considering. [MariaSolOs](https://github.com/MariaSolOs) could answer this properly as she wrote the system

feoh:
OK.

@cmdrrobin Thanks for contributing the draft, and I *totally* get it, none of us are doing this full time. Good luck with your other work!

So, given that I'm tempted to merge the blink PR since it's not a draft and folks (like me) can support it more fully.

That said I'm not doing any of this right this instant so plenty of time for others to weigh in and offer support for the native stuff.

cmdrrobin:
@feoh no worries. I think for the ease and use of it, it would be better to use blink.cmp for now. You can merge #1426 and close this MR.

I will keep the branch available (as reference), for those who wants to test or view my changes.

miguelsrmv:
Hey guys. Sorry to weigh in, but I've been following the discussion closely and would like to give my 2 cents.

I get the appeal of using built-in auto-completion ; on the other hand, blink feels like a much more flexible piece of software as it provides non-LSP completion. 

Kickstart has always been relatively lean in its plugin gallery. However, I've always seen its goal as "let's learn how all of this nvim wizardry works", rather than relying on the defaults for the sake of relying on the defaults (hence why it used lspconfig or neo-tree, for example).

My vote is for blink. However, I think there's also room for built-in LSP with blink as an optional plugin (just like some others plugins that are, by default, commented out).

Keep up the good work!

feoh:
OK. That's several people who might help me assuage the pitchfork wielding mob if people end up unhappy, I'm gonna go merge the blink PR because I think that's the right answer too.

feoh:
@cmdrrobin thanks again for your work on this! It helped me understand how the default worked a bit better, I really appreciate it!

Gonna close this now though.

rfletchr:
> OK. That's several people who might help me assuage the pitchfork wielding mob if people end up unhappy, I'm gonna go merge the blink PR because I think that's the right answer too.

no pitchforks here, @miguelsrmv nailed it.

Issue #1376, opened 2025-02-21
May I add comments about "lua is a programming language" and "lua is embedded-> the global vim exists" to init.lua?


Weyaaron:
Dear all,

I would like to contribute something, but I would like to know if this is "accepted". I would rather ask first then write something with care and not use it. This is inspired by my own experience, I used kickstart back in the day and it helped a lot. There are two concepts I that I feel like are missing and could speed up learning even more. 

"Lua is an actual programming language": This is mentioned, but it is buried deep in the code and its impact is not discussed properly. I feel like a small dedicated paragraph/section with some really easy examples could go a long way to improve the visibility of this fact. 

"Lua is embedded->Existence of the global vim ':  I feel like the existence of the global vim is not motivated/explained at all. It just exists, but it is not mentioned how or why and what the implications are. (There are no mentions of "global" or "embedded" anywhere in the document). While this is a somewhat technical topic, I feel explaining it could improve understanding of the setup quite a lot. 

Let me know what you think, based on the feedback I will start writing some first drafts soon. I am open to including either but not the other.

Crypto-Spartan:
I think this could be helpful, as long as the explanations are short and to-the-point. Would be worthwhile to link the help docs that explain it in further detail as well.

feoh:
@Crypto-Spartan Can you please go take a look at that PR?

I can't decide whether this would help and inspire new users or just confuse the heck out of them.

feoh:
Hiya! If you make the changes I suggested or suggest alternatives I'll merge it.

eamonburns:
Seeing as the linked PR is closed, should this issue be closed as well?

Weyaaron:
Good point, has been done :)

PR #1446, opened 2025-04-06
docs: Adds a paragraph about lua being a real programming language.

Weyaaron:
As proposed in #1376, I have written something up. I am open to discussion about it!

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -113,6 +113,7 @@ P.S. You can delete this when you're done too. It's your config now! :)
 -- if vim.fn.expand("%") == "file.txt" then
 -- 	vim.cmd("so ./additional_config.lua")
 -- end
+--
 -- One use-case could be a journal where the script.lua automatically moves your cursor
 -- to the end of file and inserts the curent date when you open it. (This suggestion is certainly not
 -- the easiest example, but that it is possible at all should get your mind racing
--- a/init.lua
+++ b/init.lua
@@ -109,6 +109,7 @@ P.S. You can delete this when you're done too. It's your config now! :)
 -- on valid effidence that you like this feature.
 --
 -- The following snippet loads/executes some lua file when a specific file is opened:
+--
 -- if vim.fn.expand("%") == "file.txt" then
 -- 	vim.cmd("so ./additional_config.lua")
 -- end
--- a/init.lua
+++ b/init.lua
@@ -90,6 +90,7 @@ P.S. You can delete this when you're done too. It's your config now! :)
 -- like variables, conditionals, loops and functions. Secondly, that you can interact with
 -- neovim in a concise way. The following lines contain two examples that might give you some inspiration.
 -- This snippet adds a keymap to open a predefined file in a new tab:
+--
 -- local function open_notes()
 --   vim.cmd 'tabnew'
 --   vim.cmd 'tabnext'
--- a/init.lua
+++ b/init.lua
@@ -88,7 +88,7 @@ P.S. You can delete this when you're done too. It's your config now! :)
 -- One last note before we leave you to explore: Lua is a real programming language.
 -- Why is that important? First of all, that you can use all the features of Lua,
 -- like variables, conditionals, loops and functions. Secondly, that you can interact with
--- neovim in concise way. The following lines contain two examples that might give you some inspirations.
+-- neovim in a concise way. The following lines contain two examples that might give you some inspiration.
 -- This snippet adds a keymap to open a predefined file in a new tab:
 -- local function open_notes()
 --   vim.cmd 'tabnew'
--- a/init.lua
+++ b/init.lua
@@ -83,6 +83,39 @@ I hope you enjoy your Neovim journey,
 
 P.S. You can delete this when you're done too. It's your config now! :)
 --]]
+--
+--
+-- One last note before we leave you to play around: Lua is a real programming language.
+-- Why is that important? First of all, that you can use all the features of Lua,
+-- like variables, conditionals, loops and functions. Secondly, that you can interact with
+-- neovim in concise way. The following lines contain two examples that might give you some inspirations.
+-- This snippet adds a keymap to open a predefined file in a new tab:
+-- local function open_notes()
+--   vim.cmd 'tabnew'
+--   vim.cmd 'tabnext'
+--   local command = 'e' .. '/full_path/to_file' -- '..' is the build in string concatenation
+--   vim.cmd(command)
+-- end
+-- vim.keymap.set('n', '<leader>m', open_notes)
+--
+-- This is interesting because there are certainly plugins out there
+-- that provide this functionallity in a sophisticated manner.
+-- But a plugin might change, might not provide the options you like,
+-- or you might not use this feature at all. If you do not use it,
+-- simply delete this code and you are done. If you like it, you
+-- may expand to what you need and nothing else. And if this is to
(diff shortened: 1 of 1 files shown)


feoh:
I like the idea of this but I think the text needs a bit of work.

There are some plurals where they should be singular etc.

Also, will this inspire or merely confuse new Neovim users?

Anyway thank you for your contribution!

init.lua
```diff
@@ -83,6 +83,39 @@ I hope you enjoy your Neovim journey,
 
 P.S. You can delete this when you're done too. It's your config now! :)
 --]]
+--
+--
+-- One last note before we leave you to play around: Lua is a real programming language.
```
feoh:
```suggestion
-- One last note before we leave you to explore: Lua is a real programming language.
```

init.lua
```diff
@@ -83,6 +83,39 @@ I hope you enjoy your Neovim journey,
...
+-- Why is that important? First of all, that you can use all the features of Lua,
+-- like variables, conditionals, loops and functions. Secondly, that you can interact with
+-- neovim in concise way. The following lines contain two examples that might give you some inspirations.
```
feoh:
```suggestion
-- neovim in a concise way. The following lines contain two examples that might give you some inspiration.
```

init.lua
```diff
@@ -83,6 +83,39 @@ I hope you enjoy your Neovim journey,
...
+-- This snippet adds a keymap to open a predefined file in a new tab:
```
feoh, line 92:
```suggestion
-- This snippet adds a keymap to open a predefined file in a new tab:
--
```

init.lua
```diff
@@ -83,6 +83,39 @@ I hope you enjoy your Neovim journey,
...
+-- This is interesting because there are certainly plugins out there
+-- that provide this functionallity in a sophisticated manner.
+-- But a plugin might change, might not provide the options you like,
+-- or you might not use this feature at all. If you do not use it,
+-- simply delete this code and you are done. If you like it, you
+-- may expand to what you need and nothing else. And if this is to
+-- much work, you might end up using the plugin anyway, but based
+-- on valid effidence that you like this feature.
+--
+-- The following snippet loads/executes some lua file when a specific file is opened:
```
feoh, line 111:
```suggestion
-- The following snippet loads/executes some lua file when a specific file is opened:
--
```

init.lua
```diff
@@ -83,6 +83,39 @@ I hope you enjoy your Neovim journey,
...
+-- if vim.fn.expand("%") == "file.txt" then
+-- 	vim.cmd("so ./additional_config.lua")
+-- end
```
feoh, line 115:
```suggestion
-- end
--
```

init.lua
```diff
@@ -83,6 +83,39 @@ I hope you enjoy your Neovim journey,
...
+-- One use-case could be a journal where the script.lua automatically moves your cursor
+-- to the end of file and inserts the curent date when you open it. (This suggestion is certainly not
+-- the easiest example, but that it is possible at all should get your mind racing
+-- with possibilities.)
```
feoh, line 120:
```suggestion
-- with possibilities.)
--
```

feoh:
If you accept my changes or suggest alternatives I'll merge this. Thanks for the contribution!

feoh (changes requested):

Weyaaron:
I appreciate the effort, thanks for providing the feedback. You should probably rebase the branch to get rid of all the additional commits. If you do not do this I'll gladly do it.

Weyaaron:
After the rebase it can be merged :+1:

init.lua
```diff
@@ -83,6 +83,43 @@ I hope you enjoy your Neovim journey,
...
+-- vim.keymap.set('n', '<leader>m', open_notes)
+--
+-- This is interesting because there are certainly plugins out there
+-- that provide this functionallity in a sophisticated manner.
+-- But a plugin might change, might not provide the options you like,
+-- or you might not use this feature at all. If you do not use it,
+-- simply delete this code and you are done. If you like it, you
+-- may expand to what you need and nothing else. And if this is to
+-- much work, you might end up using the plugin anyway, but based
+-- on valid effidence that you like this feature.
```
ro0gr, line 109: > effidence

is this a normative word? As a non-native speaker, I suspect it's not a standard English word.

ro0gr (commented):
Honestly I find it confusing. I'm not quite sure what the purpose of including this text here is.

If I want to learn what Lua is, I'd search for it and I have the following search result on the top: 

> [The Programming Language Lua](https://www.lua.org/) 

Additionally, after going through the examples, I believe the kickstart codebase is full of similar snippets.

Wouldn’t this be considered unnecessary noise in the documentation?

init.lua
Weyaaron: Excactly, you spotted a typo. I should have been more thorough on these.

Weyaaron:
> Honestly I find it confusing. I'm not quite sure what the purpose of including this text here is.
> 
> If I want to learn what Lua is, I'd search for it and I have the following search result on the top:
> 
> > [The Programming Language Lua](https://www.lua.org/)
> 
> Additionally, after going through the examples, I believe the kickstart codebase is full of similar snippets.
> 
> Wouldn’t this be considered unnecessary noise in the documentation?

It is not so much about what lua is, but rather about communication that you can use lua with its full power. This is "somewhat obvious", but I feel like it is not written down sufficiently. Especially considering the script  uses some of these powers later on. And I feel like providing this context will ease understanding later on.

feoh:
@ro0gr Is there a change we could make that would make this more concise or impactful or do you think the whole idea needs to be rejected?

Thanks for participating everyone. I have no skin in this game, I already know Lua and am not a newbie so all opinions are helpful :)

ro0gr:
@feoh That's a great question.

My main concern with this PR is the inclusion of random code snippets, which, on their own, aren't sufficient for a curious user to find the right answers. As I mentioned earlier, I believe the Kickstart project already provides a solid set of working examples for newcomers to explore.

Additionally, we're not leaving users alone with just raw Lua - we already offer some starting points, for example:

https://github.com/nvim-lua/kickstart.nvim/blob/d350db2449da40df003c40d440f909d74e2d4e70/init.lua#L35-L42

In my view, the best approach is to outsource deeper learning to external resources, rather than trying to inline educational content directly into comments. Unfortunately, I don't have a good list of links off the top of my head, maybe someone does? 

However, I find the [r/neovim](https://www.reddit.com/r/neovim/) community very helpful for this purpose. It might make sense to at least share a link to this and similar resources.

danarnold:
There's also already a reference to Lua being a real programming language:

https://github.com/nvim-lua/kickstart.nvim/blob/d350db2449da40df003c40d440f909d74e2d4e70/init.lua#L520-L521

feoh:
@Weyaaron What do you think of the already existing references others have pointed out?

dam9000:
In my view this PR is superfluous. There's already enough helpful hints throughout the whole kickstart. People that will be interested in learning more about lua will be able to find numerous examples and documentation about it, no need to fit everything in kickstart, this can be a never ending story, there's always more that can be added.

Weyaaron:
> @Weyaaron What do you think of the already existing references others have pointed out?

Sorry for the late reply, I appreciate the feedback/dialog. Given that multiple people have voiced concern I concede my point and retract the pr.

PR #1448, opened 2025-04-06
fix(Telescope): remove Telescope `0.1` branch lock

vaporif:
Telescope hasn't had a release in over a year. Sticking to old version will omit lots of fixes.
For instance dynamic workspace symbols won't work and will throw an error in latest nvim version.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -357,7 +357,6 @@ require('lazy').setup({
   { -- Fuzzy Finder (files, lsp, etc)
     'nvim-telescope/telescope.nvim',
     event = 'VimEnter',
-    branch = '0.1.x',
     dependencies = {
       'nvim-lua/plenary.nvim',
       { -- If encountering errors, see telescope-fzf-native README for installation instructions


rfletchr:
hi @vaporif what issue does this fix? It's pretty normal to lock plugins to a range, telescope is still in v.0.x so your change looks like its simply opting into master.

edit: ah branch lock, not tag lock. ignore me im Derping.

vaporif:
Screenshot 2025-04-07 at 22 00 27
This is for instance dynamic workspace symbols search - throws error and doesn't display anything

feoh:
I use Telescope this way too in my persona config.

Thoughts from anyone? This is a moment where I wish @tjdevries would chime in once in a while but I know he's off in the stratosphere building ssh coffee startups and coding in OCaml for his livestream :)

feoh:
To be clear I think I'mma merge this unless people start screaming in the next day or so :)

feoh:
There is actual breakage here unless I merge this so I'm merging it.

Issue #1458, opened 2025-04-10
[Suggestion] Using native LSP functionality with 0.11


baldvaritesh:
Can the current functionality for LSP in kickstarter be moved to the 0.11 native lsp APIs; would be a good starter for a base configuration which doesn't rely on `nvim-lspconfig` going forward

feoh:
Hoo boy.

Please go read https://github.com/nvim-lua/kickstart.nvim/issues/1331

If:

- Large numbers of people in the community want native LSP autocomplete *AND*
- We have someone willing to both draft a pull request AND support it on an ongoing basis *AND*
- We think the native autocomplete will actually provide a good experience for new Neovim users

I'll happily merge it. Until those conditions are met it stays with blink because I and a bunch of others think that's an excellent solution and we have the author, @Saghen here supporting their work.

Closing for now. Feel free to re-open if all the above are met.

PR #1475, opened 2025-04-16
feat: add support for new LSP config API in Neovim 0.11+

umutondersu:
## Summary
- Added support for the new LSP configuration API introduced in Neovim 0.11+
- Updated Mason version to v2

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -705,14 +705,19 @@ require('lazy').setup({
       })
       require('mason-tool-installer').setup { ensure_installed = ensure_installed }
 
-      -- Handle LSP setups
-      -- See :help vim.lsp.enable
-      for server, config in pairs(servers) do
-        vim.lsp.config(server, config)
-        vim.lsp.enable(server)
-      end
+      require('mason-lspconfig').setup {
+        ensure_installed = {}, -- explicitly set to an empty table (Kickstart populates installs via mason-tool-installer)
+        automatic_installation = false,
+        handlers = {
+          function(server_name)
+            local config = servers[server_name] or {}
+            vim.lsp.config(server_name, config)
+            vim.lsp.enable(server_name)
+          end,
+        },
+      }
       -- NOTE: Some servers still require the nvim-lspconfig setup until they are updated
-      -- Use this template inside the for loop if you encounter issues with an lsp
+      -- Add this template inside the handler function after initializing config if you encounter issues with any lsp
       --
       -- if server == 'example_server' or server == 'example_server2' then
       --   -- This handles overriding only values explicitly passed
@@ -721,12 +726,10 @@ require('lazy').setup({
       --   local capabilities = require('blink.cmp').get_lsp_capabilities()
       --   config.capabilities = vim.tbl_deep_extend('force', {}, capabilities, config.capabilities or {})
       --   require('mason-lspconfig')[server].setup(config)
-      -- else
-      --   vim.lsp.config(server, config)
-      --   vim.lsp.enable(server)
+      --   return
       -- end
       --
-      -- LSP servers and clients are able to communicate to each other what features they support.
+      --  LSP servers and clients are able to communicate to each other what features they support.
       --  With nvim-lspconfig setup, Neovim doesn't support everything that is in the LSP specification.
       --  When you add blink.cmp, luasnip, etc. Neovim now has *more* capabilities.
       --  So, we create new capabilities with blink.cmp, and then broadcast that to the servers.
--- a/init.lua
+++ b/init.lua
@@ -711,20 +711,34 @@ require('lazy').setup({
       })
       require('mason-tool-installer').setup { ensure_installed = ensure_installed }
 
-      require('mason-lspconfig').setup {
-        ensure_installed = {}, -- explicitly set to an empty table (Kickstart populates installs via mason-tool-installer)
-        automatic_installation = false,
-        handlers = {
-          function(server_name)
-            local server = servers[server_name] or {}
-            -- This handles overriding only values explicitly passed
-            -- by the server configuration above. Useful when disabling
-            -- certain features of an LSP (for example, turning off formatting for ts_ls)
-            server.capabilities = vim.tbl_deep_extend('force', {}, capabilities, server.capabilities or {})
-            require('lspconfig')[server_name].setup(server)
-          end,
-        },
-      }
+      -- Handle lsp setups differently based on Neovim version
+      -- See :help vim.lsp.config and :help vim.lsp.enable for Neovim 0.11+
+      if vim.fn.has 'nvim-0.11' == 1 then
+        for server, config in pairs(servers) do
+          -- This handles overriding only values explicitly passed
+          -- by the server configuration above. Useful when disabling
+          -- certain features of an LSP (for example, turning off formatting for ts_ls)
+          config.capabilities = vim.tbl_deep_extend('force', {}, capabilities, server.capabilities or {})
+          vim.lsp.config(name, config)
+          vim.lsp.enable(name)
+        end
+      else
+        -- For Neovim 0.10 and below, use mason-lspconfig for setup
+        require('mason-lspconfig').setup {
+          ensure_installed = {}, -- explicitly set to an empty table (Kickstart populates installs via mason-tool-installer)
+          automatic_installation = false,
+          handlers = {
+            function(server_name)
+              local server = servers[server_name] or {}
+              -- This handles overriding only values explicitly passed
+              -- by the server configuration above. Useful when disabling
+              -- certain features of an LSP (for example, turning off formatting for ts_ls)
+              server.capabilities = vim.tbl_deep_extend('force', {}, capabilities, server.capabilities or {})
+              require('lspconfig')[server_name].setup(server)
+            end,
+          },
+        }
+      end
     end,
   },
 
--- a/init.lua
+++ b/init.lua
@@ -720,7 +720,7 @@ require('lazy').setup({
       --   -- certain features of an LSP (for example, turning off formatting for ts_ls)
       --   local capabilities = require('blink.cmp').get_lsp_capabilities()
       --   config.capabilities = vim.tbl_deep_extend('force', {}, capabilities, config.capabilities or {})
-      --   require('mason-lspconfig')[server_name].setup(config)
+      --   require('mason-lspconfig')[server].setup(config)
       -- else
       --   vim.lsp.config(server, config)
       --   vim.lsp.enable(server)
--- a/init.lua
+++ b/init.lua
@@ -482,7 +482,7 @@ require('lazy').setup({
       -- Automatically install LSPs and related tools to stdpath for Neovim
       -- Mason must be loaded before its dependents so we need to set it up here.
       -- NOTE: `opts = {}` is the same as calling `require('mason').setup({})`
-      { 'mason-orgmason.nvim', opts = {} },
+      { 'mason-org/mason.nvim', opts = {} },
       { 'mason-org/mason-lspconfig.nvim', opts = {} },
       'WhoIsSethDaniel/mason-tool-installer.nvim',
 
--- a/init.lua
+++ b/init.lua
@@ -474,8 +474,8 @@ require('lazy').setup({
       -- Automatically install LSPs and related tools to stdpath for Neovim
       -- Mason must be loaded before its dependents so we need to set it up here.
       -- NOTE: `opts = {}` is the same as calling `require('mason').setup({})`
-      { 'williamboman/mason.nvim', version = 'v1.*', opts = {} },
-      { 'williamboman/mason-lspconfig.nvim', version = 'v1.*' },
+      { 'williamboman/mason.nvim', opts = {} },
+      { 'williamboman/mason-lspconfig.nvim', opts = {} },
       'WhoIsSethDaniel/mason-tool-installer.nvim',
 
       -- Useful status updates for LSP.
@@ -705,34 +705,14 @@ require('lazy').setup({
       })
       require('mason-tool-installer').setup { ensure_installed = ensure_installed }
 
-      require('mason-lspconfig').setup {
-        ensure_installed = {}, -- explicitly set to an empty table (Kickstart populates installs via mason-tool-installer)
-        automatic_installation = false,
-        handlers = {
-          function(server_name)
-            local config = servers[server_name] or {}
-            vim.lsp.config(server_name, config)
-            vim.lsp.enable(server_name)
-          end,
-        },
-      }
-      -- NOTE: Some servers still require the nvim-lspconfig setup until they are updated
-      -- Add this template inside the handler function after initializing config if you encounter issues with any lsp
-      --
-      -- if server_name == 'example_server' or server_name == 'example_server2' then
-      --   -- This handles overriding only values explicitly passed
-      --   -- by the server configuration above. Useful when disabling
-      --   -- certain features of an LSP (for example, turning off formatting for ts_ls)
-      --   local capabilities = require('blink.cmp').get_lsp_capabilities()
-      --   config.capabilities = vim.tbl_deep_extend('force', {}, capabilities, config.capabilities or {})
-      --   require('mason-lspconfig')[server_name].setup(config)
-      --   return
-      -- end
-      --
-      --  LSP servers and clients are able to communicate to each other what features they support.
-      --  With nvim-lspconfig setup, Neovim doesn't support everything that is in the LSP specification.
-      --  When you add blink.cmp, luasnip, etc. Neovim now has *more* capabilities.
-      --  So, we create new capabilities with blink.cmp, and then broadcast that to the servers.
+      -- Installed LSPs are configured and enabled automatically with mason-lspconfig
+      -- The loop below is for overriding the default configuration of LSPs with the ones in the servers table
+      for server_name, config in pairs(servers) do
+        vim.lsp.config(server_name, config)
+      end
+
+      -- NOTE: Some servers may require an old setup until they are updated. For the full list refer here: https://github.com/neovim/nvim-lspconfig/issues/3705
+      -- These servers will have to be manually set up with require("lspconfig").server_name.setup{}
     end,
   },
 
--- a/init.lua
+++ b/init.lua
@@ -290,7 +290,7 @@ require('lazy').setup({
   -- Then, because we use the `opts` key (recommended), the configuration runs
   -- after the plugin has been loaded as `require(MODULE).setup(opts)`.
 
-  { -- Useful plugin to show you pending keybinds.
+  {                     -- Useful plugin to show you pending keybinds.
     'folke/which-key.nvim',
     event = 'VimEnter', -- Sets the loading event to 'VimEnter'
     opts = {
@@ -371,7 +371,7 @@ require('lazy').setup({
       { 'nvim-telescope/telescope-ui-select.nvim' },
 
       -- Useful for getting pretty icons, but requires a Nerd Font.
-      { 'nvim-tree/nvim-web-devicons', enabled = vim.g.have_nerd_font },
+      { 'nvim-tree/nvim-web-devicons',            enabled = vim.g.have_nerd_font },
     },
     config = function()
       -- Telescope is a fuzzy finder that comes with a lot of different things that
@@ -479,7 +479,7 @@ require('lazy').setup({
       'WhoIsSethDaniel/mason-tool-installer.nvim',
 
       -- Useful status updates for LSP.
-      { 'j-hui/fidget.nvim', opts = {} },
+      { 'j-hui/fidget.nvim',       opts = {} },
 
       -- Allows extra capabilities provided by blink.cmp
       'saghen/blink.cmp',
@@ -719,8 +719,8 @@ require('lazy').setup({
           -- by the server configuration above. Useful when disabling
           -- certain features of an LSP (for example, turning off formatting for ts_ls)
           config.capabilities = vim.tbl_deep_extend('force', {}, capabilities, config.capabilities or {})
-          vim.lsp.config(name, config)
-          vim.lsp.enable(name)
+          vim.lsp.config(server, config)
+          vim.lsp.enable(server)
         end
       else
         -- For Neovim 0.10 and below, use mason-lspconfig for setup
--- a/init.lua
+++ b/init.lua
@@ -718,7 +718,7 @@ require('lazy').setup({
           -- This handles overriding only values explicitly passed
           -- by the server configuration above. Useful when disabling
           -- certain features of an LSP (for example, turning off formatting for ts_ls)
-          config.capabilities = vim.tbl_deep_extend('force', {}, capabilities, server.capabilities or {})
+          config.capabilities = vim.tbl_deep_extend('force', {}, capabilities, config.capabilities or {})
           vim.lsp.config(name, config)
           vim.lsp.enable(name)
         end
--- a/init.lua
+++ b/init.lua
@@ -647,12 +647,6 @@ require('lazy').setup({
         },
       }
 
-      -- LSP servers and clients are able to communicate to each other what features they support.
-      --  By default, Neovim doesn't support everything that is in the LSP specification.
-      --  When you add blink.cmp, luasnip, etc. Neovim now has *more* capabilities.
-      --  So, we create new capabilities with blink.cmp, and then broadcast that to the servers.
-      local capabilities = require('blink.cmp').get_lsp_capabilities()
-
       -- Enable the following language servers
       --  Feel free to add/remove any LSPs that you want here. They will automatically be installed.
       --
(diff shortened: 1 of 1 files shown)


ro0gr:
> Kickstart.nvim targets only the latest ['stable'](https://github.com/neovim/neovim/releases/tag/stable) and latest ['nightly'](https://github.com/neovim/neovim/releases/tag/nightly) of Neovim.

Do we really need backward compatibility with Neovim 0 10? I'd consider going with a lighter, happy-path solution. Since the project mainly targets newcomers and servers educational purposes, they'd likely prefer to avoid dealing with legacy approaches.

umutondersu:
I think that would be the right approach as well. The reason I intended to maintain backwards compatibility is that some parts of the code is still trying to maintain 0.10 like the `client_supports_method`. Depending on the others reaction I would also be in favor of removing it.

umutondersu:
There is also the problem new syntax not supporting all lsp configs. For example tailwindcss requires the old setup. Could be a good idea to delay this until most lsp configs migrate.

nickgnd:
Hey everyone 👋 
I remember a few related conversations that happen lately on this topic, let me share them here

https://github.com/nvim-lua/kickstart.nvim/issues/1458
https://github.com/nvim-lua/kickstart.nvim/issues/1331

I hope they can be useful, cheers ✌️

init.lua
```diff
@@ -711,20 +711,34 @@ require('lazy').setup({
...
+      -- Handle lsp setups differently based on Neovim version
+      -- See :help vim.lsp.config and :help vim.lsp.enable for Neovim 0.11+
+      if vim.fn.has 'nvim-0.11' == 1 then
+        for server, config in pairs(servers) do
+          -- This handles overriding only values explicitly passed
+          -- by the server configuration above. Useful when disabling
+          -- certain features of an LSP (for example, turning off formatting for ts_ls)
+          config.capabilities = vim.tbl_deep_extend('force', {}, capabilities, config.capabilities or {})
+          vim.lsp.config(name, config)
+          vim.lsp.enable(name)
```
yetyetanotherusername:
```suggestion
          vim.lsp.config(name, config)
          vim.lsp.enable(name)
```
`name` isn't defined here, should this say `server`?

init.lua
```diff
@@ -711,20 +711,34 @@ require('lazy').setup({
...
+        end
```
yetyetanotherusername: I tried this in neovim 0.11 but it never registers any of the lsps installed via the mason gui, only the ones that are hardcoded in `init.lua`

umutondersu: That was intended. I think every lsp should be declared explicitly inside init.lua since in the new lsp config they also need to be enabled explicitly.

umutondersu:
> Hey everyone 👋 I remember a few related conversations that happen lately on this topic, let me share them here
> 
> #1458 #1331
> 
> I hope they can be useful, cheers ✌️

- This setup config does not clash with blink.cmp so in that sense It should be okay.
- I would be able to support this in an ongoing basis
- I think using the native way would provide a better experience for newbies but that's up to debate

init.lua
yetyetanotherusername: Hmm, that kinda seems like a step back usability-wise, just my 2 cents.

In either case, if this functionality is no longer used, shouldn't `mason-lspconfig` then be removed from dependencies?

kaezrr:
In my opinion, we shouldn't bother with backwards compatibility since Kickstart is mostly aimed at beginners who will download the latest version of Neovim as per the instructions in the repo readme

umutondersu:
I have removed backwards compatibility and added useful comments for LSPs that still require the old setup

init.lua
1Naim: For mason-lspconfig, can't you do something like

```lua
                handlers = {
                    function(server_name)
                        local config = servers[server_name] or {}
                        vim.lsp.config(server_name, config)
                        vim.lsp.enable(server_name)
                    end,
                },
```

this should function the same as previously and is just effectively porting over the previous `require('lspconfig')[server_name].setup(server)` to `vim.lsp*`

init.lua
```diff
@@ -716,15 +710,29 @@ require('lazy').setup({
...
-            local server = servers[server_name] or {}
-            -- This handles overriding only values explicitly passed
-            -- by the server configuration above. Useful when disabling
-            -- certain features of an LSP (for example, turning off formatting for ts_ls)
-            server.capabilities = vim.tbl_deep_extend('force', {}, capabilities, server.capabilities or {})
-            require('lspconfig')[server_name].setup(server)
+            local config = servers[server_name] or {}
+            vim.lsp.config(server_name, config)
+            vim.lsp.enable(server_name)
           end,
```
1Naim: As an additional FYI, this was taken from my config and as such is missing the `vim.tbl_deep_extend` line. I'm not 100% sure on how vim.lsp.* actually interacts with capabilities, but if kickstart wants to give the ability to users to override default LSP configs, might want to add that back.

Related: https://cmp.saghen.dev/installation.html#merging-lsp-capabilities

However, I've never seen the capabilites block anywhere in the file so that line might just be useless (standalone and currently)

```lua
local capabilities = {
  textDocument = {
    foldingRange = {
      dynamicRegistration = false,
      lineFoldingOnly = true
    }
  }
}
```

yetyetanotherusername (approved):

iton0:
Messing around with this and found that this will not respect the lsp_format = "fallback" option in the conform.nvim config. To replicate:

1. comment out the stylua formatter line from conform.nvim config
2. save and reload neovim
3. overindent a line or any other formatting change
4. write file 

For me it did not fallback on the lua_ls server for formatting.

umutondersu:
> Messing around with this and found that this will not respect the lsp_format = "fallback" option in the conform.nvim config. To replicate:
> 
>     1. comment out the stylua formatter line from conform.nvim config
> 
>     2. save and reload neovim
> 
>     3. overindent a line or any other formatting change
> 
>     4. write file
> 
> 
> For me it did not fallback on the lua_ls server for formatting.

I couldn't replicate the issue. I do not think it would be related to these changes as conform.nvim does not use an LSP. In any case If you could check your conform logs by using `:ConformInfo` It could reveal more info about the issue.

init.lua
cbackas: The new way of merging LSP configs is done like this: ([:h vim.lsp.config](https://neovim.io/doc/user/lsp.html#vim.lsp.config()))
```lua
vim.lsp.config('*', {
  capabilities = ...
})
```

Since I guess kickstart uses blink now, [blink handles setting some * capabilities already](https://github.com/Saghen/blink.cmp/blob/main/plugin/blink-cmp.lua). If the user wanted to to add other capabilities they can put them in a `vim.lsp.config('*', ...)` call and it will merge in. My understanding is if they want to override **completion related capabilities** they would need to make sure their `vim.lsp.config('*', ...)` call is *after* blink is setup.

1Naim: Right, I guess in that case those users can handle it on a per-language basis, something like below perhaps

```diff
@@ -200,11 +200,19 @@ return {
                 ensure_installed = { 'clangd' },
                 automatic_installation = false,
                 handlers = {
+                    -- Default handler
                     function(server_name)
                         local config = servers[server_name] or {}
                         vim.lsp.config(server_name, config)
                         vim.lsp.enable(server_name)
                     end,
+
+                    clangd = function()
+                        vim.lsp.config('clangd', {
+                            capabilities = capabilites_overrides_here,
+                        })
+                        vim.lsp.enable('clangd')
+                    end
                 },
             }
         end
```

iton0:
> > Messing around with this and found that this will not respect the lsp_format = "fallback" option in the conform.nvim config. To replicate:
> > ```
> > 1. comment out the stylua formatter line from conform.nvim config
> > 
> > 2. save and reload neovim
> > 
> > 3. overindent a line or any other formatting change
> > 
> > 4. write file
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > For me it did not fallback on the lua_ls server for formatting.
> 
> I couldn't replicate the issue. I do not think it would be related to these changes as conform.nvim does not use an LSP. In any case If you could check your conform logs by using `:ConformInfo` It could reveal more info about the issue.

Accidentally commented out my lua_ls server lol. No longer issue

init.lua
cbackas: Since the servers table is getting referenced and passed into `vim.lsp.config`, adding `capabilities` to the 'clangd' entry in the servers table above would do the same thing

1Naim: Yep, you're right.

init.lua
```diff
@@ -716,15 +710,29 @@ require('lazy').setup({
...
+      --
+      -- if server == 'example_server' or server == 'example_server2' then
+      --   -- This handles overriding only values explicitly passed
+      --   -- by the server configuration above. Useful when disabling
+      --   -- certain features of an LSP (for example, turning off formatting for ts_ls)
+      --   local capabilities = require('blink.cmp').get_lsp_capabilities()
+      --   config.capabilities = vim.tbl_deep_extend('force', {}, capabilities, config.capabilities or {})
+      --   require('mason-lspconfig')[server].setup(config)
+      --   return
+      -- end
```
roryhen: Thanks for this contribution, was just trying this out, should it be `server_name` here?
```suggestion
      -- if server_name == 'example_server' or server_name == 'example_server2' then
      --   -- This handles overriding only values explicitly passed
      --   -- by the server configuration above. Useful when disabling
      --   -- certain features of an LSP (for example, turning off formatting for ts_ls)
      --   local capabilities = require('blink.cmp').get_lsp_capabilities()
      --   config.capabilities = vim.tbl_deep_extend('force', {}, capabilities, config.capabilities or {})
      --   require('mason-lspconfig')[server_name].setup(config)
      --   return
      -- end
```

init.lua
1Naim: mason-lspconfig.nvim v2.0.0 removes this now. It auto-enables installed LSPs if they're using the new vim.lsp.* API. See https://github.com/mason-org/mason-lspconfig.nvim/releases/tag/v2.0.0

init.lua
```diff
@@ -711,20 +711,34 @@ require('lazy').setup({
...
-      }
+      -- Handle lsp setups differently based on Neovim version
+      -- See :help vim.lsp.config and :help vim.lsp.enable for Neovim 0.11+
+      if vim.fn.has 'nvim-0.11' == 1 then
+        for server, config in pairs(servers) do
+          -- This handles overriding only values explicitly passed
+          -- by the server configuration above. Useful when disabling
+          -- certain features of an LSP (for example, turning off formatting for ts_ls)
+          config.capabilities = vim.tbl_deep_extend('force', {}, capabilities, server.capabilities or {})
+          vim.lsp.config(name, config)
```
1Naim: I guess the original hunk of the patch is relevant (and should be used?) now.

jaek187: Sorry I don't know how to mention lines, but shouldn't it be `vim.lsp.config(server, config)`? As well as `vim.lsp.enable(server)`

1Naim: Yes that was fixed in a later commit in this PR, I was quoting from an earlier one.

pandar00:
AFAIK `mason-lspconfig` [setting's](https://github.com/mason-org/mason-lspconfig.nvim?tab=readme-ov-file#default-configuration) `handler` is a no-op. ([PR](https://github.com/mason-org/mason-lspconfig.nvim/commit/80f2cd7734ff21da4cf1489c9695e440cdcd139f))
LSP settings are moved to `after/lsp/<lang_server_name>.lua` ([credit](https://zenn.dev/vim_jp/articles/migrate-nvim-lspconfig-v0_11); in JP)
[lsp-config doc](https://neovim.io/doc/user/lsp.html#lsp-config) also explains the config is merged with the default.
> Note: The merge semantics of configurations follow the behaviour of [vim.tbl_deep_extend()](https://neovim.io/doc/user/lua.html#vim.tbl_deep_extend()).

I've been updating my old neovim configs with the latest kickstart but the LSP settings weren't reflected in `:LspInfo`, but after following the above, LSP correctly reads them

umutondersu:
> AFAIK `mason-lspconfig` [setting's](https://github.com/mason-org/mason-lspconfig.nvim?tab=readme-ov-file#default-configuration) `handler` is a no-op. ([PR](https://github.com/mason-org/mason-lspconfig.nvim/commit/80f2cd7734ff21da4cf1489c9695e440cdcd139f)) LSP settings are moved to `after/lsp/<lang_server_name>.lua` ([credit](https://zenn.dev/vim_jp/articles/migrate-nvim-lspconfig-v0_11); in JP) [lsp-config doc](https://neovim.io/doc/user/lsp.html#lsp-config) also explains the config is merged with the default.
> 
> > Note: The merge semantics of configurations follow the behaviour of [vim.tbl_deep_extend()](https://neovim.io/doc/user/lua.html#vim.tbl_deep_extend()).
> 
> I've been updating my old neovim configs with the latest kickstart but the LSP settings weren't reflected in `:LspInfo`, but after following the above, LSP correctly reads them

If we were to move LSP configs to their own file, I don't know how mason-tool-installer would work without the servers table maybe by iterating over file names?

In your config I see that you still have the servers table. If It is still gonna be preset wouldn't be better to use vim.lsp.config with a loop (like my first hunk) rather than having 2 separate places for LSP servers?

umutondersu:
As mentioned in previous conversations, Mason v2 has started introducing a significant delay during startup. For now, I’ve pinned the Mason version to avoid this issue.

From my testing, the problem appears to be related to mason-tool-installer; without it, I don’t experience any delays. It would be helpful to hear from someone with more insight into this before proceeding further with Mason v2.

pxy1337:
My personal temporary solution is to set `debounce_hours` in `mason-tool-installer`, to 24 in my case.

umutondersu:
After the recent updates mason-tool-installer seems to not have this issue anymore. I will commit a change for Mason V2

init.lua
```diff
@@ -482,8 +482,8 @@ require('lazy').setup({
       -- Automatically install LSPs and related tools to stdpath for Neovim
       -- Mason must be loaded before its dependents so we need to set it up here.
       -- NOTE: `opts = {}` is the same as calling `require('mason').setup({})`
-      { 'mason-org/mason.nvim', opts = {} },
-      'mason-org/mason-lspconfig.nvim',
+      { 'mason-orgmason.nvim', opts = {} },
```
DanWlker: i think you're missing a /

init.lua
```diff
@@ -719,20 +713,14 @@ require('lazy').setup({
...
-            -- certain features of an LSP (for example, turning off formatting for ts_ls)
-            server.capabilities = vim.tbl_deep_extend('force', {}, capabilities, server.capabilities or {})
-            require('lspconfig')[server_name].setup(server)
-          end,
-        },
-      }
+      -- Installed LSPs are configured and enabled automatically with mason-lspconfig
+      -- The loop below is for overriding the default configuration of LSPs with the ones in the servers table
+      for server_name, config in pairs(servers) do
+        vim.lsp.config(server_name, config)
```
DanWlker, line 724: Also, before all the servers work, should we just use `require("lspconfig").server_name.setup{}`? since that currently works for everything

tienbg-zigexn:
Currently I'm using the handlers in mason-lspconfig like this:
```lua
                handlers = {
                    function(server_name)
                        if server_name == "standardrb" or server_name == "rubocop" then
                            local personal_path = vim.fn.expand("~") .. "/programming/me"
                            local current_dir = vim.fn.getcwd()
                            local in_personal_project = string.find(current_dir, "^" .. personal_path) ~= nil

                            if
                                (server_name == "standardrb" and not in_personal_project)
                                or (server_name == "rubocop" and in_personal_project)
                            then
                                return
                            end
                        end

                        local server = servers[server_name] or {}

                        server.capabilities = vim.tbl_deep_extend("force", {}, capabilities, server.capabilities or {})
                        require("lspconfig")[server_name].setup(server)
                    end,
                },
```

The purpose is to load `standardrb` only in personal projects and `rubocop` on others. How can I migrate this into new version?

umutondersu:
> Currently I'm using the handlers in mason-lspconfig like this:
> 
> ```lua
>                 handlers = {
>                     function(server_name)
>                         if server_name == "standardrb" or server_name == "rubocop" then
>                             local personal_path = vim.fn.expand("~") .. "/programming/me"
>                             local current_dir = vim.fn.getcwd()
>                             local in_personal_project = string.find(current_dir, "^" .. personal_path) ~= nil
> 
>                             if
>                                 (server_name == "standardrb" and not in_personal_project)
>                                 or (server_name == "rubocop" and in_personal_project)
>                             then
>                                 return
>                             end
>                         end
> 
>                         local server = servers[server_name] or {}
> 
>                         server.capabilities = vim.tbl_deep_extend("force", {}, capabilities, server.capabilities or {})
>                         require("lspconfig")[server_name].setup(server)
>                     end,
>                 },
> ```
> 
> The purpose is to load `standardrb` only in personal projects and `rubocop` on others. How can I migrate this into new version?

Mason-lspconfig enables all installed LSPs by default, You can disable individual LSPs with `vim.lsp.enable(server_name,enable)` instead of returning.

oriori1703 (approved):

init.lua
```diff
@@ -655,12 +655,6 @@ require('lazy').setup({
         },
       }
 
-      -- LSP servers and clients are able to communicate to each other what features they support.
-      --  By default, Neovim doesn't support everything that is in the LSP specification.
-      --  When you add blink.cmp, luasnip, etc. Neovim now has *more* capabilities.
-      --  So, we create new capabilities with blink.cmp, and then broadcast that to the servers.
-      local capabilities = require('blink.cmp').get_lsp_capabilities()
```
carloscalla, line 662: I think this blink capabilities should be kept in the code and just added to all lsp servers.
```lua
      vim.lsp.config("*", {
        capabilities = capabilities
      })
```
Or be added in the `for` loop below:

```lua
      for server_name, config in pairs(servers) do
        config.capabilities = vim.tbl_deep_extend('force', {}, capabilities, config.capabilities or {})
        vim.lsp.config(server_name, config)
      end
```

1Naim: > think this blink capabilities should be kept in the code and just added to all lsp servers

Why? It's already getting called within the plugin itself https://github.com/Saghen/blink.cmp/tree/main/plugin#L1-L5

oriori1703 (changes requested):
maybe you should only enable lsps that were installed using `ensure_installed`.
This way you will mirror the original behaviour.

i.e.
```lua
      ---@type MasonLspconfigSettings
      ---@diagnostic disable-next-line: missing-fields
      require('mason-lspconfig').setup {
        automatic_enable = vim.tbl_keys(servers or {}),
      }
```

init.lua
junkblocker: Does this set it for any LSP configured after blink  has done it or does it apply to previously configured LSPs only? Is that what `*` is here?

init.lua
pbnj-dragon: In my testing, the lsp server configs do not get merged with the defaults from `lspconfig`:

```lua
for server_name, server_config in pairs(servers) do
    require("lspconfig")[server_name].setup(server_config)
end
```

Unless I am missing something.

pbnj: @DanWlker - In my testing, the lsp server configs do not get merged with the defaults from `lspconfig`:

```lua
for server_name, server_config in pairs(servers) do
    require("lspconfig")[server_name].setup(server_config)
end
```

Unless I am missing something.

oriori1703 (approved):

init.lua
DanWlker: image
image

I'm only using `require("lspconfig").server_name.setup{}`, not quite sure what you mean by lsp configs not getting merged

init.lua
1Naim: Any.

init.lua
```diff
@@ -699,6 +693,11 @@ require('lazy').setup({
           },
         },
       }
+      ---@type MasonLspconfigSettings
+      ---@diagnostic disable-next-line: missing-fields
+      require('mason-lspconfig').setup {
+        automatic_enable = vim.tbl_keys(servers or {}),
```
SetsuikiHyoryu, line 699: @umutondersu @oriori1703 

I looked into the source code of mason-lspconfig. When the keys of `servers` are passed as a table to `automatic_enable`, it causes mason-lspconfig to ignore any LSPs that are installed via Mason but not explicitly listed in `servers`.

In the `mason-lspconfig.nvim\doc\mason-lspconfig.txtmason-lspconfig.nvim\doc\mason-lspconfig.txt` document, it mentions that doing this will **only** start **certain** servers:

````lua
-- To only enable certain servers to be automatically enabled:
-- ```lua
--   automatic_enable = {
--     "lua_ls",
--     "vimls"
--   }
-- ```
````

In the `setup` that in the `mason-lspconfig.nvim\lua\mason-lspconfig\init.lua`，if `automatic_enable` is not set to false, both `enable_all` and `init` will be executed:

```lua
function M.setup(config)
    if config then
        settings.set(config)
    end

    --- ...

    local registry = require "mason-registry"
    registry.refresh(vim.schedule_wrap(function(success, updated_registries)
        if success and #updated_registries > 0 and settings.current.automatic_enable ~= false then
            require("mason-lspconfig.features.automatic_enable").enable_all()
        end
    end))

    if settings.current.automatic_enable ~= false then
        require("mason-lspconfig.features.automatic_enable").init()
    end

    -- ...
end
```

In `mason-lspconfig\features\automatic_enable.lua`, The logic for enabling servers in `init` and `enable_all` is the same:

```lua
return {
    init = function()
        enabled_servers = {}
        _.each(enable_server, registry.get_installed_package_names())
        -- We deregister the event handler primarily for testing purposes where .setup() is called multiple times in the
        -- same instance.
        registry:off("package:install:success", enable_server_scheduled)
        registry:on("package:install:success", enable_server_scheduled)
    end,
    enable_all = function()
        _.each(enable_server, registry.get_installed_package_names())
    end,
}
```

However, in `enable_server`, if `automatic_enable` is a table and an LSP installed by Mason is not listed in `automatic_enable`, it will not execute `vim.lsp.config` and `vim.lsp.enable`.

```lua
local function enable_server(mason_pkg)
    -- ...

    local lspconfig_name = mappings.get_mason_map().package_to_lspconfig[mason_pkg]

    -- ...

    local automatic_enable = settings.current.automatic_enable

    if type(automatic_enable) == "table" then
        local exclude = automatic_enable.exclude
        if exclude then
            if _.any(_.equals(lspconfig_name), exclude) then
                -- This server is explicitly excluded.
                return
            end
        else
            if not _.any(_.equals(lspconfig_name), automatic_enable) then
                -- This server is not explicitly enabled.
                return
            end
        end
    elseif automatic_enable == false then
        return
    end

    -- ...

    vim.lsp.enable(lspconfig_name)

    -- ...
end
```

In other words, even if I’ve already installed the LSPs via Mason, I still have to list all the LSPs I use in `servers` or in `or {}`. Isn't that kind of ... not very "lazy"?

oriori1703: It seems you are right about `automatic_enable`.
If set to true it will enable all the LSPs, but when given a table it will only enable those LSPs.
After looking into it more, enabling all the LSPs also seems to be the old behaviour mason.

Despite that , I still prefer declarativly specifying all the LSPs in the config file, instead of downloading them using the mason interface.
This way I can commit those LSPs and have a reproducible setup.
This is the same reason I commit the `lazy-lock.json` file.

Though I guess this is a matter of taste, and could break setups for people that were used to the old way, so I'm ok with both options :)

@umutondersu @SetsuikiHyoryu  what do you think?

SetsuikiHyoryu: @oriori1703 @umutondersu 

Personally, I prefer the more “lazy” approach, but I tend to follow the method recommended by the Neovim official because I feel it’s more native. That means the operational habits won’t differ too much even when I have to work in a barebones environment. However, I’m not entirely sure what the official philosophy is on this. Judging from the existence of the `vim.lsp.enable` interface, it seems that the official intent is for LSP to be explicitly enabled.

~~Perhaps, if Mason provides an interface to retrieve the list of installed packages, a compromise could be to merge that list into the loop that calls `vim.lsp.enable` (the table passed to `automatic_enable`).~~

Sorry, it seems that  
`_.each(enable_server, registry.get_installed_package_names())`  
is already configuring only the installed packages. Setting `automatic_enable = true` doesn't appear to enable LSPs that haven't been installed.

Graeme22:
My config was broken until I added the changes from this PR. As far as I understand Mason v2 doesn't support the handler approach that was being used before, so none of my LSP configuration was actually being applied.

Mithrandir2k18:
If this PR aims to only support nvim 0.11+, then it maybe also should remove the `client_supports_method` function?

Mithrandir2k18:
Also, maybe I'm just blind, I didn't checkout the PR, but this [name variable](https://github.com/nvim-lua/kickstart.nvim/blob/3f0cd7546940beb1283ef1442a944d10c48a8431/init.lua#L722) seems undefined?

Edit: my bad, I thought I was on the forks commit list, not on the PRs, where the commit order is inverted.

umutondersu:
@Mithrandir2k18 
> If this PR aims to only support nvim 0.11+, then it maybe also should remove the client_supports_method function?

I started the PR supporting older versions as well, However most people thought it wasn't necessary because kickstart only support the latest neovim version and having cleaner code would be better for beginners.

I didn't touch the clients_supports_method as this pr is only focused on the new LSP API of neovim

> Also, maybe I'm just blind, I didn't checkout the PR, but this [name variable](https://github.com/nvim-lua/kickstart.nvim/blob/3f0cd7546940beb1283ef1442a944d10c48a8431/init.lua#L722) seems undefined?

It has been fixed. You are looking at an older version of the branch

spundun:
> > Kickstart.nvim targets only the latest ['stable'](https://github.com/neovim/neovim/releases/tag/stable) and latest ['nightly'](https://github.com/neovim/neovim/releases/tag/nightly) of Neovim.
> 
> Do we really need backward compatibility with Neovim 0 10? I'd consider going with a lighter, happy-path solution. Since the project mainly targets newcomers and servers educational purposes, they'd likely prefer to avoid dealing with legacy approaches.

I personally love to keep the code free of legacy dependencies, but Ubuntu, Debian and some Fedoras don't yet have 0.11 , maybe someone familiar with the code could contribute an upgrade to the most common distros?

https://packages.ubuntu.com/search?keywords=neovim&searchon=names&suite=plucky&section=all

https://packages.debian.org/search?keywords=neovim&searchon=names&suite=stable&section=all

https://packages.fedoraproject.org/pkgs/neovim/neovim/

guru245:
Hi @spundun ,

Neovim also provides a prebuilt package: AppImage and tarball. We don't have to build Neovim for the distrubution. Please check out this link: https://github.com/neovim/neovim/releases/tag/v0.11.2

spundun:
Hi @guru245 , Thanks for the response. That's a cool new thing I learned today :).

The README.md just says "If you are experiencing issues, please make sure you have the latest versions.". Since kickstart puts priority on being new-user-friendly, I'd suggest you add more detailed guidelines there. To many users, "make sure you have the latest version" might just mean making sure you have all the latest software updates from your package manager, especially if you're on the latest version of some standard OS.

Maybe adding detailed instructions along the lines... if `nvim --version` reports a version lower than the Required version then here are the instructions on how to install it in the user-file-system and set it as your editor before you begin.

mybuddymichael:
I spent _way_ too long trying to figure out why inlay hints wouldn't appear anywhere! Turns out my LSP settings weren't even being loaded. I'm super glad I found this PR!

Ktsierra:
any updates on this pr? the current implementation  of ```mson-lspconfig```  doesn't work, ```handlers``` was removed in V2.

logrusx:
I recommend #1663 over this PR.

It states clearly what it does, it does it correctly and stays to the point.

This one is actually dealing with mason-lspconfig 2.0, rather than nvim 0.11, does more than it says and pushes authors preferences.

init.lua
logrusx, line 699: > Despite that , I still prefer declarativly specifying all the LSPs in the config file, instead of downloading them using the mason interface. This way I can commit those LSPs and have a reproducible setup. This is the same reason I commit the `lazy-lock.json` file.

This is unacceptable for plugins that need to set additional settings before the LSP is setup. For one - nvim-java for jdtls. It is also absolutely unnecessary.

Pushing your preferences into a public repository is not very kind to others with different preferences.

I find #1663 better than this one.

NaniNoni:
Is there anything major preventing this from being merged? I merged this patchset into my personal config and everything works fine.

init.lua
```diff
@@ -719,20 +718,14 @@ require('lazy').setup({
...
-            -- certain features of an LSP (for example, turning off formatting for ts_ls)
-            server.capabilities = vim.tbl_deep_extend('force', {}, capabilities, server.capabilities or {})
-            require('lspconfig')[server_name].setup(server)
-          end,
-        },
-      }
+      -- Installed LSPs are configured and enabled automatically with mason-lspconfig
+      -- The loop below is for overriding the default configuration of LSPs with the ones in the servers table
+      for server_name, config in pairs(servers) do
+        vim.lsp.config(server_name, config)
```
justinmk, line 724: This is the right approach 👍 . Using `vim.lsp.config` is the only supported approach for Nvim 0.11+.

Nvim 0.11+ should not use `require('lspconfig')`. 

nvim-lspconfig will soon be deleting all of its [`lua/lspconfig/*`](https://github.com/neovim/nvim-lspconfig/tree/master/lua/lspconfig) configs (which have been deprecated for 6+ months) in favor of [`lsp/*`](https://github.com/neovim/nvim-lspconfig/tree/master/lsp).

logrusx, line 724: That alone won't do anything. You can't just swap [servername].setup with vim.lsp.config. You should now provide the config in its entirety.
Also you need to enable the server somewhere.

codevogel:
I just went down a rabbit hole of issues. 

I initially thought the `svelte-language-server` had a bug which prevented me from importing through `$lib`, but it turns out it all boiled down to my `kickstart-modular.nvim` fork including this deprecated configuration, which prevented the passing of _all_ defined lsp settings to the actual server.

Is there anything holding this back? There seem to be a few PR's open all attempting to addres the same issue, so I think it's fair to say that it should probably get some priority by now.

I'd love for this PR to be merged so it can propagate down to [kickstart-modular](https://github.com/dam9000/kickstart-modular.nvim) and eventually my fork.

yetyetanotherusername:
> Is there anything holding this back?

This repo is in limbo, perhaps https://github.com/nvim-lua/kickstart.nvim/pull/1627 can clear the situation up for you.

codevogel:
Ah, @yetyetanotherusername thanks for pointing me there.

oriori1703:
This has finally been implemented by TJ (with some fixes by me inspired by this PR).

Thanks for your help :D (and sorry for not getting merged)

PR #1483, opened 2025-04-20
Improve fidget.nvim to handle all notifications

kaezrr:
Since we are already using fidget.nvim to handle LSP updates, why not extend its functionality to override vim.notify() and allow it to show all notifications for more consistency and better UX.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -478,8 +478,15 @@ require('lazy').setup({
       'williamboman/mason-lspconfig.nvim',
       'WhoIsSethDaniel/mason-tool-installer.nvim',
 
-      -- Useful status updates for LSP.
-      { 'j-hui/fidget.nvim', opts = {} },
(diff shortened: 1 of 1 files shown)


oriori1703:
IMO the notifications are little bit too small and out of the way, for anything important.
Maybe you can change their size?

kaezrr:
> IMO the notifications are little bit too small and out of the way, for anything important.

Yes that is by design, fidget.nvim is designed to provide low profile and non distracting notifications. 

If we want more eye catching notifications we can use something like nvim-notify or snacks-notifier

feoh:
I'm very concerned that people won't like this change. I'd need to see a *lot* of posts in support in order to merge this.

feoh:
I'm sorry I don't think this change is a good for for new Neovim users. Thank you for your contribution in any case.

If you feel strongly about this and want to re-open it, please find other people who can chime in in support of this change.

PR #1488, opened 2025-04-22
feat(plugin): tutorial.nvim

### LINKED ISSUE #1487: plugin request: tutorial.nvim
I've made a plugin specifically to be added to something like kickstart.nvim. It's called [tutorial.nvim](https://github.com/Hashino/tutorial.nvim).

It keeps a floating buffer with the most (in my opinion) useful keymaps for absolute beginners.

I believe it'd be a great addition to kickstart.nvim

Hashino:
simple plugin to show a neovim tutorial on screen at all times

closes #1487

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -248,6 +248,11 @@ require('lazy').setup({
   -- Use `opts = {}` to automatically pass options to a plugin's `setup()` function, forcing the plugin to be loaded.
   --
 
+  { -- shows a floating buffer with the most useful vim keymaps
+    'Hashino/tutorial.nvim',
+    opts = {},
+  },
+
   -- Alternatively, use `config = function() ... end` for full control over the configuration.
   -- If you prefer to call `setup` explicitly, use:
   --    {
--- a/init.lua
+++ b/init.lua
@@ -248,11 +248,6 @@ require('lazy').setup({
   -- Use `opts = {}` to automatically pass options to a plugin's `setup()` function, forcing the plugin to be loaded.
   --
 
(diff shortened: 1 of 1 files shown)


init.lua
```diff
@@ -248,6 +248,11 @@ require('lazy').setup({
   -- Use `opts = {}` to automatically pass options to a plugin's `setup()` function, forcing the plugin to be loaded.
   --
 
+  { -- shows a floating buffer with the most useful vim keymaps
+    'Hashino/tutorial.nvim',
+    opts = {},
+  },
```
5brian: looks pretty extreme to be active by default to existing users, maybe comment it out?

Hashino: my rationale is that the kind of user that needs this plugin the most don't even know how to enable/disable a plugin yet.

I envision this plugin as being the first one users will remove from their configs. my goal is strictly to avoid the "can't quit vim" meme scenario

guru245:
I have a question. Doesn't this floating buffer conflict with which-key window?

Update: I got my hands on the plugin and I can see the conflict. The following is one of the example that happens with Telescope.
image

Hashino:
@guru245 made an update to the plugin that fixes incompatibilities with Telescope.

on the which-key case, the plugin still displays but which-key shows on top, which I think is a reasonable behavior
image

Hashino:
There's a bug that I can't quite fix yet. After last update (ignore certain buffers) the plugin buffer disappears if the buffer is focused.

Will work on fixing it

Hashino:
fixed the bug on last commit

Issue #1264, opened 2024-12-13
vim.opt will be discontinued?


git-dimiz:
Hi,

I'm quite new to the neovim ecosystem and stubled upon this project. Thanks for the great work :slightly_smiling_face: . I'm currently a bit confused regarding `vim.o` vs `vim.opt` and stubbled upon this comment in the github issue: https://github.com/neovim/neovim/issues/30383#issuecomment-2351519326 . As far as I can tell `lua.init` is also using `vim.opt` which will be discontinued? Shouldn't this script then rather encurage to use `vim.o`?

feoh:
Hi.

Thanks for pointing this out. It's a good distant early warning.

I'm going to leave this bug around because we'll probably eventually need to do the work of translating all the calls, but we are a "leaf node" in the Neovim ecosystem.

We'll need ALL the plugins this project encompasses to be fixed first, then our configuration will follow suit.

oriori1703:
https://github.com/nvim-lua/kickstart.nvim/pull/1495 replace all but 2 occurrences of vim.opt, and commented on the remaining ones.

I think this is fine to close until real steps are taken upstream to deprecate vim.opt

PR #1495, opened 2025-04-26
Replace vim.opt with vim.o

oriori1703:
Even though this change is not mandatory until all of the plugins use the new `vim.o` interface, as discussed in #1264 , I think there it is still beneficial switching to `vim.o`. 

In contrast to `vim.opt`, `vim.o` opens the documentation on hover, and provides type hints which will be useful for beginners (and also experts who are just lazy ;).

In addition it is more intuitive to get the value of the option (e.g. `vim.bo.modifiable` instead of `vim.opt_local.modifiable:get()`).

The main downside is that it is less convenient in the cases where `vim.opt` provides a table (This is tracked in https://github.com/neovim/neovim/issues/20107).
So for now I left `vim.opt.listchars` and `vim.opt.rtp` unchanged.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -223,8 +223,11 @@ if not (vim.uv or vim.loop).fs_stat(lazypath) then
   if vim.v.shell_error ~= 0 then
     error('Error cloning lazy.nvim:\n' .. out)
   end
-end ---@diagnostic disable-next-line: undefined-field
-vim.opt.rtp:prepend(lazypath)
+end
+
+---@type vim.Option
+local rtp = vim.opt.rtp
+rtp:prepend(lazypath)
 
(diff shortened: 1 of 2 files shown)


brianhuster:
I think there should be document about the difference between them, and why you use `vim.o` for this but keep `vim.opt` for that

oriori1703:
> I think there should be document about the difference between them, and why you use `vim.o` for this but keep `vim.opt` for that

I added a comment explaining why I choose to use `vim.opt` and referencing the manual for an explanation about the difference.
What do you think?

brianhuster:
LGTM

feoh (approved):

PR #1498, opened 2025-04-27
fix: correct gitsigns mapping for <leader>tD

guru245:
The previous mapping <leader>tD was described as "[T]oggle git show [D]eleted" but was mapped to preview_hunk_inline, which didn't match the description. This commit updates the mapping to toggle_deleted, which fits the description.

**NOTE**


guru245:
toggle_deleted has been deprecateed. This PR is wrong.

PR #1516, opened 2025-05-08
Change to Mason's new address

guru245:
Mason has changed their URL recently and so we also apply accordingly

https://github.com/mason-org/mason.nvim/releases


raiderss:
Interesting solution to this problem! I've been working on similar challenges recently.

feoh (approved):

PR #1524, opened 2025-05-11
fix #1519: disable completion in *only* shell command mode

### LINKED ISSUE #1519: Bug: External Commands via `:!` Freeze Neovim
## Before Reporting an Issue

- [x] **By checking this, I confirm that the above steps are completed. I understand leaving this unchecked will result in this report being closed immediately.**

## Describe the bug

While going through the `:Tutor` in Neovim, I tried executing an external command using `:!` . When I typed `:!echo HELLO`, only the first character (`e`) was typed, and after that, the UI completely froze. No keystrokes worked, I couldn't switch modes, or exit — Neovim became unresponsive

## To Reproduce

1. Open Neovim with Kickstart config.
2. Enter Normal Mode.
3. Type `:!`.
4. Try to type an external shell command like `echo HELLO.
5. Observe that only the first character is entered and Neovim becomes unresponsive.

##  Additional Info
When I tried running the same command (`:!echo HELLO`) using Neovim without any custom config (i.e., using `nvim -u NONE`), the command worked perfectly, and there were no freezes or input issues.

## Desktop

- OS: Linux (Ubuntu) using WSL
- Terminal: Ubuntu Terminal

## Neovim Version

```
NVIM v0.12.0-dev 
Build type: RelWithDebInfo
LuaJIT 2.1.1703358377
```

guru245:
When inside of git bash or WSL on windows, you may experience a hang with shell commands.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -857,6 +857,12 @@ require('lazy').setup({
         default = { 'lsp', 'path', 'snippets', 'lazydev' },
         providers = {
           lazydev = { module = 'lazydev.integrations.blink', score_offset = 100 },
+          cmdline = {
+            -- ignores cmdline completions when executing shell commands
+            enabled = function()
+              return vim.fn.getcmdtype() ~= ':' or not vim.fn.getcmdline():match "^[%%0-9,'<>%-]*!"
+            end,
+          },
         },
       },
 


guru245:
On second thought, I don't think this PR is necessary because this bug only happens when git bash or WSL is used. If we apply this change, it will disable shell completion in the native Linux environment.

PR #1527, opened 2025-05-12
fix(mason-lspconfig): add missing field `automatic_enable` in setup

SetsuikiHyoryu:
I know this field is set by default, but if it's not explicitly written, Lua will throw an error:

```txt
Diagnostics:
1. Missing required fields in type `MasonLspconfigSettings`: `automatic_enable` [missing-fields]
```

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -722,6 +722,7 @@ require('lazy').setup({
       require('mason-lspconfig').setup {
         ensure_installed = {}, -- explicitly set to an empty table (Kickstart populates installs via mason-tool-installer)
         automatic_installation = false,
+        automatic_enable = true, -- See https://github.com/mason-org/mason-lspconfig.nvim?tab=readme-ov-file#default-configuration
         handlers = {
           function(server_name)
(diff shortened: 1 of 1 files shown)


nickgnd:
`automatic_enable` has been introduced in the latest version of Mason 2.0.

I think this change should be included in https://github.com/nvim-lua/kickstart.nvim/pull/1475 in case.

SetsuikiHyoryu:
> `automatic_enable` has been introduced in the latest version of Mason 2.0.
> 
> I think this change should be included in #1475 in case.

@nickgnd I'm agree with you.

carloscalla:
This field `automatic_enable` was introduced in mason v2 but also `automatic_installation` and `handlers` were deprecated. As mentioned above, a more complete migration to mason v2 is being done in https://github.com/nvim-lua/kickstart.nvim/pull/1475

Please consider closing this PR to avoid confusion.

SetsuikiHyoryu:
> Please consider closing this PR to avoid confusion.

Ok.

PR #1532, opened 2025-05-12
fix: Load guess-indent.nvim

rivenirvana:


feoh (approved):

oriori1703 (approved):

init.lua
```diff
@@ -247,7 +247,10 @@ rtp:prepend(lazypath)
 -- NOTE: Here is where you install your plugins.
 require('lazy').setup({
   -- NOTE: Plugins can be added with a link (or for a github repo: 'owner/repo' link).
-  'NMAC427/guess-indent.nvim', -- Detect tabstop and shiftwidth automatically
+  {
+    'NMAC427/guess-indent.nvim', -- Detect tabstop and shiftwidth automatically
+    opts = {},
```
guru245, line 252: How about we add this?
    `event = { 'BufReadPost' },`

init.lua
```diff
@@ -247,7 +247,10 @@ rtp:prepend(lazypath)
 -- NOTE: Here is where you install your plugins.
 require('lazy').setup({
   -- NOTE: Plugins can be added with a link (or for a github repo: 'owner/repo' link).
-  'NMAC427/guess-indent.nvim', -- Detect tabstop and shiftwidth automatically
+  {
+    'NMAC427/guess-indent.nvim', -- Detect tabstop and shiftwidth automatically
```
guru245: How about we add this?
```suggestion
    'NMAC427/guess-indent.nvim', -- Detect tabstop and shiftwidth automatically
    event = { 'BufReadPost' },
```

rivenirvana: Depends, I'd like to know how necessary this is for a pretty small plugin.

I looked at the source, and it looks like if I have to specify loading it on BufReadPost, I'd also have to specify loading it on BufNewFile just to cover the cases where the plugin is intended to work. And will have to watch out for future changes to the plugin's triggers (unlikely).

I wouldn't mind, since the `event` key is already documented and I don't have to do it myself, though I also want to know if it's an issue for the maintainers for a key to be used before its documentation, cause then I'd have to move the code too.

guru245: Thank you for the answer. I agree that it's overkill to use `event` for such a small plugin.

rivenirvana: Checked a bit, and the plugin does not load in a simple scenario (nv to empty buffer -> paste some code -> `:w filename` -> continue editing), when BufReadPost and BufNewFile are both specified.
I think it's unnecessary to specify 3 separate events as entry points for a plugin that's attached to a very ubiquitous/common function of the editor (reading existing files/watching for new files). Most people invoke nv with a file to read already, too. Thank you for the suggestion, anyway.

guru245:
Hi @feoh and @oriori1703 ,

This PR is very simple to fix the issue I've caused. It's been more than two weeks since the PR came in. Could you merge this or is there any process to merge?

I was just hoping my mess would be cleared away. 😊

rivenirvana:
Additional note to maintainers, this makes it so that the documentation right above it does not seem to be immediately related or applicable, as it's now a table and not just a link anymore. Let me know if it needs to be adjusted as well.

oriori1703:
> Hi @feoh and @oriori1703 ,
> 
> This PR is very simple to fix the issue I've caused. It's been more than two weeks since the PR came in. Could you merge this or is there any process to merge?
> 
> I was just hoping my mess would be cleared away. 😊

I wish I could merge this, but I'm not a maintainer 😅

elijaholmos (approved):
I had auto-indent issues after pulling in the latest version of `master`. This change fixed those issues for me. Let's get it merged in!

CapSel:
This change does not fix the issue for me - in C++ files I have indents of size 2spaces. Adding a line before or after an indented line adds 2tabs.
adding
```
opt = {},
config = function()
  require('guess-indent').setup{}
end,
```
helps.

guru245:
Hello @CapSel 

For that issue, I think you should report it to https://github.com/NMAC427/guess-indent.nvim as the issue has nothing to do with this PR.

Crazy-Hopper:
> This change does not fix the issue for me - in C++ files I have indents of size 2spaces. Adding a line before or after an indented line adds 2tabs. adding
> 
> ```
> opt = {},
> config = function()
>   require('guess-indent').setup{}
> end,
> ```
> 
> helps.

@CapSel maybe because you have opt instead of opts ?

CapSel:
> @CapSel maybe because you have opt instead of opts ?

@Crazy-Hopper 
yes, thank you for pointing that out, that fixed my problem :D

Lucas256-cmd:
Lol I struggled to get it working and eventually found what was wrong myself. Also about to create a new issue but then found this, it would indeed be good to merge this pls. Or maybe another idea: you could hint to new users there is something wrong with the config somewhere and they have to find it. That way nothing in the config has to change but users are at least aware there is some small error.

Lucas256-cmd:
Lol I struggled to get it working and eventually found what was wrong myself. Also about to create a new issue but then found this, it would indeed be good to merge this pls. Or maybe another idea: you could hint to new users there is something wrong with the config somewhere and they have to find it. That way nothing in the config has to change but users are at least aware there is some small error.

nikagar4:
Why is this not being merged?

szechp:
> Why is this not being merged?

the repo is dead, the contributors are burned out and we cant get the maintainer to notice us to take it over.

https://github.com/nvim-lua/kickstart.nvim/pull/1627#issuecomment-3283511227
https://github.com/nvim-lua/kickstart.nvim/pull/1627#issuecomment-3123542324

ChillerDragon:
I can confirm this works. Nice pr, fixes a bug.

ChillerDragon (approved):

harvzor (approved):

tjdevries:
Sorry for delay, merged loading this in separately

PR #1533, opened 2025-05-13
feat: add range formatting to formatter

guru245:
Range formatting in Conform is a standout feature that many other formatters lack. However, it was previously missing from kickstart. I’ve now added it, based on the guidance from [this section of the conform.nvim documentation](https://github.com/stevearc/conform.nvim/blob/master/doc/recipes.md#format-command).

This change is fully backward-compatible. Existing users can continue to use `<leader>k` to format the entire file as before. If they select a block of code and press `<leader>k`, only the selected range will be formatted.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -740,16 +740,6 @@ require('lazy').setup({
     'stevearc/conform.nvim',
     event = { 'BufWritePre' },
     cmd = { 'ConformInfo' },
-    keys = {
-      {
-        '<leader>f',
-        function()
-          require('conform').format { async = true, lsp_format = 'fallback' }
-        end,
-        mode = '',
-        desc = '[F]ormat buffer',
-      },
-    },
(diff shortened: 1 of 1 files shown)


szechp:
I've been using this for the last couple of days, and I'm running in all sort of issues when I use this on visual selection: 

1. it obviously doesn't have the context of the whole file, so indentation on Lua files is all sorts of broken.
2. Weird behavior where variables or names of functions get converted to lowercase.

I reverted it back to the original behavior.

guru245:
@szechp Thank you for the feedback and testing.

As a C programmer, I only tested it in C/C++, which was working. But as you said, I also found it's not working in lua. I will close this PR and report the issue to the conform owner. When it's resolved, I will get back and reopen this.

Sorry for the trouble I've caused.

PR #1534, opened 2025-05-13
feat: Enhances breakpoint editing

blehrer:
The keymapping `<leader>B` (in `kickstart.plugins.debug`) is now configured to guide users through the process of adding a `condition`, `hitCondition`, and `logMessage` to a breakpoint.

This also allows attributes that have already been set to be edited, instead of having to start from scratch each time you want to update one of these fields.

These breakpoints will persist for the duration of the session.

Tested with  v0.11.1

### DIFF
--- a/lua/kickstart/plugins/debug.lua
+++ b/lua/kickstart/plugins/debug.lua
@@ -85,6 +85,10 @@ return {
             prompt = 'Edit Breakpoint',
             format_item = function(item) return ('%s: %s'):format(item, props[item].value) end,
           }, function(choice)
+            if choice == nil then
+              -- User cancelled the selection
+              return
+            end
             props[prompt].setter(vim.fn.input {
               prompt = ('[%s] '):format(prompt),
               default = props[prompt].value,
--- a/lua/kickstart/plugins/debug.lua
+++ b/lua/kickstart/plugins/debug.lua
@@ -36,8 +36,9 @@ return {
     {
       '<leader>B',
       function()
+        require 'dap.protocol'
         local dap = require 'dap'
-        -- Search for an existing breakpoing on this line in this buffer
+        -- Search for an existing breakpoint on this line in this buffer
         ---@return dap.SourceBreakpoint bp that was either found, or an empty placeholder
         local function find_bp()
           local buf_bps = require('dap.breakpoints').get(vim.fn.bufnr())[vim.fn.bufnr()]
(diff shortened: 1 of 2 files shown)


blehrer:
small note to the reviewers: `h:dap.toggle_breakpoint()` misspells variable names. Compare [docs][docs] to [src][src].

[docs]: https://github.com/mfussenegger/nvim-dap/blob/8df427aeba0a06c6577dc3ab82de3076964e3b8d/doc/dap.txt#L795-L809
[src]: https://github.com/mfussenegger/nvim-dap/blob/8df427aeba0a06c6577dc3ab82de3076964e3b8d/lua/dap/protocol.lua#L295-L301

lua/kickstart/plugins/debug.lua
```diff
@@ -64,9 +64,67 @@ return {
     {
       '<leader>B',
       function()
-        require('dap').set_breakpoint(vim.fn.input 'Breakpoint condition: ')
+        require 'dap.protocol'
```
oriori1703: Why is this needed?

lua/kickstart/plugins/debug.lua
```diff
@@ -64,9 +64,67 @@ return {
...
+          ---@type dap.SourceBreakpoint
+          local bp = { condition = '', logMessage = '', hitCondition = '', line = vim.fn.line '.' }
+          for _, candidate in ipairs(buf_bps) do
+            if candidate.line and candidate.line == vim.fn.line '.' then
+              bp = candidate
+              break
+            end
+          end
+          return bp
+        end
```
oriori1703, line 51: This simplifies the code a bit
```suggestion
        local dap = require 'dap'

        -- Search for an existing breakpoint on this line in this buffer
        ---@return dap.SourceBreakpoint bp that was either found, or an empty placeholder
        local function find_bp()
          local buf_bps = require('dap.breakpoints').get(vim.fn.bufnr())[vim.fn.bufnr()]
          ---@type dap.SourceBreakpoint
          for _, candidate in ipairs(buf_bps) do
            if candidate.line and candidate.line == vim.fn.line '.' then
              return candidate
            end
          end

          return { condition = '', logMessage = '', hitCondition = '', line = vim.fn.line '.' }
        end

```

lua/kickstart/plugins/debug.lua
```diff
@@ -64,9 +64,67 @@ return {
...
+              end,
+            },
+          }
+          local menu_options = {}
+          for k, v in pairs(props) do
+            table.insert(menu_options, ('%s: %s'):format(k, v.value))
+          end
+          vim.ui.select(menu_options, {
+            prompt = 'Edit Breakpoint',
+          }, function(choice)
```
oriori1703, line 77: This should solve an error that would happen if a user cancelled the dialog 

```suggestion
          }, function(choice)
            if choice == nil then
              -- User cancelled the selection
              return
            end
```

lua/kickstart/plugins/debug.lua
```diff
@@ -64,9 +64,67 @@ return {
...
+            local prompt = (tostring(choice)):gsub(':.*', '')
```
oriori1703: for a cleaner implementation, you should use the `format_item` opt.
This way you don't need to remove the formatting from the user's choice.

i.e. something like this:
```suggestion
          for k, _ in pairs(props) do
            table.insert(menu_options, k)
          end
          vim.ui.select(menu_options, {
            prompt = 'Edit Breakpoint',
            format_item = function(item) return ('%s: %s'):format(item, props[item].value) end,
          }, function(choice)
```

oriori1703 (changes requested):
This looks a like a great addition 😄 

I commented about a few suggestions / fixes

lua/kickstart/plugins/debug.lua
blehrer: It is where `dap.SourceBreakpoint` is declared. This is used to describe the return type of `find_bp()`. It might be more idiomatic to use a `@module` ... but also, it seems like my lsp can find the definition of this type without doing so. I will remove it.

lua/kickstart/plugins/debug.lua
```diff
@@ -64,9 +64,72 @@ return {
...
+              return ('%s: %s'):format(item, props[item].value)
+            end,
+          }, function(choice)
+            if choice == nil then
+              -- User cancelled the selection
+              return
+            end
+            props[prompt].setter(vim.fn.input {
+              prompt = ('[%s] '):format(prompt),
+              default = props[prompt].value,
```
oriori1703: After the PR changes `prompt` is not defined anymore.
This also happened to me when I adapted your changes to my fork 😆 

```suggestion
              props[choice].setter(vim.fn.input {
              prompt = ('[%s] '):format(choice),
              default = props[choice].value,

```

lua/kickstart/plugins/debug.lua
oriori1703: I noticed the same :)

lua/kickstart/plugins/debug.lua
blehrer: i ran into the same thing.... currently working on a fix...

blehrer: or... i'll just merge yours ;)

oriori1703 (approved):
LGTM 😄

PR #1550, opened 2025-05-18
fix: remove unnecessary diagnostic config lines

ronandalton:
The following lines do nothing and can be removed:

```lua
 format = function(diagnostic)
   local diagnostic_message = {
     [vim.diagnostic.severity.ERROR] = diagnostic.message,
     [vim.diagnostic.severity.WARN] = diagnostic.message,
     [vim.diagnostic.severity.INFO] = diagnostic.message,
     [vim.diagnostic.severity.HINT] = diagnostic.message,
   }
   return diagnostic_message[diagnostic.severity]
 end,
```

The following line also serves little purpose, seemingly only reducing the spacing between the end of lines and diagnostic virtual text by two characters, so it can also be removed (EDIT: this change was omitted in the final patch):

```lua
  spacing = 2,
```

The following line causes diagnostics which aren't of severity level error to not have the specific text area underlined. This is less useful than the default of showing underlines for all diagnostic levels, so remove this line as well:

```lua
  underline = { severity = vim.diagnostic.severity.ERROR },
```

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -631,7 +631,6 @@ require('lazy').setup({
       vim.diagnostic.config {
         severity_sort = true,
         float = { border = 'rounded', source = 'if_many' },
-        underline = { severity = vim.diagnostic.severity.ERROR },
         signs = vim.g.have_nerd_font and {
           text = {
             [vim.diagnostic.severity.ERROR] = '󰅚 ',
@@ -643,15 +642,6 @@ require('lazy').setup({
         virtual_text = {
           source = 'if_many',
           spacing = 2,
-          format = function(diagnostic)
-            local diagnostic_message = {
-              [vim.diagnostic.severity.ERROR] = diagnostic.message,
-              [vim.diagnostic.severity.WARN] = diagnostic.message,
-              [vim.diagnostic.severity.INFO] = diagnostic.message,
-              [vim.diagnostic.severity.HINT] = diagnostic.message,
-            }
-            return diagnostic_message[diagnostic.severity]
-          end,
         },
       }
 


guru245:
I agree with the first one. But IMHO the second and the third one are a matter of preference and so we don't really need to change.

ronandalton:
I somewhat agree, but if it's a matter of preference then I think it shouldn't be in kickstart to begin with. There should be a good usability/QoL reason to deviate from the default nvim behavior and increase the config size, and I'm not seeing it. The underline config makes the experience *worse* than the default because now you can't tell which part of the line has the issue. The `spacing = 2` line can stay if that's what people want - I don't think that makes much difference either way.

rivenirvana:
For posterity's sake, the current default value for `spacing` is 4, if anyone wants to contest its inclusion in the config.

guru245:
@ronandalton Oh I didn't know the underline config is pointing to the exact part that has the issue. Good to know. Thanks.

The remaining part is spacing. I can see the spacing is 4 by default. But it was already set to 2 for inexplicable reason and many people are already using it and have gotten used to it I think. So I was wondering if it was worth changing it. I'm just saying my humble opinion. 😊

ronandalton:
That seems fair, since it has already been changed and returning the spacing to 4 would show less text on the screen in comparison to now, which could be seen as a regression.

I will update my patch to omit the removal of the `spacing = 2` line.

guru245 (approved):
LGTM

oriori1703 (approved):
Looks good :D

tjdevries:
Thanks, I ended up doing something similar. Appreciate it.

PR #1565, opened 2025-05-22
Fixed Guess-Indent not loading by default

Gonzako:
By default guess-indent is just downloaded and not loaded as it now requires a setup call. 
This fixes it.


Gonzako:
Oh, duplicate of  #1532 my bad.

Issue #1570, opened 2025-05-25
[Documentation] Add comments to make clear the order of vim options that are the equivalent of .vim scripts


xtbqimuxmvtx:
Total lua / nvim noob, but I just had a pain trying to install and configure github.com/vimwiki/vimwiki which I've used previously on vanilla vim. I read through the mountain of text prior to the first vim.o.* line and expected to find something about the order that variables are set; but didn't find anything.

Turns out global variables need to be declared prior to the plugins being loaded? Any time before the lazy installer is fine? I will add a pull request to the documentation over there to add "some" sort of lazy installation & configuration instructions.

I am now using and is working:

``` lua
vim.g.vimwiki_path = '~/vimwiki/'
vim.g.vimwiki_syntax = 'markdown'
vim.g.vimwiki_ext = 'md'
-- Setup lazy.nvim path
--...
    require('lazy').setup({
    'vimwiki/vimwiki'
--...
})

**Comment by rivenirvana** [comment]
Does lazy's `init` key not work? Refer to https://lazy.folke.io/ (currently not loading for me) For example:
```lua
{
  'vimwiki/vimwiki',
  init = function()
    vim.g.vimwiki_path = '~/vimwiki/'
    vim.g.vimwiki_syntax = 'markdown'
    vim.g.vimwiki_ext = 'md'
  end,
}
```

xtbqimuxmvtx:
Yeah that works. Thanks. Regardless as a non-programmer I find it particularly unusual that it matter what order a "global variable" is set.

Does my issue apply to options like:
vim.o
vim.opt
vim.api ?

rivenirvana:
It should work. The `init` key just makes sure that the function assigned to it is executed before the plugin in its scope loads, the function can do any setup/valid code of lua you want it to do.
About the global variable issue, just a quirk with using plugins written in vimscript in a lua-based plugin manager. They expect their custom configs to be set in the vimrc file, so the only way to set it when lazy code is already running is via the `init` key (its intended purpose as per the docs). Otherwise, you set it before and outside lazy's code.
Also, global does mean it is accessible everywhere, but since code is read downwards, globals are usually set above where it's expected to be used, i.e. before the entire {chunk} of lazy's code.

PR #1582, opened 2025-06-02
docs: Add documentation for lazy's `init` key

rivenirvana:
Guide users towards lazy's `init` key for plugins written in VimScript.
Alleviates issues like #1570.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -267,6 +267,10 @@ require('lazy').setup({
   --        end,
   --    }
   --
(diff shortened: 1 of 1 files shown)


oriori1703 (approved):
LGTM 😄

tjdevries:
Thanks, this is good info but I think I'll skip because for most people they are not encountering too many vimscript plugins these days and I try to keep everything as short as we can.

PR #1590, opened 2025-06-07
feat: improved lsp setup for nvim > 0.11

kartikvashistha:
The aim of this change is to provide a much clearer distinction between how servers are installed & configured (either directly via Mason or manually). This is a slight alternative to [this](https://github.com/nvim-lua/kickstart.nvim/pull/1475) excellent PR for LSP configuration for neovim > 0.11 onwards.

Since the new `mason-lspconfig` changes enable all servers installed via Mason by default, this might cause confusion for newcomers around server enable if they had a language server installed manually but wanted to re-use the default `nvim-lspconfig` configs. 
Furthermore, not all language servers might be available on Mason for download - however they might already have default configs available on `nvim-lspconfig` for re-use ([example](https://github.com/mason-org/mason-lspconfig.nvim/issues/527)) or have neither available on Mason and `nvim-lspconfig`.

I'm hoping another benefit of this change is that it _should_ also aid in reducing burden on the mason & nvim-lspconfig maintainers when it comes to answering these specific kind of lsp questions.

As always, I'm open to any feedback here. I'm not too fussed if the above linked PR gets merged over this, but wanted to float this idea by the maintainers and community to gather any & all feedback.

Cheers! 🙂

**Minor Changes:**
- Commented and preserved the capabilities step for educational purposes only;
- Auto formatter might have formatted some code, which I quite like tbh . Happy to revert it if needed.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -668,7 +668,15 @@ require('lazy').setup({
       --  1) via the mason package manager; or
       --  2) via your system's package manager; or
       --  3) via a release binary from a language server's repo that's accessible somewhere on your system.
-      --
+
+      -- The servers table comprises of the following sub-tables:
+      -- 1. mason
+      -- 2. others
+      -- Both these tables have an identical structure of language server names as keys and
+      -- a table of language server configuration as values.
+      ---@class LspServersConfig
+      ---@field mason table<string, vim.lsp.Config>
+      ---@field others table<string, vim.lsp.Config>
       local servers = {
         --  Add any additional override configuration in any of the following tables. Available keys are:
         --  - cmd (table): Override the default command used to start the server
@@ -736,7 +744,7 @@ require('lazy').setup({
       -- to the default language server configs as provided by nvim-lspconfig or
       -- define a custom server config that's unavailable on nvim-lspconfig.
       for server, config in pairs(vim.tbl_extend('keep', servers.mason, servers.others)) do
-        if vim.fn.empty(config) ~= 1 then
+        if not vim.tbl_isempty(config) then
           vim.lsp.config(server, config)
         end
       end
@@ -748,7 +756,7 @@ require('lazy').setup({
       }
 
       -- Manually run vim.lsp.enable for all language servers that are *not* installed via Mason
-      if vim.fn.empty(servers.others) ~= 1 then
+      if not vim.tbl_isempty(servers.others) then
         vim.lsp.enable(vim.tbl_keys(servers.others))
       end
     end,
--- a/init.lua
+++ b/init.lua
@@ -1,5 +1,4 @@
 --[[
-
 =====================================================================
 ==================== READ THIS BEFORE CONTINUING ====================
 =====================================================================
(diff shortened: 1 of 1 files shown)


init.lua
```diff
@@ -713,26 +726,31 @@ require('lazy').setup({
...
-            -- certain features of an LSP (for example, turning off formatting for ts_ls)
-            server.capabilities = vim.tbl_deep_extend('force', {}, capabilities, server.capabilities or {})
-            require('lspconfig')[server_name].setup(server)
-          end,
-        },
+        automatic_enable = true, -- automatically run vim.lsp.enable() for all servers that are installed via Mason
       }
+
+      -- Manually run vim.lsp.enable for all language servers that are *not* installed via Mason
+      if vim.fn.empty(servers.others) ~= 1 then
```
oriori1703:
```suggestion
      if not vim.tbl_isempty(servers.others) then
```

init.lua
```diff
@@ -713,26 +726,31 @@ require('lazy').setup({
...
       vim.list_extend(ensure_installed, {
         'stylua', -- Used to format Lua code
       })
       require('mason-tool-installer').setup { ensure_installed = ensure_installed }
 
+      -- Either merge all additional server configs from the `servers.mason` and `servers.others` tables
+      -- to the default language server configs as provided by nvim-lspconfig or
+      -- define a custom server config that's unavailable on nvim-lspconfig.
+      for server, config in pairs(vim.tbl_extend('keep', servers.mason, servers.others)) do
+        if vim.fn.empty(config) ~= 1 then
```
oriori1703:
```suggestion
        if  not vim.tbl_isempty(config) then
```

init.lua
```diff
@@ -659,45 +658,59 @@ require('lazy').setup({
...
+      --  2) via your system's package manager; or
+      --  3) via a release binary from a language server's repo that's accessible somewhere on your system.
       --
-      --  Add any additional override configuration in the following tables. Available keys are:
-      --  - cmd (table): Override the default command used to start the server
-      --  - filetypes (table): Override the default list of associated filetypes for the server
-      --  - capabilities (table): Override fields in capabilities. Can be used to disable certain LSP features.
-      --  - settings (table): Override the default settings passed when initializing the server.
-      --        For example, to see the options for `lua_ls`, you could go to: https://luals.github.io/wiki/settings/
       local servers = {
```
oriori1703, line 680:
```suggestion
      ---@class LspServersConfig
      ---@field mason table<string, vim.lsp.Config>
      ---@field others table<string, vim.lsp.Config>
      local servers = {
```
maybe you should add a type hint to the table.
This way users will get autocomplete when configuring servers.

You could also add something like this to the comment about the config keys:
> \<K\> could be used to "hover" any of the servers see more info about the available keys

oriori1703 (changes requested):
I quite like the idea of teaching how to use servers that were not installed by mason, so I think I prefer this over https://github.com/nvim-lua/kickstart.nvim/pull/1475 (even though it is also great :)

Though I think the formatting changes should be left out of this PR.
Also, I left a couple of suggestions for minor improvements.

oriori1703 (changes requested):
Can you remove the formatting changes?

Other that everything looks great :smile:

gozes:
Just an FYI, I have ported this fixes to my kickstart base config. after adding the loop to call `lsp.config` load time went to 200ms just on the lspconfig

theammir:
> Just an FYI, I have ported this fixes to my kickstart base config. after adding the loop to call `lsp.config` load time went to 200ms just on the lspconfig

It's something I'm experiencing with lots of LSPs installed via Mason with `automatic_enable = true`. More than half of my startup time goes to enable the LSPs. I wonder why is it so expensive, though.

moktavizen:
> > Just an FYI, I have ported this fixes to my kickstart base config. after adding the loop to call `lsp.config` load time went to 200ms just on the lspconfig
> 
> It's something I'm experiencing with lots of LSPs installed via Mason with `automatic_enable = true`. More than half of my startup time goes to enable the LSPs. I wonder why is it so expensive, though.

I also noticed that, right now loading LSP at `VeryLazy` event is enough to fix it.

```lua
-- Main LSP Configuration
'neovim/nvim-lspconfig',
event = 'VeryLazy',
cmd = { 'LspInfo', 'LspInstall', 'LspUninstall' },
```

theammir:
> > > Just an FYI, I have ported this fixes to my kickstart base config. after adding the loop to call `lsp.config` load time went to 200ms just on the lspconfig
> > 
> > 
> > It's something I'm experiencing with lots of LSPs installed via Mason with `automatic_enable = true`. More than half of my startup time goes to enable the LSPs. I wonder why is it so expensive, though.
> 
> I also noticed that, right now loading LSP at `VeryLazy` event is enough to fix it.
> 
> ```lua
> -- Main LSP Configuration
> 'neovim/nvim-lspconfig',
> event = 'VeryLazy',
> cmd = { 'LspInfo', 'LspInstall', 'LspUninstall' },
> ```

Works great! Do you want to open a PR maybe? Or is it a temporary solution that shouldn't really belong?

logrusx:
Overal I like the idea of separating the mason and other servers.

However this PR is dealing with mason-lspconfig 2.0 rather than nvim 0.11 and it does more than it says. 

The idea of separating mason and other servers may be a subject of another PR.

I recommend #1663 over this one as it does exactly what it says and does it correctly.

init.lua
```diff
@@ -659,45 +658,67 @@ require('lazy').setup({
       --  By default, Neovim doesn't support everything that is in the LSP specification.
       --  When you add blink.cmp, luasnip, etc. Neovim now has *more* capabilities.
       --  So, we create new capabilities with blink.cmp, and then broadcast that to the servers.
-      local capabilities = require('blink.cmp').get_lsp_capabilities()
-
-      -- Enable the following language servers
-      --  Feel free to add/remove any LSPs that you want here. They will automatically be installed.
+      -- NOTE: The following line is now commented as blink.cmp extends capabilites by default from its internal code:
```
michal-mnich, line 661: In this case, don't you need to put:
```lua
{ 'saghen/blink.cmp', opts = {} },
```
in the `dependencies` for the capabilities to get populated?

silverdyne:
> > > Just an FYI, I have ported this fixes to my kickstart base config. after adding the loop to call `lsp.config` load time went to 200ms just on the lspconfig
> > 
> > 
> > It's something I'm experiencing with lots of LSPs installed via Mason with `automatic_enable = true`. More than half of my startup time goes to enable the LSPs. I wonder why is it so expensive, though.
> 
> I also noticed that, right now loading LSP at `VeryLazy` event is enough to fix it.
> 
> ```lua
> -- Main LSP Configuration
> 'neovim/nvim-lspconfig',
> event = 'VeryLazy',
> cmd = { 'LspInfo', 'LspInstall', 'LspUninstall' },
> ```

Unfortunately, this only seems to work if the LSP servers are already installed and configured.

I tried this with a fresh `nvim` install and for some reason, no servers would be installed until I removed those two lines.

fadookie:
I manually patched this change into my modular kickstart config and it seems to fix the issue I was having with my configs for LSPs not propagating. I was also wondering how to install non-mason LSPs. This PR addresses both my questions, thank you!

moktavizen:
> Unfortunately, this only seems to work if the LSP servers are already installed and configured.
> 
> I tried this with a fresh `nvim` install and for some reason, no servers would be installed until I removed those two lines.

Yes, it's a known issue https://github.com/WhoIsSethDaniel/mason-tool-installer.nvim/issues/39

The workaround is 

```lua
require('mason-tool-installer').setup { ensure_installed = ensure_installed }

vim.api.nvim_command('MasonToolsInstall')
```

inlineflow:
This helped me fix my issues with LSP configs not being ran. Hope this gets merged, so that fresh people don't have to suffer and flail around as I did.

tjdevries:
I think i merged something that should do similar thing. Thanks for the PR!

PR #1598, opened 2025-06-12
fix: Actually start up guess-indent

StandingPadAnimations:
Guess-indent requires setup to be called in order to function properly.


bewuethr:
See also #1532

tjdevries:
Sorry for delay, loaded it now

PR #1605, opened 2025-06-16
feat(grep-string): works with visual selection too

FrancescElies:
telescope's `grep_string` works in visual mode too.

**NOTE**: If I missed the reason why this was not added feel free to directly close this pr


oriori1703 (approved):
I don't see any reason not to include this 😄

tjdevries:
Thanks!

PR #1616, opened 2025-06-24
feat: add toggle to show/hide diagnostics

guru245:
Sometimes, diagnostic messages could be irritating. So it wouldn't hurt to have a toggle to show/hide diagnostics.


dam9000 (approved):
I agree with this, I have the same keymap in my config. :)

oriori1703 (approved):
I now stole this for my config 😄

tjdevries:
I think for now I'll say no, but i will think about this a bit more.

PR #1619, opened 2025-06-26
fix: update mason-lspconfig setup to not use deprecated handlers

moguls753:
Since the `handlers` setting is removed in [mason-lspconfig v2.0](https://github.com/mason-org/mason-lspconfig.nvim/releases/tag/v2.0.0), I think the following makes sense.

- Remove deprecated `handlers` API and `automatic_installation`

- Set `automatic_enable = false` since setup() is invoked explicitly

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -721,18 +721,17 @@ require('lazy').setup({
 
       require('mason-lspconfig').setup {
         ensure_installed = {}, -- explicitly set to an empty table (Kickstart populates installs via mason-tool-installer)
-        automatic_installation = false,
-        handlers = {
-          function(server_name)
-            local server = servers[server_name] or {}
-            -- This handles overriding only values explicitly passed
-            -- by the server configuration above. Useful when disabling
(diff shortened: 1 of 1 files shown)


guru245:
Hi @moguls753 
Would you care to see #1475 ? Does it include your change?

moguls753:
Hey @guru245,

I totally missed #1475—thanks for pointing it out! That PR switches over to the new Neovim ≥0.11 LSP API, while this one sticks with the classic

```lua
require('lspconfig').<server>.setup{…}
```

Given that many distros (for example, Ubuntu 25 LTS ships 0.9.5) still ship older Neovims, I think it makes sense to keep using the legacy API, which isn’t going away anytime soon. For me #1475 is working perfectly fine though.

What do you think?

VlaDexa:
I think that kickstart supports only the latest stable, and should stick with that

moguls753:
In that case, #1475 seems perfectly fine — I’ll close this PR. Thanks for the pointer!

PR #1621, opened 2025-06-30
fix: guess-indent plugin not working

gartmeier:
The guess-indent plugin wasn't doing anything after cloning kickstart.nvim - no auto-detection and the `:GuessIndent` command didn't exist.

Turns out the plugin needs `opts = {}` in the lazy config to actually run its setup function. Without it, lazy doesn't bother calling setup, so you get nothing.

Simple fix:
```lua
{ 'NMAC427/guess-indent.nvim', opts = {} }, -- Detect tabstop and shiftwidth automatically
```

Now it actually works like it should.


gartmeier:
Sorry, just noticed there are already existing PRs for this same issue: #1532 and #1598.

Closing this one. Thanks for maintaining the project!

PR #699, opened 2024-03-05
Add plugin for linting

fredrikaverpil:
### Why this change?

- There is no linter included in kickstart.
- Fixes #698

### What was changed?

- Add https://github.com/mfussenegger/nvim-lint as the choice of linter. This is a popular linter that is the default linter in nvim distros such as LazyVim.

PR #1624, opened 2025-07-02
add markdownlint to Mason auto-install

### LINKED ISSUE #1510: 'mfussenegger/nvim-lint' causes BufEnter errors with telescope
## Before Reporting an Issue

- [X] **By checking this, I confirm that the above steps are completed. I understand leaving this unchecked will result in this report being closed immediately.**

## Describe the bug
You can see a sample of the error message here: https://github.com/nvim-telescope/telescope.nvim/issues/3129#issue-2307528065

Mixing nvim-lint (from file https://github.com/nvim-lua/kickstart.nvim/blob/master/lua/kickstart/plugins/lint.lua) with telescope causes BufEnter errors.

This causes problems with files and caused me to lose a file through the buffer.

It seems to be fixed by disabling certain filetypes from linting, such as markdown (https://github.com/nvim-telescope/telescope.nvim/issues/3129#issuecomment-2853741859)

## To Reproduce
1. Enable mfussenegger/nvim-lint (enabled by uncommeting "kickstart.plugins.lint")
2. Have a file in the buffer

## Desktop
- OS: Both Manjaro & Arch
- Terminal: foot

## Neovim Version

```
NVIM v0.11.1
Build type: RelWithDebInfo
LuaJIT 2.1.1741730670
```

batoorsayed:
Fixes #1510  

[lua/kickstart/plugins/lint.lua](https://github.com/nvim-lua/kickstart.nvim/blob/3338d3920620861f8313a2745fd5d2be39f39534/lua/kickstart/plugins/lint.lua#L9) adds `markdownlint` as a linter. But it was not added to list of tools that Mason installed.

This adds markdownlint under `ensure_installed`, which should solve the following error:
```
E5108: Error executing lua: ...hare/nvim/lazy/neo-tree.nvim/lua/neo-tree/utils/init.lua:799: BufEnter Autocommands for "*": Vim(append):Error running markdownlint: ENOENT: no such file or directory
```

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -716,6 +716,7 @@ require('lazy').setup({
       local ensure_installed = vim.tbl_keys(servers or {})
       vim.list_extend(ensure_installed, {
         'stylua', -- Used to format Lua code
+        'markdownlint', -- Markdown linter
       })
       require('mason-tool-installer').setup { ensure_installed = ensure_installed }
 


batoorsayed:
As a side comment, I'd rather comment out [markdownlint](https://github.com/nvim-lua/kickstart.nvim/blob/3338d3920620861f8313a2745fd5d2be39f39534/lua/kickstart/plugins/lint.lua#L9). Having a linter for .md is rather quite annoying.

init.lua
```diff
@@ -716,6 +716,7 @@ require('lazy').setup({
       local ensure_installed = vim.tbl_keys(servers or {})
       vim.list_extend(ensure_installed, {
         'stylua', -- Used to format Lua code
+        'markdownlint', -- Markdown linter
```
guru245, line 719: I suggest commenting out `'markdownlint',` as ` require 'kickstart.plugins.lint',` in init.lua is also commented out.

batoorsayed: That's a good point. 
User must be given clear instructions in that case, otherwise it will leave the point of friction lingering for some:

Enabling linting by uncommenting `require 'kickstart.plugins.lint'` in `init.lua`  will also enable `markdownlint` in `lint.lua`. 
_Unless_ user explicitly disables `markdownlint` in `lint.lua`, they _have to_  also ensure that thay have installed `markdownlint` to prevent any issues. 
Which can be done by uncommenting `markdownlint`  in `'mason-tool-installer'` setup.

guru245: IMHO, kickstart is not a distribution but starting point. Therefore, kickstart doesn't have to be seamless. The users must experience this kind of small, tiny issue. And your PR provides a good pointer to solve it. If they see your PR, they will understand how kickstart works. If they don't have any issue when enabling `require 'kickstar.plugin.lint'`, they won't understand how to add the other lint. 

I think we need to add `--markdownlint,` along with comments saying where to check to enable a lint. This is just my opinion. Let's hear from the others.

rivenirvana: I suggest just documenting that if you enable kickstart's lint plugin, you also have to uncomment markdownlint to be auto-installed by mason if the user does not want to see the error. Regardless if kickstart is supposed to be seamless or not, having to stumble upon a PR to find the solution to an issue is just a poor experience. kickstart's big thing is about docs, so let's rely on that.

Or maybe this also puts into question markdownlint being a dependency of the lint plugin's default implementation.

batoorsayed: I'm actually 100% on board with removing `.md` as a default dependency. `markdownlint` also requires `npm` as a dependency when installing through Mason. I'm pretty sure majority of users will not see the bug, I was working on a clean MacOS install so I didn't have npm in my system.

TL;DR : It just seems like having  `.md` as a default is more trouble than what it's worth. Any reason why it was implemented in first place?

rivenirvana: I guess it was like that when it was merged.
#699 
https://github.com/nvim-lua/kickstart.nvim/commit/8e24ca32e32b866a0bec443b937fe75bd5756fc8

batoorsayed: hah, thanks. Yea, commenting out `markdownlint` seems like the most optimal choice imo. We can definitely provide further instructions to clarify  things if needed.

tjdevries:
No thanks, appreciate it tho!

PR #1628, opened 2025-07-04
feat(diagnostics): enhance diagnostics display, dynamic virtual text like VS Code Error Lens

glaulher:
Hi there! 👋

This PR introduces a small enhancement to the diagnostic experience in Neovim.

Instead of keeping the diagnostics always in virtual text or virtual lines, this new setup dynamically switches between the two based on the current mode:

Insert mode: diagnostics are shown inline using virtual text — similar to the behavior of the Error Lens extension in VS Code.

Normal mode: diagnostics switch to virtual lines below the affected code, offering a cleaner and more elegant overview.

This approach provides real-time feedback while coding and a less intrusive display when navigating the code. It's a feature that requires a plugin in editors like VS Code, but here it’s done natively with just a bit of Lua! 😄

I believe this improves the overall developer experience by making diagnostics both more immediate and more readable.

Let me know what you think — open to suggestions!

Thanks 🙏

PR #1629, opened 2025-07-05
docs(readme): add plugin configuration index with line references

glaulher:
Hi! 👋

This PR adds a new ## Configuration section to the README.md that provides quick access to plugin configuration.

The goal is to help users find exactly where each plugin is configured — whether directly in init.lua or in a dedicated file under lua/kickstart/plugins/.

This can be especially useful for:

- New users trying to understand how each plugin is integrated

- Users customizing their own config and needing a quick reference

- Reducing friction when searching through the file

Line references for plugins configured in init.lua have been included for convenience.

Hope this small addition improves usability! Let me know if you'd like adjustments.

Thanks for the great project! 🙏

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,35 @@ examples of adding popularly requested plugins.
     * [Restructure the configuration](https://github.com/nvim-lua/kickstart.nvim/issues/218)
     * [Reorganize init.lua into a multi-file setup](https://github.com/nvim-lua/kickstart.nvim/pull/473)
 
+
+## Configuration
+
+This section provides quick links to the configuration of each plugin.
+
+### Plugins configured in `init.lua`
+
+- [guess-indent.nvim](init.lua#L250)
+- [gitsigns.nvim](init.lua#L275)
+- [which-key.nvim](init.lua#L302)
+- [telescope.nvim](init.lua#L362)
+- [lazydev.nvim](init.lua#L469)
+- [nvim-lspconfig](init.lua#L480)
+- [conform.nvim](init.lua#L762)
+- [blink.cmp](init.lua#L803)
+- [tokyonight.nvim](init.lua#L906)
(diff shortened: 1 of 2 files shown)


glaulher:
Note: This PR also includes the changes proposed in #1628.

If the diagnostic enhancements from that PR are not wanted here, I'm happy to adjust the branch and submit only the relevant changes for this one. Just let me know!

oriori1703:
I do like the idea, but wouldn't that be a pain to update each time a line is added or removed above the linked lines?
Maybe we could just have an easily searchable banner for each plugin, and document that?

Or maybe just greping (or search using `/`) the plugin name is good enough?

rivenirvana:
I think it would be more worthwhile to reevaluate #473 instead, which would also nudge new users towards learning how to interact with and navigate between multiple files in neovim, via the defaults it has (commands to buffers, splits & navigation, etc), or via the configuration kickstart provides (telescope, search keymaps, etc).

guru245:
image
A friend of mine is managing init.lua like this. 😁Then, we don't have to go through the pain to update comments.

When I want to know what plugin I use, I execute `:lazy home`. If I want to go to the plugin setting, I run `<leader>sg` or `<leader>s/`.

glaulher:
Thanks everyone for the thoughtful feedback!

I’m currently using the modular version of Kickstart by dam9000 (as mentioned in #473), and my idea with this PR was simply to make it easier for new users of Kickstart to locate the exact lines where each plugin is configured — whether in `init.lua` or in separate files.

That said, I completely agree that maintaining line references manually would introduce unnecessary overhead and could quickly become unmanageable with frequent updates to the codebase.

I’ll go ahead and close this PR — thank you again for the helpful discussion and for all the work on this great project! 🙏

PR #1631, opened 2025-07-07
Raise the minimum nvim version to 0.11

oriori1703:
Raise the minimum version because kickstart only supports the latest stable and nightly versions.


guru245:
I love this PR. I thought the same thing.

slg95:
I was having trouble with Neovim `0.10.2` and the current stable is `0.11.2`, so I think this makes sense to do

tjdevries:
Thanks, ended up merging this basically over two prs. good call

PR #1481, opened 2025-04-18
Feature/snacks picker

oriori1703:
Switch from `telescope` to `snacks.nvim`. 

Benefits of snacks compiled from the comments:

- It has more pickers out of the box
- it supports frequency sorting
- it is faster (according to some people, but I don't feel the difference personally)
- it is better maintained
- it is integrated with a lot of other plugins (but so is telescope)
- it is as easy to configure as telescope
- it might make the transition to rolling your own config using kickstart, if you are coming from an nvim distribution that already uses snacks (which I believe all the big ones do)
- and ... it also a bit nicer (but that is subjective)

I just switched my config to use snacks instead of telescope, so I'm opening a PR in case it interests anyone else.

I know that this might be controversial because both Kickstart and telescope were create by teej  (which did an awesome job on both BTW), so feel free to close this if you feel it doesn't suite the project.

Also, if you want, we can also easily use `snacks.explorer` and `snacks.indent` to replace `neo-tree` and `indent-blanklines`

PR #1632, opened 2025-07-10
Optional snacks

szechp:
as discussed here: https://github.com/nvim-lua/kickstart.nvim/pull/1481 i have implemented snacks picker as an optional replacement for telescope. I hope that we can use this implementation to add more optional plugins, so we don't waste time on deciding if we want whatever new shiny plugin needs to be added or not.

Now we can provide a stable base, but we can also add optional plugins and users can decide themselves.

(thanks to @oriori1703, i just copied his snacks config from the other mr)

### DIFF
--- a/lua/kickstart/plugins/snacks.lua
+++ b/lua/kickstart/plugins/snacks.lua
@@ -1,0 +1,154 @@
+-- NOTE: Plugins can specify dependencies.
+--
+-- The dependencies are proper plugin specifications as well - anything
+-- you do for a plugin at the top level, you can do for a dependency.
+--
+-- Use the `dependencies` key to specify the dependencies of a particular plugin
+
+return { -- Fuzzy Finder (files, lsp, etc)
+  'folke/snacks.nvim',
+  priority = 1000,
+  lazy = false,
+  dependencies = {
+    -- Useful for getting pretty icons, but requires a Nerd Font.
+    { 'nvim-tree/nvim-web-devicons', enabled = vim.g.have_nerd_font },
+  },
+
+  -- snacks.nvim is a plugin that contains a collection of QoL improvements.
+  -- One of those plugins is called snacks-picker
+  -- It is a fuzzy finder, inspired by Telescope, that comes with a lot of different
+  -- things that it can fuzzy find! It's more than just a "file finder", it can search
+  -- many different aspects of Neovim, your workspace, LSP, and more!
+  --
+  -- Two important keymaps to use while in a picker are:
+  --  - Insert mode: <c-/>
+  --  - Normal mode: ?
+  --
+  -- This opens a window that shows you all of the keymaps for the current
+  -- Snacks picker. This is really useful to discover what nacks-picker can
+  -- do as well as how to actually do it!
+
+  -- [[ Configure Snacks Pickers ]]
+  -- See `:help snacks-picker` and `:help snacks-picker-setup`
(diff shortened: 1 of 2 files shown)


dam9000:
nicely done!

perhaps some commits could be squashed, these two together:

add vim.g toggle to disable telescope
remove vim.g option again, use lazyvims enable functionality

and these two as well:

add snacks(picker) as an optional plugin
add some comments to snacks picker to get it inline with telescope config

which would end up in 3 commits, easier to follow changes.

szechp:
Ah sure, I was thinking commits will be squashed on merge anyway, so I was a little messy. I'll fix that

oriori1703:
This looks great, but as I said in https://github.com/nvim-lua/kickstart.nvim/pull/1481#issuecomment-3062709770, I do think it would be better to just have one picker.

Though, I think it would be beneficial to have the autocmds defined in the plugin, even if in the end only one of the pickers would be chosen,  because it would be easier to switch to the other :smile:

szechp:
I should split this into two prs then

One for the telescope change and one for snacks, so we can better decide on that

lua/kickstart/plugins/snacks.lua
```diff
@@ -0,0 +1,154 @@
...
+      desc = '[S]earch [N]eovim files',
+    },
+  },
+
+  -- This runs on LSP attach per buffer (see main LSP attach function in 'neovim/nvim-lspconfig' config for more info,
+  -- it is better explained there). This is a little bit redundant, but we can switch off pickers for an optional
+  -- picker like this one here more easily when the keymaps are defined in the plugin itself.
+  -- It sets up buffer-local keymaps, autocommands, and other LSP-related settings
+  -- whenever an LSP client attaches to a buffer.
+  config = function()
```
Jacobus-afk, line 140: Just a question, should this not be `init` instead of `config`?

szechp: > Just a question, should this not be `init` instead of `config`?

This used to be together with this:

https://github.com/nvim-lua/kickstart.nvim/pull/1640

I separated them into two prs. This one will likely not be merged, at least according to what I feel is the current consensus regarding kickstart. 

If you ever want to replace telescope, you can always take this once the smaller change is merged

Jacobus-afk: Thanks @szechp, i did replace telescope with snacks, but it looks like using `config` as is will  override any `opts` being set.. only thing that worked for me was to replace it with `init`

tjdevries:
This is a nice PR - i'm not sure what to do with it though. Thinking most likely would be to close because I don't want to put a ton of alternatives for every single plugin. Honestly (even tho obviously i love telescope since it's my baby haha) i am thinking about putting one of the other plugins here because for me telescope is basically finished software (does everything i like) but I know other fuzzy finders are pushing new things forward

szechp:
That's ok :) I guess everyone that's starts with this repo should be able to exchange their picker. But It would be nice to get the telescope calls outside of the plugin Lua back inside the file. I'll do another Mr with that

tjdevries:
Ya, I think I merged your prev request that moved all the telescope config into the telescope plugin. So it should be very easy to disable now and switch to new picker for people

szechp:
Ah I already did that, lol

tjdevries:
haha well it was only six months ago lololol

PR #1640, opened 2025-07-13
Telescope config in plugin

szechp:
I previously added these changes to another pr of mine, but after some thought, im pulling those changes apart, so we can decide on replacing the picker, more infos here: https://github.com/nvim-lua/kickstart.nvim/pull/1632

this pr just moves some telescope specific config from the lspconfig to the telescope plugin config declaration, just to entangle those two.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -360,14 +360,15 @@ require('lazy').setup({
 
   { -- Fuzzy Finder (files, lsp, etc)
     'nvim-telescope/telescope.nvim',
-    -- To disable the default Telescope plugin and replace it with
-    -- another picker (like snacks), set enabled to false and
-    -- Enable your replacement picker by requiring it explicitly (e.g., 'kickstart.plugins.snacks')
-    -- By default, Telescope is included and acts as your picker for everything.
+   -- By default, Telescope is included and acts as your picker for everything.
 
-    -- Note: When you customize your config for yourself,
-    -- it’s best to remove the Telescope plugin config entirely
-    -- instead of just disabling it here, to keep your config clean.
+   -- If you would like to switch to a different picker (like snacks, or fzf-lua)
+   -- you can disable the Telescope plugin by setting enabled to false and enable
+   -- your replacement picker by requiring it explicitly (e.g. 'custom.plugins.snacks')
+
+   -- Note: If you customize your config for yourself,
+   -- it’s best to remove the Telescope plugin config entirely
+   -- instead of just disabling it here, to keep your config clean.
     enabled = true,
     event = 'VimEnter',
     dependencies = {
--- a/init.lua
+++ b/init.lua
@@ -360,15 +360,15 @@ require('lazy').setup({
 
   { -- Fuzzy Finder (files, lsp, etc)
     'nvim-telescope/telescope.nvim',
-   -- By default, Telescope is included and acts as your picker for everything.
+    -- By default, Telescope is included and acts as your picker for everything.
 
-   -- If you would like to switch to a different picker (like snacks, or fzf-lua)
-   -- you can disable the Telescope plugin by setting enabled to false and enable
-   -- your replacement picker by requiring it explicitly (e.g. 'custom.plugins.snacks')
+    -- If you would like to switch to a different picker (like snacks, or fzf-lua)
+    -- you can disable the Telescope plugin by setting enabled to false and enable
+    -- your replacement picker by requiring it explicitly (e.g. 'custom.plugins.snacks')
 
-   -- Note: If you customize your config for yourself,
-   -- it’s best to remove the Telescope plugin config entirely
-   -- instead of just disabling it here, to keep your config clean.
+    -- Note: If you customize your config for yourself,
+    -- it’s best to remove the Telescope plugin config entirely
+    -- instead of just disabling it here, to keep your config clean.
     enabled = true,
     event = 'VimEnter',
     dependencies = {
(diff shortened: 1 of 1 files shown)


gmnz:
I think moving telescope keymaps out of lspconfig and into telescope's config is a good idea, but I still disagree with providing alternative picker in "kickstart.plugins.snacks" or anywhere else in the code. That just opens the door for a lot of other plugins to be included (alternative file browser plugins, alternative autopairs plugins etc.) which is completely unneeded. Kickstart shouldn't become a catalogue of plugins, it should stay small and reasonably usable. Every new nvim user will find out on his own soon after starting to use nvim that there's a ton of other plugins which he can use instead of the default ones chosen by kickstart if he's at least a bit interested in how nvim and the community around it works. If he's not interested, he should use LazyVim or VSCode or whatever.

szechp:
I mean it's not in the code. It's a comment. It's there to encourage people to make it their own.

gmnz:
No, it encourages them to use snacks because it mentions snacks. If you want to encourage people to make the config their own, you shouldn't mention specific plugins, just add some general comment like "There are many other nvim plugins which add additional functionality to nvim or alternatives to plugins already included in kickstart". Everybody can type into google "nvim telescope alternative" and find out about snacks or fzf or any other alternative this way and kickstart's code and comments stay much cleaner this way and people don't have to fight about which plugin should be mentioned in the comments and which not, because all of them can't be mentioned, so who'll decide which will be? It's better to not mention any other plugins except those that are included in kickstart. There's already enough fighting about the default plugins included in kickstart, no need for even more fighting about plugins included in comments.

szechp:
yeah theres a bunch of plugins in there at the moment: https://github.com/nvim-lua/kickstart.nvim/blob/master/init.lua#L976

so if you want them removed, make a pr yourself, we can discuss that there. Until then im gonna leave this here, and maybe a maintainer has some thoughts on that and can chime in.

gmnz:
> yeah theres a bunch of plugins in there at the moment: https://github.com/nvim-lua/kickstart.nvim/blob/master/init.lua#L976
> 
> so if you want them removed, make a pr yourself, we can discuss that there. Until then im gonna leave this here, and maybe a maintainer has some thoughts on that and can chime in.

But those aren't alternatives to a plugin that's already included in kickstart but plugins adding completely new features.

init.lua
```diff
@@ -360,6 +360,15 @@ require('lazy').setup({
...
   { -- Fuzzy Finder (files, lsp, etc)
     'nvim-telescope/telescope.nvim',
+    -- To disable the default Telescope plugin and replace it with
+    -- another picker (like snacks), set enabled to false and
+    -- Enable your replacement picker by requiring it explicitly (e.g., 'kickstart.plugins.snacks')
+    -- By default, Telescope is included and acts as your picker for everything.
+
+    -- Note: When you customize your config for yourself,
+    -- it’s best to remove the Telescope plugin config entirely
+    -- instead of just disabling it here, to keep your config clean.
```
oriori1703: I do agree it might encourage snacks a bit too much.\
what do you think about this?
 ```suggestion
    -- By default, Telescope is included and acts as your picker for everything.

    -- If you would like to switch to a different picker (like snacks, or fzf-lua)
    -- you can disable the Telescope plugin by setting enabled to false and enable
    -- your replacement picker by requiring it explicitly (e.g. 'custom.plugins.snacks')

    -- Note: If you customize your config for yourself,
    -- it’s best to remove the Telescope plugin config entirely
    -- instead of just disabling it here, to keep your config clean.
```

Comment (rivenirvana) [reply init.lua]: For line 363-364, I think removing code blocks for plugin configs should be more encouraged when replacing same feature plugins, instead of recommending toggling an `enabled` key and keeping stale configs. In fact, this is what the note in line 368 says, and if the configs are already laid out intuitively (i.e. features cleanly separated), then that should also be unneeded.

For line 365, there's already documentation for this in line 983 onwards in current `init.lua`, specifically about importing custom.plugins, and `custom/plugins/init.lua`.

Line 366 can be shortened and merged instead with the top description in line 361.

File: init.lua [hunk 2]
```diff
@@ -437,6 +446,44 @@ require('lazy').setup({
       vim.keymap.set('n', '<leader>s.', builtin.oldfiles, { desc = '[S]earch Recent Files ("." for repeat)' })
       vim.keymap.set('n', '<leader><leader>', builtin.buffers, { desc = '[ ] Find existing buffers' })
 
+      -- This runs on LSP attach per buffer (see main LSP attach function in 'neovim/nvim-lspconfig' config for more info,
+      -- it is better explained there). This is a little bit redundant, but we can switch off telescope for an optional
+      -- picker like snacks more easily when the keymaps are defined in the plugin itself.
+      -- It sets up buffer-local keymaps, autocommands, and other LSP-related settings
+      -- whenever an LSP client attaches to a buffer.
```
rivenirvana, line 454: As with my other comment above, I think it should be removed if it's redundant and the configs are intuitive (allows for easy plugin swapping). And the reasoning "when the keymaps are defined in the plugin itself" is more irrelevant to the user as it is more about a PR description.

Line 452-453 can be shortened to describe 455, but it should just be removed as well as this is redundant to what is noted anyway in parentheses.

szechp: i dont know know about you, but when i see the telescope plugin with those autocommands, im kinda confused when there's no clarification there. So im gonna leave the reference in there. This is only for people new to the config, one can always remove stuff if they make it their own.

oriori1703 (approved):
LGTM :)

tjdevries:
Closed by #1843 - thanks i had to rebase and fix some conflicts. nice thinking, this is quite elegant solve IMO

PR #1642, opened 2025-07-13
Enhance Kickstart.nvim UX with Buffer Completions, Options, and Autocommands

MN-nagy:
This pull request aims to enhances the `kickstart.nvim` configuration by integrating features that directly improve user experience, with a core focus on **session persistence**, **efficient navigation**, and **smarter completion functionality**. The changes meticulously maintain the project's foundational minimal and extensible philosophy.

---

### ⚙️ Options

* `vim.opt.undofile = true`: **Enables persistent undo history** across Neovim sessions. This enhancement allows users to undo and redo changes even after closing and reopening files, boosting productivity and safety.
* `vim.opt.smoothscroll = true`: Activates **smooth, line-by-line scrolling** (requires Neovim 0.10+). This visual improvement enhances readability and navigation, especially within large files and when traversing completion menus.
* `vim.opt.colorcolumn = '100'`: This option is intentionally commented out. If uncommented, it would visually highlight column 100 to indicate a preferred maximum line length, but it remains disabled to uphold the configuration's minimalist design principle.

### ⌨️ Keymaps

* Added `<leader>Q` mapping: Introduces a convenient, quick shortcut to **close the current buffer** (`:bd<CR>`). This streamlines buffer management and improves workflow efficiency.

### 🔄 Autocommands

* Modified `TextYankPost`: Enhanced by adding `timeout = 200`. This modification effectively **limits the duration of the yank highlight**, reducing visual distraction. Furthermore, the explicit `timeout` value offers users a clear point of customization to increase or decrease the highlight duration based on their personal preference, promoting a more tailored experience.
* Added `BufReadPost`: Implements an autocommand to **restore the cursor position on file open** (`pattern = '*'`). This ensures session continuity improves user experience by returning users directly to their last editing location across all file types.
* Added `BufWritePre`: Automatically **creates any missing parent directories when saving a file**. This eliminates a common friction point, improving usability when creating new files with nested paths.
* Commented out `FileType`: This autocommand, if enabled, would disable automatic comment continuation for all filetypes. It remains commented out, offering users the flexibility to enable or disable it based on their specific coding conventions and preferences.

### 💡 `blink.cmp` Configuration

* Integrated `buffer` source into `sources.default`: This addition to `blink.cmp` provides **relevant word completions directly from currently open buffers**. The source is further refined with a filter restricting completions to **user-chosen filetypes**, currently pre-configured for **markdown and text files**.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -170,7 +170,7 @@ vim.o.confirm = true
 vim.o.wrap = false
 
 -- Highlight max chars per line
--- vim.o.colorcolumn = '100'
+-- vim.o.colorcolumn = '120'
 
 -- [[ Basic Keymaps ]]
 --  See `:help vim.keymap.set()`
@@ -253,16 +253,6 @@ vim.api.nvim_create_autocmd('BufWritePre', {
     end
   end,
 })
-
--- disable automatic comment on newline
--- vim.api.nvim_create_autocmd('FileType', {
---   desc = 'Disable automatic comment on newline',
---   group = vim.api.nvim_create_augroup('kickstart-disable-auto-comment', { clear = true }),
---   pattern = '*',
---   callback = function()
---     vim.opt_local.formatoptions:remove { 'c', 'r', 'o' }
---   end,
--- })
 
 -- [[ Install `lazy.nvim` plugin manager ]]
 --    See `:help lazy.nvim.txt` or https://github.com/folke/lazy.nvim for more info
--- a/init.lua
+++ b/init.lua
@@ -167,13 +167,13 @@ vim.o.scrolloff = 10
 vim.o.confirm = true
 
 -- Enable undo/redo changes even after closing and reopening a file
-vim.opt.undofile = true
+vim.o.undofile = true
 
--- Enable smooth scrolling
-vim.opt.smoothscroll = true
+-- Disable line wrapping
+vim.o.wrap = false
 
 -- Highlight max chars per line
--- vim.opt.colorcolumn = '100'
+-- vim.o.colorcolumn = '100'
 
 -- [[ Basic Keymaps ]]
 --  See `:help vim.keymap.set()`
--- a/init.lua
+++ b/init.lua
@@ -906,14 +906,15 @@ require('lazy').setup({
         providers = {
           lazydev = { module = 'lazydev.integrations.blink', score_offset = 100 },
           buffer = {
-            score_offset = -1,
-            filter = function(buffer)
-              -- Filetypes for which buffer completions are enabled; add filetypes to extend
+            -- Make buffer compeletions appear at the end.
+            score_offset = -100,
+            enabled = function()
+              -- Filetypes for which buffer completions are enabled; add filetypes to extend:
               local enabled_filetypes = {
                 'markdown',
                 'text',
               }
-              local filetype = vim.bo[buffer].filetype
+              local filetype = vim.bo.filetype
               return vim.tbl_contains(enabled_filetypes, filetype)
             end,
           },
--- a/init.lua
+++ b/init.lua
@@ -166,6 +166,15 @@ vim.o.scrolloff = 10
 -- See `:help 'confirm'`
 vim.o.confirm = true
 
+-- Enable undo/redo changes even after closing and reopening a file
+vim.opt.undofile = true
+
+-- Enable smooth scrolling
+vim.opt.smoothscroll = true
+
+-- Highlight max chars per line
+-- vim.opt.colorcolumn = '100'
+
 -- [[ Basic Keymaps ]]
 --  See `:help vim.keymap.set()`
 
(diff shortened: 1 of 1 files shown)


init.lua
```diff
@@ -166,6 +166,15 @@ vim.o.scrolloff = 10
 -- See `:help 'confirm'`
 vim.o.confirm = true
 
+-- Enable undo/redo changes even after closing and reopening a file
+vim.opt.undofile = true
+
+-- Enable smooth scrolling
+vim.opt.smoothscroll = true
```
guru245: ['smoothscroll'](https://neovim.io/doc/user/options.html#'smoothscroll') option to scroll by screen line rather than by text line **when ['wrap'](https://neovim.io/doc/user/options.html#'wrap') is set**. 

`vim.o.wrap` is `true` by default, but I believe almost every developer doesn't set `true` to `wrap`. Many might add `vim.o.wrap = false` by themselves. If so, this option, `smoothscroll`, will have no effect.

In conclusion, we should think about adding `vim.o.wrap = false` explicitly in the other PR or in your PR instead of adding `smoothscroll`.

guru245 (commented):
We should use `vim.o` instead of `vim.opt` across the board. See #1495

init.lua
```diff
@@ -215,10 +227,46 @@ vim.api.nvim_create_autocmd('TextYankPost', {
...
+  desc = 'Restore cursor position on file open',
+  group = vim.api.nvim_create_augroup('kickstart-restore-cursor', { clear = true }),
+  pattern = '*',
+  callback = function()
+    local line = vim.fn.line '\'"'
+    if line > 1 and line <= vim.fn.line '$' then
+      vim.cmd 'normal! g\'"'
+    end
   end,
 })
```
guru245: Let's think about using a plugin for this. I use ['farmergreg/vim-lastplace'](https://github.com/farmergreg/vim-lastplace). Which way do you think is better?

MN-nagy: I personally think adding an **autocommand** is the better approach, as **kickstart.nvim** isn't a distro. We should avoid adding new plugins if possible.

init.lua
MN-nagy: That's a **valid point**. I'll remove `smoothscroll` and explicitly set `vim.opt.wrap = false`. You're right that most developers prefer lines not to wrap, so this will provide a more **consistent and expected experience** for `kickstart.nvim` users.

init.lua
```diff
@@ -166,6 +166,15 @@ vim.o.scrolloff = 10
 -- See `:help 'confirm'`
 vim.o.confirm = true
 
+-- Enable undo/redo changes even after closing and reopening a file
+vim.o.undofile = true
```
oriori1703: This already exists at [line 125](https://github.com/nvim-lua/kickstart.nvim/pull/1642/files#diff-0fc5c43f6f1d7605fd7d16c088ba6b83f096eee8e2049517407bbaa27aa3a4f7L125)

MN-nagy: Thanks for pointing that out, I've removed the duplicate.

init.lua
```diff
@@ -184,6 +190,9 @@ vim.keymap.set('n', '<leader>q', vim.diagnostic.setloclist, { desc = 'Open diagn
 -- or just use <C-\><C-n> to exit terminal mode
 vim.keymap.set('t', '<Esc><Esc>', '<C-\\><C-n>', { desc = 'Exit terminal mode' })
 
+-- Close current buffer
+vim.keymap.set('n', '<leader>Q', ':bd<CR>', { desc = 'Close current buffer' })
```
oriori1703, line 194: IMO this is not very useful because `<C-w>q` (which quits the window instead of the buffer) serves the purpose of closing splits, If I need to actually close buffers I usually use `<M-d>` inside of Telescope's `buffers` picker.

What do you think?

init.lua
```diff
@@ -166,6 +166,12 @@ vim.o.scrolloff = 10
 -- See `:help 'confirm'`
 vim.o.confirm = true
 
+-- Disable line wrapping
+vim.o.wrap = false
+
+-- Highlight max chars per line
+-- vim.o.colorcolumn = '100'
```
oriori1703: I think 80 and 120 are more common values, but that is just a nitpick :)

init.lua
```diff
@@ -215,10 +224,46 @@ vim.api.nvim_create_autocmd('TextYankPost', {
...
+-- disable automatic comment on newline
+-- vim.api.nvim_create_autocmd('FileType', {
+--   desc = 'Disable automatic comment on newline',
+--   group = vim.api.nvim_create_augroup('kickstart-disable-auto-comment', { clear = true }),
+--   pattern = '*',
+--   callback = function()
+--     vim.opt_local.formatoptions:remove { 'c', 'r', 'o' }
+--   end,
+-- })
+
```
oriori1703: IMO this might be too niche

oriori1703 (commented):
I do think some of the changes are a bit too niche, but the rest seem like pretty good editions :smile:

init.lua
MN-nagy: Thanks for the feedback, I agree that `<C-w>q` and Telescope's buffer picker are good ways to manage windows and buffers in advanced workflows.

However, since **Kickstart.nvim is primarily aimed at users new to Neovim**, I think keeping the `<leader>Q` mapping for `:bd` is a good idea. as for many beginners, "closing a file" intuitively means **closing the buffer**, not just the window.

init.lua
MN-nagy: That's a fair point. My goal with adding these autocommands was to give users a better idea of what's possible with them, as the original repo only included just one example.

If you still think it's too much to add to `init.lua`, I can make a new file with these autocommands, plus some other useful ones, and add it to `lua/kickstart/autocmds.lua` or a similar location. That way, users could still have them as examples.

What do you think?

MN-nagy:
@oriori1703  @guru245

Thanks for the feedback, I really appreciate you guys taking the time to review everything 🙏. I'm still pretty new to Lua and Neovim myself, so I'm learning a lot as I go.

Apologies if this PR has caused any headaches – I definitely don't want to make extra work for y'all 😅😅.

oriori1703 (approved):

tjdevries:
Thanks, but my feeling is these are a bit too specific to put in. I'll think more about the buffer source tho, that does seem pretty reasonable

PR #1627, opened 2025-07-04
Soft fork and merge non-controversial fixes and improvments

oriori1703:
> [!IMPORTANT]  
~If you are a new kickstart user, I would suggest you use those changes until they are merged into this repo.
I would try to keep this branch maintained until that point.~ 
I'm now a maintainer of kickstart.nvim, and most of the changes from this PR have been merged into the main repo.
There is no need to use this branch anymore :D

There hasn't been any real maintainer activity since the middle of May. This seem to coincide with a [comment](https://github.com/nvim-lua/kickstart.nvim/pull/1481#issuecomment-2873734227) from @feoh on my PR.

I feel kinda bad for possibly causing the burnout of the only active maintainer of kickstart with my PR , so I decided to try ease the maintenance burned by reviewing most of the open PRs and merging them to my fork. 

Those merged PRs include fixes for major and common reported issues (lsp for nvim 0.11, Mason 2.0, and guess-indent fix), as well as some non conversational (IMO) fixes and improvements (unlike my snacks PR :)
The merged PRs are:
- ~https://github.com/nvim-lua/kickstart.nvim/pull/1463~
- ~https://github.com/nvim-lua/kickstart.nvim/pull/1605~
- ~https://github.com/nvim-lua/kickstart.nvim/pull/1532~
- https://github.com/nvim-lua/kickstart.nvim/pull/1550
- ~https://github.com/nvim-lua/kickstart.nvim/pull/1616~
- https://github.com/nvim-lua/kickstart.nvim/pull/1590
- ~https://github.com/nvim-lua/kickstart.nvim/pull/1484~
- ~https://github.com/nvim-lua/kickstart.nvim/pull/1494~
- https://github.com/nvim-lua/kickstart.nvim/pull/1534
- https://github.com/nvim-lua/kickstart.nvim/pull/1582
- ~https://github.com/nvim-lua/kickstart.nvim/pull/1626~
- ~https://github.com/nvim-lua/kickstart.nvim/pull/1631~
- ~https://github.com/nvim-lua/kickstart.nvim/pull/1713~
- ~https://github.com/nvim-lua/kickstart.nvim/pull/1695~
- ~https://github.com/nvim-lua/kickstart.nvim/pull/1671~
- ~https://github.com/nvim-lua/kickstart.nvim/pull/1634~ not needed after [this change in blink](https://github.com/Saghen/blink.cmp/commit/1cc44a31f02fa54de3c1f89937bf48a2ac59d8eb)
[...]

PR #1652, opened 2025-07-18
guess-indent now requires setup

BitPusher16:
**NOTE**


dotfrag:
This already exists in https://github.com/nvim-lua/kickstart.nvim/pull/1627 and https://github.com/nvim-lua/kickstart.nvim/pull/1532.

FWIW, with lazy.nvim, you don't need to run the `.setup()` function. You can use either `opts = {}` or `config = true`. I personally prefer `opts = {}` because if I wanna change an option it feels more seamless.

tjdevries:
loaded now, sorry for delay

PR #1657, opened 2025-07-19
Switch nvim-treesitter to the new `main` branch

oriori1703:
As described in `nvim-treesitter`'s README:
> The `master` branch is frozen and provided for backward compatibility only. All future updates happen on the [`main` branch](https://github.com/nvim-treesitter/nvim-treesitter/blob/main/README.md), which will become the default branch in the future.

This is a proactive branch to switch to the new branch.
I have been using it for a couple of month and it seems pretty stable to me, though I'm not sure if this is really ready to be merged right now.
There maybe features that are not yet supported (or will never be) like incremental selection.
Does anyone else have experience with this transition, and can give feedback?

PS, I also added support for treesitter based folds, but left it commented out. Let me know if you think I should enable it in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ External Requirements:
 - Basic utils: `git`, `make`, `unzip`, C Compiler (`gcc`)
 - [ripgrep](https://github.com/BurntSushi/ripgrep#installation),
   [fd-find](https://github.com/sharkdp/fd#installation)
+- [tree-sitter CLI](https://github.com/tree-sitter/tree-sitter/blob/master/crates/cli/README.md#installation)
 - Clipboard tool (xclip/xsel/win32yank or other depending on the platform)
 - A [Nerd Font](https://www.nerdfonts.com/): optional, provides various icons
   - if you have it set `vim.g.have_nerd_font` in `init.lua` to true
@@ -185,7 +186,7 @@ winget install --accept-source-agreements chocolatey.chocolatey
 2. install all requirements using choco, exit the previous cmd and
 open a new one so that choco path is set, and run in cmd as **admin**:
 ```
-choco install -y neovim git ripgrep wget fd unzip gzip mingw make
+choco install -y neovim git ripgrep wget fd unzip gzip mingw make tree-sitter
 ```
 </details>
 <details><summary>WSL (Windows Subsystem for Linux)</summary>
@@ -195,7 +196,7 @@ wsl --install
 wsl
 sudo add-apt-repository ppa:neovim-ppa/unstable -y
 sudo apt update
-sudo apt install make gcc ripgrep unzip git xclip neovim
+sudo apt install make gcc ripgrep fd-find treesitter-cli unzip git xclip neovim
 ```
 </details>
 
@@ -205,14 +206,14 @@ sudo apt install make gcc ripgrep unzip git xclip neovim
 ```
 sudo add-apt-repository ppa:neovim-ppa/unstable -y
 sudo apt update
-sudo apt install make gcc ripgrep unzip git xclip neovim
+sudo apt install make gcc ripgrep fd-find treesitter-cli unzip git xclip neovim
 ```
 </details>
 <details><summary>Debian Install Steps</summary>
 
 ```
 sudo apt update
-sudo apt install make gcc ripgrep unzip git xclip curl
+sudo apt install make gcc ripgrep fd-find tree-sitter-cli unzip git xclip curl
 
 # Now we install nvim
 curl -LO https://github.com/neovim/neovim/releases/latest/download/nvim-linux-x86_64.tar.gz
@@ -228,14 +229,14 @@ sudo ln -sf /opt/nvim-linux-x86_64/bin/nvim /usr/local/bin/
 <details><summary>Fedora Install Steps</summary>
 
 ```
-sudo dnf install -y gcc make git ripgrep fd-find unzip neovim
+sudo dnf install -y gcc make git ripgrep fd-find tree-sitter-cli unzip neovim
 ```
 </details>
 
 <details><summary>Arch Install Steps</summary>
 
 ```
-sudo pacman -S --noconfirm --needed gcc make git ripgrep fd unzip neovim
+sudo pacman -S --noconfirm --needed gcc make git ripgrep fd tree-sitter-cli unzip neovim
 ```
 </details>
 
--- a/init.lua
+++ b/init.lua
@@ -940,28 +940,34 @@ require('lazy').setup({
   },
   { -- Highlight, edit, and navigate code
     'nvim-treesitter/nvim-treesitter',
+    lazy = false,
     build = ':TSUpdate',
-    main = 'nvim-treesitter.configs', -- Sets main module to use for opts
-    -- [[ Configure Treesitter ]] See `:help nvim-treesitter`
-    opts = {
-      ensure_installed = { 'bash', 'c', 'diff', 'html', 'lua', 'luadoc', 'markdown', 'markdown_inline', 'query', 'vim', 'vimdoc' },
-      -- Autoinstall languages that are not installed
-      auto_install = true,
-      highlight = {
-        enable = true,
-        -- Some languages depend on vim's regex highlighting system (such as Ruby) for indent rules.
-        --  If you are experiencing weird indenting issues, add the language to
-        --  the list of additional_vim_regex_highlighting and disabled languages for indent.
-        additional_vim_regex_highlighting = { 'ruby' },
-      },
-      indent = { enable = true, disable = { 'ruby' } },
-    },
+    branch = 'main',
+    -- [[ Configure Treesitter ]] See `:help nvim-treesitter-intro`
+    config = function()
+      local parsers = { 'bash', 'c', 'diff', 'html', 'lua', 'luadoc', 'markdown', 'markdown_inline', 'query', 'vim', 'vimdoc' }
+      require('nvim-treesitter').install(parsers)
+
+      vim.api.nvim_create_autocmd('FileType', {
+        pattern = parsers,
+        callback = function()
+          -- enables syntax highlighting and other treesitter features
+          vim.treesitter.start()
+
+          -- enables treesitter based folds
+          -- for more info on folds see `:help folds`
+          -- vim.wo.foldexpr = 'v:lua.vim.treesitter.foldexpr()'
+
+          -- enables treesitter based indentation
+          vim.bo.indentexpr = "v:lua.require'nvim-treesitter'.indentexpr()"
+        end,
+      })
+    end,
     -- There are additional nvim-treesitter modules that you can use to interact
     -- with nvim-treesitter. You should go explore a few and see what interests you:
     --
-    --    - Incremental selection: Included, see `:help nvim-treesitter-incremental-selection-mod`
     --    - Show your current context: https://github.com/nvim-treesitter/nvim-treesitter-context
-    --    - Treesitter + textobjects: https://github.com/nvim-treesitter/nvim-treesitter-textobjects
+    --    - Treesitter + textobjects: https://github.com/nvim-treesitter/nvim-treesitter-textobjects (using the `main` branch)
   },
 
   -- The following comments only work if you have downloaded the kickstart repo, not just copy pasted the
--- a/init.lua
+++ b/init.lua
@@ -947,11 +947,10 @@ require('lazy').setup({
     config = function()
       ---@param buf integer
       ---@param language string
-      ---@return boolean
-      local function treesitter_attach(buf, language)
-        -- check if parser exists before starting highlighter
+      local function treesitter_try_attach(buf, language)
+        -- check if parser exists and load it
         if not vim.treesitter.language.add(language) then
-          return false
+          return
         end
         -- enables syntax highlighting and other treesitter features
         vim.treesitter.start(buf, language)
@@ -962,10 +961,9 @@ require('lazy').setup({
 
         -- enables treesitter based indentation
         vim.bo.indentexpr = "v:lua.require'nvim-treesitter'.indentexpr()"
-        return true
       end
 
-      local available_parsers = require('nvim-treesitter.config').get_available()
+      local available_parsers = require('nvim-treesitter').get_available()
       vim.api.nvim_create_autocmd('FileType', {
         callback = function(args)
           local buf, filetype = args.buf, args.match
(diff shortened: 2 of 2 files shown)


guru245:
I tried to use the following repro.lua, executed with `nvim --clean -u repro.lua` and have the same problem.

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
    { -- Highlight, edit, and navigate code
      lazy = false,
      build = ':TSUpdate',
      branch = 'main',
      -- [[ Configure Treesitter ]] See `:help nvim-treesitter-intro`
      config = function()
        local parsers = { 'bash', 'c', 'diff', 'html', 'lua', 'luadoc', 'markdown', 'markdown_inline', 'query', 'vim', 'vimdoc' }
        require('nvim-treesitter').install(parsers)

        vim.api.nvim_create_autocmd('FileType', {
          pattern = parsers,
          callback = function()
            -- enbales syntax highlighting and other treesitter features
            vim.treesitter.start()

            -- enbales treesitter based folds
            -- for more info on folds see `:help folds`
            -- vim.wo.foldexpr = 'v:lua.vim.treesitter.foldexpr()'

            -- enables treesitter based indentation
            vim.bo.indentexpr = "v:lua.require'nvim-treesitter'.indentexpr()"
          end,
        })
      end,
      -- There are additional nvim-treesitter modules that you can use to interact
      -- with nvim-treesitter. You should go explore a few and see what interests you:
      --
      --    - Incremental selection: Included, see `:help nvim-treesitter-incremental-selection-mod`
      --    - Show your current context: https://github.com/nvim-treesitter/nvim-treesitter-context
      --    - Treesitter + textobjects: https://github.com/nvim-treesitter/nvim-treesitter-textobjects
    },
  },
})
```

When I run `:checkhealth nvim-treesitter`, it says,
`- ❌ ERROR tree-sitter CLI not found `

guru245:
I realized that my test was completely wrong and so I deleted my comments. Now that I've found out how to use `main` branch, I will retest and get back to you.

dotfrag:
Thank you for the PR. A bit unrelated to kickstart but.. I'm a bit confused as to where all my previous options should go. Could you provide some insight? This is my current setup https://github.com/dotfrag/dotvim/blob/main/lua/plugins/treesitter.lua

I can't see those in the new branch, are they removed?

Also noticed a typo on line 957 (enbales).

guru245:
I finally succeeded in running `main` branch. Everything is working as expected.

Installing tree-sitter CLI was a main hurdle. I think others will have the same difficulty and so I leave a how-to-do here:
1. Download a prebuilt [tree-sitter CLI](https://github.com/tree-sitter/tree-sitter/releases).
2. Place it in your `bin` directory. (For me, ~/.local/bin)
3. Change it to `tree-sitter`.
4. Apply the PR and `:Lazy sync`
5. Tada~

My thought:
This transition will cause a lot of trouble for beginners. They must setup tree-sitter CLI even before using neovim just like `fd` and `rg`. Therefore, we must describe how to install tree-sitter CLI in README.md.

oriori1703:
>Therefore, we must describe how to install tree-sitter CLI in README.md.

You are probably right :smile: . I will add that.

BTW, as mentioned in their [README](https://github.com/tree-sitter/tree-sitter/blob/master/crates/cli/README.md#installation), you can also install using `cargo`:

```sh
cargo install --locked tree-sitter-cli
```

or with `npm`:

```sh
npm install tree-sitter-cli
```

In addition to that, in some distros, you can probably use your package manager. For example, in fedora you can install it using:
```sh
sudo dnf install tree-sitter-cli
```

guru245:
Jeez. I was wondering how you found the right README.md in the source tree--there are so many of them. 😂 

Maybe it would help if I share some of the difficulties I had while trying to get started:
* I haven't used `rust` before, so I have to figure out how to use `cargo install`. 
* I don't know what `npm` is--sorry for my ignorance. I tried but I got many errors.
* Ubuntu (Probably the most widely used distro) doesn't provide the package.

💡 Not a request, just thinking out loud
In `blink.cmp`, when we use `fuzzy = { implementation = 'rust' },` it downloads the code, builds it with `cargo`, puts the binary in a `bin` directory,  and cleans up the code. Maybe doing something similar in the new tree-sitter would make the transition a lot easier.

dotfrag:
> Thank you for the PR. A bit unrelated to kickstart but.. I'm a bit confused as to where all my previous options should go. Could you provide some insight? This is my current setup https://github.com/dotfrag/dotvim/blob/main/lua/plugins/treesitter.lua
> 
> I can't see those in the new branch, are they removed?
> 
> Also noticed a typo on line 957 (enbales).

Replying to myself, I don't think this is  fully ready for everyone. I've tried multiple times to get text objects to work (with main branch) and I haven't been able to. Maybe it's a skill issue on my end but it appears to do nothing.

The context plugin does not have a working version.

And for me the autocmd doesn't work, i.e. I get no treesitter highlighting when opening a file. This is most likely on me as it seems to work for you, so I'm still investigating.

Update: Highlighting wasn't working for me, because not all parsers match filetype names. For example, for a shell script, the parser is `bash`, while the filetype is `sh`. You can see my workaround in the link.

Ktsierra:
I think the auto_install feature is essential for people like me that are new to neovim, i tried to make it work with the main branch but couldn't, maybe the rewrite is still too green?

dotfrag:
> I think the auto_install feature is essential for people like me that are new to neovim, i tried to make it work with the main branch but couldn't, maybe the rewrite is still too green?

Same here, I work in the terminal all day and open all sorts of files, I can't bother manually adding them one by one..

oriori1703:
@dotfrag
> The context plugin does not have a working version.

I works for me. Could you possibly be affected by [this breaking change](https://github.com/nvim-treesitter/nvim-treesitter-context/issues/620)?

oriori1703:
> I think the auto_install feature is essential for people like me that are new to neovim, i tried to make it work with the main branch but couldn't, maybe the rewrite is still too green?

I do agree that the `auto_install` feature is probably a deal breaker for merging this PR.
Though it could probably be replicated using a simple autocmd. I will take a crack at it when I have time.

For the time being, it is probably best to leave this PR as a draft.

dotfrag:
> @dotfrag
> 
> > The context plugin does not have a working version.
> 
> I works for me. Could you possibly be affected by [this breaking change](https://github.com/nvim-treesitter/nvim-treesitter-context/issues/620)?

Yes! Thank you.

dotfrag:
> I do agree that the `auto_install` feature is probably a deal breaker for merging this PR. Though it could probably be replicated using a simple autocmd. I will take a crack at it when I have time.

This is a (probably) badly optimised way to do it:

```lua
        vim.api.nvim_create_autocmd("FileType", {
          callback = function()
            local ft = vim.bo.filetype
            if not parsers[ft] then
              require("nvim-treesitter").install(ft)
            end
          end,
        })
```

Updated version with exclude option https://github.com/dotfrag/dotvim/blob/ecdc0ab1dca98c901f1cea2a8cf6d057e596fece/lua/plugins/treesitter.lua#L39-L54.

MeanderingProgrammer:
I wrote a plugin to fill the gaps between the `master` and `main` branches: https://github.com/MeanderingProgrammer/treesitter-modules.nvim

You're welcome to pull any of the implementations from there or use it directly.

Here's a minimal implementation of `auto_install`:

```lua
---@param buf integer
---@param language string
---@return boolean
local function attach(buf, language)
    -- check if parser exists before starting highlighter
    if not vim.treesitter.language.add(language) then
        return false
    end
    vim.treesitter.start(buf, language)
    return true
end

vim.api.nvim_create_autocmd('FileType', {
    callback = function(args)
        local buf, filetype = args.buf, args.match
        local language = vim.treesitter.language.get_lang(filetype)
        if not language then
            return
        end
        if attach(buf, language) then
            return
        end
        -- attempt to start highlighter after installing missing language
        require('nvim-treesitter').install(language):await(function()
            attach(buf, language)
        end)
    end,
})
```

dotfrag:
Thanks @MeanderingProgrammer, this works for the most part, except the cases mentioned above, where `bash` parser does not match `sh` filetype. I'm guessing a translation table is needed, unless `vim.treesitter` provides this somewhere. I believe `master` version did, but maybe it was left out from `main`.

After using several sophisticated print statement, I see that the translation from `sh` to `bash` happens correctly, but it still doesn't work for these cases 🤔. Aaand it seems to create ~infinite loop~ excessive calls with [snacks](https://github.com/folke/snacks.nvim) notification windows and blink completion windows.

MeanderingProgrammer:
`vim.treesitter.language.get_lang('sh')` should return `'bash'`, which should make all the other logic work correctly.

I don't believe I have anything special going on that makes this associate the filetype with the correct language.

What does this command `:lua vim.print(vim.treesitter.language.get_lang('sh'))` print for you? And if it prints `bash` then everything should be working so am unsure what's not working.

dotfrag:
`vim.treesitter.language.add('bash')` returns true even if it's uninstalled. Maybe that is one of the reasons. Not entirely sure to be honest. Thanks for spending time on this - I'll figure out the rest :)

MeanderingProgrammer:
It might be the case that you do have `bash` installed (even if you didn't expect to), but that is very strange if that's not the case.

If I do `:TSUninstall bash` then run `:lua vim.print(vim.treesitter.language.add('bash'))` I get this output:

```
nil
No parser for language "bash"
```

dotfrag:
> It might be the case that you do have `bash` installed (even if you didn't expect to), but that is very strange if that's not the case.
> 
> If I do `:TSUninstall bash` then run `:lua vim.print(vim.treesitter.language.add('bash'))` I get this output:
> 
> ```
> nil
> No parser for language "bash"
> ```

That wasn't my experience, and after a lot of troubleshooting I only managed to get the expected behavior after entirely resetting my setup. I deleted the following dirs and then it was fine:

```shell
  "${XDG_CONFIG_HOME:-${HOME}/.config}/nvim"
  "${XDG_DATA_HOME:-${HOME}/.local/share}/nvim"
  "${XDG_STATE_HOME:-${HOME}/.local/state}/nvim"
  "${XDG_CACHE_HOME:-${HOME}/.cache}/nvim"
```

I suspect some state/cache was stuck, because I was doing `:TSUninstall` then getting true on the check, but treesitter was obviously not starting. Thanks for your response, it helped me narrow down the issue.

oriori1703:
@MeanderingProgrammer finally got around to adding `auto_install` based on your implementation.

I made a few changes because I noticed that parsers that are bundled with nvim (e.g.  lua, c, markdown, etc ...) are not getting installed by from `nvim-treesitter` because the first call to `attach` succeeds.

It would be great if you could take a look and verify that I didn't break anything :D

MeanderingProgrammer:
I don't think anything is broken here, it will work. Some minor notes:

- since the `attach` function return value no longer leads to an early return of the callback, you can remove the return value entirely
- should avoid calling `attach` until the parser is installed via `nvim-treesitter`, avoids potential highlight changes on the initial run

I'm not sure if it makes much of a difference, but if you're interested would look something like:

```lua
---@param buf integer
---@param language string
local function attach(buf, language)
    -- check if parser exists before starting highlighter
    if not vim.treesitter.language.add(language) then
        return
    end
    vim.treesitter.start(buf, language)
    vim.bo.indentexpr = "v:lua.require'nvim-treesitter'.indentexpr()"
end

local available = require('nvim-treesitter').get_available()
vim.api.nvim_create_autocmd('FileType', {
    callback = function(args)
         local buf, filetype = args.buf, args.match
         local language = vim.treesitter.language.get_lang(filetype)
         if not language then
             return
         end
         local installed = require('nvim-treesitter').get_installed('parsers')
         if vim.tbl_contains(installed, language) then
             -- parser is already installed attach
             attach(buf, language)
         elseif vim.tbl_contains(available, language) then
             -- parser is not installed but available install it first then attach
             require('nvim-treesitter').install(language):await(function()
                 attach(buf, language)
             end)
         end
     end,
})
```

cmdrrobin:
@MeanderingProgrammer your solution doesn't work for me. For some reason the `require('nvim-treesitter').get_installed('parsers')` is empty, which will trigger an install every time.

dotfrag:
Hey @cmdrrobin, I just got around to testing Meandering's version and it seems to work fine for me, `.get_installed()` returns the correct table.

However,  I did have similar problems with what you mention in the past. I assumed it was a cache issue/bug with treesitter, so the solution I found was to reset any nvim cache. I'm sure this is an overkill, but I didn't wanna spend time finding the specific file/directory to delete. These are the dirs I deleted:

```shell
  "${XDG_DATA_HOME:-${HOME}/.local/share}/nvim"
  "${XDG_STATE_HOME:-${HOME}/.local/state}/nvim"
  "${XDG_CACHE_HOME:-${HOME}/.cache}/nvim"
```

You probably only need to clean `state` and `cache`, not `local`.

oriori1703:
> * since the attach function return value no longer leads to an early return of the callback, you can remove the return value entirely

Makes perfect sense.

> * should avoid calling `attach` until the parser is installed via `nvim-treesitter`, avoids potential highlight changes on the initial run

If you only call `attach` on parser in the `available` list you will ignore parsers in runtimepath that were installed manually / by a different plugin.

I have pushed some changes to incorporate you suggestions while still supporting parsers that were installed manually.
I also reworded some the comments.

@MeanderingProgrammer what do you think?

MeanderingProgrammer:
> If you only call attach on parser in the available list you will ignore parsers in runtimepath that were installed manually / by a different plugin.

Makes sense, that's essentially the same thing as adding an `else` to the logic I pasted above:

```lua
if vim.tbl_contains(installed, language) then
    -- parser is already installed attach
    attach(buf, language)
elseif vim.tbl_contains(available, language) then
    -- parser is not installed but available install it first then attach
    require('nvim-treesitter').install(language):await(function()
        attach(buf, language)
    end)
else
    -- parser is not installed or available try to attach in case
    -- it was installed outside of nvim-treesitter
    attach(buf, language)
end
```

I'm pretty sure your logic is functionally equivalent, so is fine as is. The only thing I would change is using the top level method to get available parsers:

- `require('nvim-treesitter.config').get_available()` -> `require('nvim-treesitter').get_available()`

captivus:
Thanks for this @oriori1703 ! I've just manually deployed your fix to `init.lua` in a fresh install and it seems to have done the trick!

mabraithwaite:
@oriori1703 as of 2 days ago it looks like the main branch is the new default branch ([commit](https://github.com/nvim-treesitter/nvim-treesitter/commit/d3218d988f72ed34414959c9ccd802d393432d6e)) so `branch = 'main'` doesn't look to be required any more.

~Also, not sure adding `tree-sitter CLI` as an external dependency is required. If you don't have it you just get warning in the `:checkhealth` => `⚠️ WARNING tree-sitter executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)`~

captivus:
@mabraithwaite I wasn't able to build the parsers without the `tree-sitter-cli` installed.

mabraithwaite:
@captivus alright yeah, for the warning message I hadn't deleted my `lazy-lock.json` so was still getting the master version, my bad

captivus:
No worries, mate!

shivan-s:
@captivus I followed your PR, and it works well for me. You're a legend, thank you!

tjdevries:
I think i managed this with some other changes. you can open new issues if something is scuffed

PR #1689, opened 2025-08-07
Fix: Prevent mason-lspconfig crash + highlight key fix

rajatshetty982:
1. Fix mason-lspconfig crash Explicitly sets: automatic_enable = false This prevents a known issue in recent versions of mason-lspconfig where a nil `enable()` function causes a crash. Keeping this disabled in Kickstart improves stability for new users.

2. Fix highlight key typo A previous commit changed `highlight` to `hl` in the `on_yank` autocmd, but related code still used `highlight`, causing inconsistency. This commit restores consistency. Feel free to revise the preferred style.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -215,7 +215,7 @@ vim.api.nvim_create_autocmd('TextYankPost', {
   desc = 'Highlight when yanking (copying) text',
   group = vim.api.nvim_create_augroup('kickstart-highlight-yank', { clear = true }),
   callback = function()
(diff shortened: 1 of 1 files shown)


guru245:
1. We already have a few similar PRs such as #1590.
2. `hl` is not a typo. `highlight` changed to `hl` in nvim 0.11.x. You should use 0.11.x.

rajatshetty982:
Got it👍🏽

PR #1695, opened 2025-08-11
Add 'gr' group to which-keys

thiago-negri:
Adds 'gr' to which-keys documentation, so users can see that LSP actions are grouped after 'gr' key binds.

It used to be directly under 'g', which required no additional grouping info on which-keys. See #1427 for further explanation on the key binds change.


guru245:
I use the same. 😊

oriori1703 (approved):
Looks great

oriori1703:
This was added in 80b1ee1789d19b3c949ab50ba959f412cfc5e54b

PR #1738, opened 2025-10-06
fix: remove usages of deprecated `vim.lsp.protocol.Methods`

MariaSolOs:
These were deprecated in https://github.com/neovim/neovim/pull/35998


tjdevries:
thanks, i did in separate pr just because i had already generated some conflicts. my b on the delay

PR #1743, opened 2025-10-08
Migrate mason-lspconfig handlers to vim.lsp.config

ReadHegel:
As stated in the [`mason-lspconfig` changelog](https://github.com/mason-org/mason-lspconfig.nvim/blob/6bdb14f230de0904229ec367b410fb817e59b072/CHANGELOG.md), they **removed the `handlers` configuration** in their plugin, making Kickstart no longer responsible for automatic LSP server configuration.
```
### Removed Features

- Removed the `handlers` setting and `.setup_handlers()` function.
- Replaced by the new native `vim.lsp.config()` API.
- Introduced a new `automatic_enable` setting.

```

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -700,6 +700,16 @@ require('lazy').setup({
         },
       }
 
+      -- Now setup those configurations
+      for name, config in pairs(servers) do
+        local config = config or {}
+        -- This handles overriding only values explicitly passed
+        -- by the server configuration above. Useful when disabling
+        -- certain features of an LSP (for example, turning off formatting for ts_ls)
+        config.capabilities = vim.tbl_deep_extend('force', {}, capabilities, config.capabilities or {})
+        vim.lsp.config(name, config)
+      end
+
       -- Ensure the servers and tools above are installed
       --
       -- To check the current status of installed tools and/or manually install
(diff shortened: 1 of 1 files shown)


meierjan:
If I understand this correctly, all config settings that are currently (so, before this fix) set within `init.lua` via `local servers = {}` are not applied: contrary to what is documented. I'm a nobody, so it took me quite some time to debug and figure out that this is the source of the problem.

@tjdevries, if that is correct, any chance we can merge this to prevent other starters from going crazy, as their LSP configs are never applied?

Thank you for the project & the fix ❤️

runemoennike:
I spent way too long realizing this was the reason my attempts to fix an issue with ElixirLS seemed to have no effect at all :(

init.lua
```diff
@@ -700,6 +700,16 @@ require('lazy').setup({
...
       }
 
+      -- Now setup those configurations
+      for name, config in pairs(servers) do
+        local config = config or {}
+        -- This handles overriding only values explicitly passed
+        -- by the server configuration above. Useful when disabling
+        -- certain features of an LSP (for example, turning off formatting for ts_ls)
+        config.capabilities = vim.tbl_deep_extend('force', {}, capabilities, config.capabilities or {})
+        vim.lsp.config(name, config)
```
versi786, line 710: I think this is missing one line as mentioned [here](https://github.com/nvim-lua/kickstart.nvim/issues/1776#issuecomment-3654419545). Without this change the LSP does not start automatically when a file is opened.
```suggestion
        vim.lsp.config(name, config)
        vim.lsp.enable(name)
```

tjdevries:
Ya, my bad - should be fixed now

PR #1746, opened 2025-10-14
config guess-indent in setup

WindSoilder:
**NOTE**
It seems that `guess-indent` doesn't work unless you assign `config` manually, here is what I get from https://github.com/NMAC427/guess-indent.nvim:
```
-- using packer.nvim
use {
  'nmac427/guess-indent.nvim',
  config = function() require('guess-indent').setup {} end,
}
```
And after define `config` property, it works well for me.


bewuethr:
This is also the subject of #1532, #1598, #1652, and included in #1627.

tjdevries:
Thanks, I loaded this now.

PR #1747, opened 2025-10-14
fix: removes deprecated code in favor of new code

Tigatok:
Removes deprecation. I found older issues that seemed to do the same thing, and more, but I think this is all that is needed.


tjdevries:
Thanks, took care of this and some related lsp stuff.

Issue #1771, opened 2025-11-17
Failing to install treesitter


byk4:
<!-- Any bug report not following this template will be immediately closed. Thanks -->

## Before Reporting an Issue
- I have read the kickstart.nvim README.md.
- I have read the appropriate plugin's documentation.
- I have searched that this issue has not been reported before.

- [x] **By checking this, I confirm that the above steps are completed. I understand leaving this unchecked will result in this report being closed immediately.**

## Describe the bug
<!-- A clear and concise description of what the bug is. -->
I have a fresh installation of neovim and I just cloned modular kickstart from https://github.com/dam9000/kickstart-modular.nvim but I was having problems with treesitter installation. So I removed nvim-data as well as deleted the config and cloned the og config from this repo but still I am having the problems. All of this was working fine on my prevous windows 10 os and on mint os. But now I am getting these errors
```
nvim-treesitter[luadoc (or any language)] Error during compilation
nvim-treesitter[1/11, failed: 1] Checking out locked revision
```

## To Reproduce
<!-- Steps to reproduce the behavior. -->
1. We windows 10 machine
2. install neovim
3. git clone this repo
4. get treesitter installation errors.

## Desktop
<!-- please complete the following information. -->
- OS: Windows 10
- Terminal: Windows 10 default command prompt

## Neovim Version
<!-- Output of running `:version` from inside of neovim. -->
:version
NVIM v0.11.5                                                                                                            Build type: Release                                                                                                     LuaJIT 2.1.1741730670                                                                                                   Run ":verbose version" for more info      

## TS Info
config info
```                                                                                                                                                                                                                                                                            {
[...]

PR #1799, opened 2025-12-19
Treesitter fix

kdave5:
Updated nvim-treesitter config to follow the stable 'master' branch (as opposed to the now default 'main' branch) as a temporary fix.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -940,6 +940,7 @@ require('lazy').setup({
   },
   { -- Highlight, edit, and navigate code
     'nvim-treesitter/nvim-treesitter',
+    branch = "master",
     build = ':TSUpdate',
     main = 'nvim-treesitter.configs', -- Sets main module to use for opts
     -- [[ Configure Treesitter ]] See `:help nvim-treesitter`
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ test.sh
 nvim
 
 spell/
+lazy-lock.json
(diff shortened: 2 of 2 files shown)


bewuethr:
There's also #1657 with all the changes to accommodate `main`.

init.lua
```diff
@@ -940,6 +940,7 @@ require('lazy').setup({
   },
   { -- Highlight, edit, and navigate code
     'nvim-treesitter/nvim-treesitter',
+    branch = "master",
     build = ':TSUpdate',
     main = 'nvim-treesitter.configs', -- Sets main module to use for opts
```
adrianveen, line 945: ~You just need to change `.configs` to `.config` on line 945 per this issue: https://github.com/nvim-lua/kickstart.nvim/issues/1771~

clason: This is **WRONG**, I don't understand why people keep repeating this nonsense. This PR is the **ONLY** correct fix(!). 

(Transitioning to the `main` branch would be a breaking change for users, since that is effectively a different plugin with much reduced functionality (but much better maintainability.))

adrianveen: My bad - crossed it out.

tjdevries:
ok, i think i did the right fix. Sorry for delays.

clason:
Note: It's unlikely that the `master` branch will be compatible with Nvim 0.12 when it is released (and nightlies before that, of course -- but if you run nightlies you are _asking_ for breaking changes). It is also no longer compatible with the latest tree-sitter CLI (0.26.x, so you need to install <=0.25.10 to generate from grammar).

Issue #1802, opened 2025-12-21
Failed to run 'config' for nvim-treesitter


steve0613:
## Before Reporting an Issue
- I have read the kickstart.nvim README.md.
- I have read the appropriate plugin's documentation.
- I have searched that this issue has not been reported before.

- [ X ] **By checking this, I confirm that the above steps are completed. I understand leaving this unchecked will result in this report being closed immediately.**

## Describe the bug
..local/share/nvim/lazy/lazy.nvim/lua/lazy/core/loader.lua:387: module 'nvim-treesitter.configs' not found:
no field package preload[' nvim-treesitter.configs'] cache_loader: module 'nvim-treesitter.configs' not found cache_loader_lib: module 'nvim-treesitter.configs' not found no file './nvim-treesitter/configs.lua'
no file '/usr/local/share/luajit-2.1/nvim-treesitter/configs.lua' no file '/usr/local/share/lua/5.1/nvim-treesitter/configs.lua' no file '/usr/local/share/lua/5.1/nvim-treesitter/configs/init.lua' no file './nvim-treesitter/configs.so'
no file '/usr/local/lib/lua/5.1/nvim-treesitter/configs.so' no file '/usr/local/lib/lua/5.1/loadall.so' no file './nvim-treesitter.so'
no file '/usr/local/lib/lua/5.1/nvim-treesitter.so' no file /usr/local/lib/lua/5.1/loadall.so'
# stacktrace:
- .config/nvim/lua/lazy-plugins.lua:15
- .config/nvim/init.lua:106

## To Reproduce

1. This occurs the first time you run nvim and each time thereafter.  In my research I think there was an update to tree sitter and syntax has changed in the way that it is initialized but I could never figure it out.  

## Desktop

- OS: macOS Tahoe 26.2
- Terminal: Apple's system terminal

## Neovim Version
NVIM v0.11.5                                                                                                            
Build type: Release                                                                                                     
LuaJIT 2.1.1765228720

<img width="886" height="661" alt="Image" src="https://github.com/user-attachments/assets/aa6aa2e1-905e-4e35-8a53-05f49a8460f3" />

akaJuliaan:
take a look at #1657

hippietrail:
Duplicate of #1732

hemanthkotaprolu:
Just change the following line in your `~/.config/nvim/init.lua` file.

```
main = 'nvim-treesitter.configs', -- Sets main module to use for opts
```

to 

[...]

PR #1795, opened 2025-12-16
Update init.lua

mvargasmoran:
nvim-treesitter changed the name from configs to config This updates this so fresh installs don't complain about a missing file. that's all.

***************************************************************************
**NOTE**
Please verify that the `base repository` above has the intended destination!
Github by default opens Pull Requests against the parent of a forked repository.
If this is your personal fork and you didn't intend to open a PR for contribution
to the original project then adjust the `base repository` accordingly.
**************************************************************************

PR #1748, opened 2025-10-16
fix: nvim-treesitter rename configs.lua -> config.lua

philwhln:
Addresses https://github.com/nvim-lua/kickstart.nvim/issues/1732#issuecomment-3409100884

PR #1811, opened 2025-12-29
Fix main module path for nvim-treesitter

subh3ndu:
Screenshot 2025-12-29 231704
The above error is I was getting while installing kickstart.nvim in windows terminal (Git bash), changing `nvim-treesitter.configs` to `nvim-treesitter.config` fixes the error.

**NOTE**

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -941,7 +941,7 @@ require('lazy').setup({
   { -- Highlight, edit, and navigate code
     'nvim-treesitter/nvim-treesitter',
     build = ':TSUpdate',
-    main = 'nvim-treesitter.configs', -- Sets main module to use for opts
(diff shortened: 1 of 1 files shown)


techieshark:
Confirming this change fixes the issue here: https://github.com/nvim-lua/kickstart.nvim/issues/1802

bewuethr:
Duplicates: #1795, #1748

xiechunyao:
changing .configs to .config only suppress the error messages, but Treesitter doesn't actually function properly. For the fastest fix that actually works, just add `branch = 'master'`. more info: https://github.com/nvim-lua/kickstart.nvim/issues/1802#issuecomment-3704075358

PR #1826, opened 2026-01-12
Fix nvim-treesitter lazy.nvim config by removing invalid 'main' field

r0landd:
### Problem

On Windows, setting `main = 'nvim-treesitter.configs'` causes lazy.nvim to attempt
`require('nvim-treesitter.configs')`, but this module does not exist in
nvim-treesitter. As a result, Neovim fails to load the plugin.

Other platforms may not trigger the same error, but the configuration
is invalid and can cause issues.

### Solution

Remove the `main` field and rely on lazy.nvim’s default behavior, which
correctly calls `require('nvim-treesitter.configs').setup(opts)`.

### Result

Neovim starts normally on Windows (and other platforms) and treesitter
is configured as expected.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -941,7 +941,6 @@ require('lazy').setup({
   { -- Highlight, edit, and navigate code
     'nvim-treesitter/nvim-treesitter',
     build = ':TSUpdate',
-    main = 'nvim-treesitter.configs', -- Sets main module to use for opts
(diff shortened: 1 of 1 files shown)


r0landd:
Removing main avoids the error, but the config still needs a proper setup() call. I’ll reopen with a corrected config.

PR #1847, opened 2026-01-28
fix: lua_ls no longer exists in mason registry

veonim:
I just did a fresh install today and was thrown into an error that lua_ls not found.

Switched to lua-language-server, making this PR for base repository because other people might face this issue.

Edit:
Added `{ "mason-org/mason-lspconfig.nvim", opts = {} }` based on the comment of [vimjeremy](https://github.com/nvim-lua/kickstart.nvim/pull/1847#issuecomment-3809910676), worked flawlessly

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -481,7 +481,10 @@ require('lazy').setup({
       -- Mason must be loaded before its dependents so we need to set it up here.
       -- NOTE: `opts = {}` is the same as calling `require('mason').setup({})`
       { 'mason-org/mason.nvim', opts = {} },
+
+      -- This plugin streamlines Neovim's LSP setup by automating server installation and activation, providing helpful management commands, and mapping mason.nvim packages to nvim-lspconfig configurations.
       { 'mason-org/mason-lspconfig.nvim', opts = {} },
+
       'WhoIsSethDaniel/mason-tool-installer.nvim',
 
       -- Useful status updates for LSP.
--- a/init.lua
+++ b/init.lua
@@ -614,7 +614,7 @@ require('lazy').setup({
       -- You can press `g?` for help in this menu.
       local ensure_installed = vim.tbl_keys(servers or {})
       vim.list_extend(ensure_installed, {
-        'lua_ls', -- Lua Language server
+        'lua-language-server', -- Lua Language server
         'stylua', -- Used to format Lua code
         -- You can add other tools here that you want Mason to install
       })
@@ -628,7 +628,7 @@ require('lazy').setup({
       end
 
       -- Special Lua Config, as recommended by neovim help docs
-      vim.lsp.config('lua_ls', {
+      vim.lsp.config('lua-language-server', {
         on_init = function(client)
           if client.workspace_folders then
             local path = client.workspace_folders[1].name
@@ -652,7 +652,7 @@ require('lazy').setup({
           Lua = {},
         },
       })
-      vim.lsp.enable 'lua_ls'
+      vim.lsp.enable 'lua-language-server'
     end,
   },
 
--- a/init.lua
+++ b/init.lua
@@ -481,6 +481,7 @@ require('lazy').setup({
       -- Mason must be loaded before its dependents so we need to set it up here.
       -- NOTE: `opts = {}` is the same as calling `require('mason').setup({})`
       { 'mason-org/mason.nvim', opts = {} },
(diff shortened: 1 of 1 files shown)


vimjeremy:
This can be fixed by adding `{ "mason-org/mason-lspconfig.nvim", opts = {} }` to mason's dependencies

Personally, I think this is the cleaner way because it bridges the gap between lspconfig and mason and allows the installation and auto-enabling of lsp servers by the actual configuration name so maintaining the project will be easier over time.

veonim:
@vimjeremy Yeah, it did handled that perfectly. Thanks for suggestion, updated the PR

init.lua
```diff
@@ -481,6 +481,7 @@ require('lazy').setup({
       -- Mason must be loaded before its dependents so we need to set it up here.
       -- NOTE: `opts = {}` is the same as calling `require('mason').setup({})`
       { 'mason-org/mason.nvim', opts = {} },
+      { 'mason-org/mason-lspconfig.nvim', opts = {} },
```
ro0gr, line 486: It might need a brief explanation of why its needed.

It turns out, that's not necessarily clear, especially for newcomers, e.g.
https://www.reddit.com/r/neovim/comments/1pnqyi8/please_clarify_the_purpose_of_masonlspconfig/

vimjeremy: mason-lspconfig enables auto-enable and enables the installing of LSP servers by configuration name.
eg. in nvim-lspconfig the configurations for lua-language-server is named lua_ls

so with mason-lspconfig you can :MasonInstall lua_ls instead of lua-language-server which is actually done in kickstart.nvim

ro0gr: Sorry for my confusing comment then 😅 

I mean it seems reasonable to supply this dependency with a comment explaining why it's needed, to avoid removing it again without a good reason https://github.com/veonim/kickstart.nvim/commit/ad246eb5dd5fdd3e4ad42ae974b919ff887f2f0a

nemanjaglumac:
If it means anything, I can confirm it works like a charm.
Thank you

ArionMiles:
Thanks for this!

poodbooq (approved):

youssofatepsor (approved):

badbabykosh:
> I just did a fresh install today and was thrown into an error that lua_ls not found.
> 
> Switched to lua-language-server, making this PR for base repository because other people might face this issue.
> 
> Edit: Added `{ "mason-org/mason-lspconfig.nvim", opts = {} }` based on the comment of [vimjeremy](https://github.com/nvim-lua/kickstart.nvim/pull/1847#issuecomment-3809910676), worked flawlessly

I'm a noob at this. At what line exactly in init.lua do I add this? And what should it look like exactly after I add it?

kalebingerick:
Hey @badbabykosh , you'll want to add that dependency somewhere around [this line](https://github.com/nvim-lua/kickstart.nvim/blob/e79572c9e6978787af2bca164a85ab6821caeb7b/init.lua#L483). I added it right after.

kindofabuzz:
I've done all these changes to init.lua and still get the error.

veonim:
> I've done all these changes to init.lua and still get the error.

You can check the diff in the PR and see where it is placed.

zwhitchcox (approved):

init.lua
```diff
@@ -481,6 +481,10 @@ require('lazy').setup({
       -- Mason must be loaded before its dependents so we need to set it up here.
       -- NOTE: `opts = {}` is the same as calling `require('mason').setup({})`
       { 'mason-org/mason.nvim', opts = {} },
+
+      -- This plugin streamlines Neovim's LSP setup by automating server installation and activation, providing helpful management commands, and mapping mason.nvim packages to nvim-lspconfig configurations.
```
giogt, line 485: This comment line is too long. There seems to be a limit of 100 chars for all other comments in this file.

giogt: I couldn't find a way to configure Stylua to format comment lines with a 100 column width by changing https://github.com/nvim-lua/kickstart.nvim/blob/master/.stylua.toml

All the other comments are formatted that way though, not sure if manually or with a different formatter.

oriori1703:
I fixed this in https://github.com/nvim-lua/kickstart.nvim/commit/1ba17893188d9bdbe483e35b77ce3678b6cd30c3 by re-adding mason-lspconfig as a dependency.
Let me know if you still experience issues 😃

PR #1896, opened 2026-02-22
Clarify comment in .gitignore for lazy-lock.json

giogt:
The current comment in .gitignore for lazy-lock.json suggests to uncomment it yourself, but you actually need to comment it instead.

This change modifies the comment to clarify the recommended approach for kickstart users.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,8 +5,9 @@ nvim
 
 spell/
 
-# You can uncomment this yourself if you want to lock the lazy-lock.json,
-# but for kickstart, it makes sense to leave it ignored.
+# You likely want to comment this, since it's recommended to track lazy-lock.json in version
+# control, see https://lazy.folke.io/usage/lockfile
+# For kickstart, it makes sense to leave it ignored.
 lazy-lock.json
 
 .DS_Store


oriori1703:
Makes sense :)

.gitignore
```diff
@@ -5,8 +5,9 @@ nvim
 
 spell/
 
-# You can uncomment this yourself if you want to lock the lazy-lock.json,
-# but for kickstart, it makes sense to leave it ignored.
+# You likely want to comment this, since it's recommended to track lazy-lock.json in version
+# control, see https://lazy.folke.io/usage/lockfile
+# For kickstart, it makes sense to leave it ignored.
```
rmacklin, line 10: I'd suggest a slight further clarification:
```suggestion
# For the canonical kickstart git repository, it makes sense to leave it ignored.
```
since otherwise it could be confusing for someone who interprets kickstart to more generally refer to this base set of nvim configuration they start from.

rmacklin: Opened a PR here (slightly improved from what I suggested above): https://github.com/nvim-lua/kickstart.nvim/pull/1914

PR #1914, opened 2026-02-28
Revise comment for lazy-lock.json in .gitignore

rmacklin:
This is a follow-up to https://github.com/nvim-lua/kickstart.nvim/pull/1896 to hopefully make things even clearer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,9 +5,10 @@ nvim
 
 spell/
 
-# You likely want to comment this, since it's recommended to track lazy-lock.json in version
-# control, see https://lazy.folke.io/usage/lockfile
-# For kickstart, it makes sense to leave it ignored.
+# In your personal fork, you likely want to comment this, since it's recommended to track
+# lazy-lock.json in version control - see https://lazy.folke.io/usage/lockfile
+# For the official `nvim-lua/kickstart.nvim` git repository, we leave it ignored to avoid unneeded
+# merge conflicts.
 lazy-lock.json
 
 .DS_Store


eamonburns:
I'm not sure users really need to know the specifics of _why_ it is ignored in the official kickstart.nvim repo, they just need to know that it is, and that it is recommended they don't in their own repo.

https://github.com/nvim-lua/kickstart.nvim/blob/dabce46993048a4d1a9c0b6f5bd002848cfe9802/README.md?plain=1#L72-L74

rmacklin:
I can see that line of thinking, but I was explicitly asked about why it's ignored, so at least some people do care to understand. And since I don't think the brief reasoning is otherwise harmful for anyone else, IMO it's worth having for any similarly curious newcomers. (Also anyone can delete the whole line + comment in their own gitignore, so even if it bothers someone they don't have to keep it.)

.gitignore
```diff
@@ -7,7 +7,8 @@ spell/
 
 # You likely want to comment this, since it's recommended to track lazy-lock.json in version
 # control, see https://lazy.folke.io/usage/lockfile
-# For kickstart, it makes sense to leave it ignored.
+# For the canonical kickstart git repository, we leave it ignored to avoid unneeded diff churn
+# and conflict resolutions for PRs that wouldn't otherwise have conflicts.
```
oriori1703: What do you think of the following?
```suggestion
# In your personal fork, you likely want to comment this, since it's recommended to track lazy-lock.json in version
# control, see https://lazy.folke.io/usage/lockfile
# For the official `nvim-lua/kickstart.nvim` git repository, we leave it ignored to avoid to avoid merge conflicts.
```

rmacklin: Updated: 164cedf2124b63aca67573c3c1d1365d27aab750

rez1-dev (approved):

oriori1703:
LGTM :D

PR #1925, opened 2026-03-07
Clarify gitsigns keymap for which-key

nathanzeng:
The keymaps for gitsigns are not enabled by default, so I found it confusing that they were documented in which-key already.

This PR simply adds a clarifying comment on that which-key configuration.


oriori1703:
seems reasonable :)

PR #1945, opened 2026-03-14
fix #1933: Allow mason to automatically enable its installed servers

### LINKED ISSUE #1933: Mason installed LSP server don't seems to work
I just updated to the latest version of kickstart.nvim. I could not make my python lsp work until I uncommented the auto installation of pyright in the config. If I only installed it with Mason, it was not working when I opened a python file.

I tried installing the python parser manually with TSInstall python and it still didn't work.

It's probably a skill issue on my part, but I wonder what changed, in the past I could add a lsp to Mason and it was working without touching anything in the init.lua file.

Thanks!

vivianludrick:
using `require('mason-lspconfig').setup()` now autoenables the lsps which mason has installed
this pr will preserve the approach previous kickstart used to have of auto enabling them

also `mason-lspconfig` now has the ability to autoinstall the listed servers so the mason-tool-installer plugin is redundant now

[Link to `ensure_installed` ability in mason-lspconfig](https://github.com/mason-org/mason-lspconfig.nvim?tab=readme-ov-file#configuration)

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -496,7 +496,6 @@ require('lazy').setup({
       },
       -- Maps LSP server names between nvim-lspconfig and Mason package names.
       'mason-org/mason-lspconfig.nvim',
-      'WhoIsSethDaniel/mason-tool-installer.nvim',
 
       -- Useful status updates for LSP.
       { 'j-hui/fidget.nvim', opts = {} },
@@ -655,7 +654,7 @@ require('lazy').setup({
         -- You can add other tools here that you want Mason to install
       })
 
-      require('mason-tool-installer').setup { ensure_installed = ensure_installed }
+      require('mason-lspconfig').setup { ensure_installed = ensure_installed }
 
       for name, server in pairs(servers) do
         vim.lsp.config(name, server)


rez1-dev (approved):

sambartik (changes requested):
This breaks the setup. `ensure_installed` in `mason-lspconfig` only accepts LSPs. 

The current auto install method works with other mason packages that are not LSPs.

oriori1703:
As sambartik said, mason-tool-installer is still needed for non-lsp tools.
The correct way to implement it should be by adding a call to mason-lspconfig's `.setup()` after the setup of mason-tool-installer. e.g. :
```lua
  require('mason-lspconfig').setup(automatic_enable=true)
```
(`automatic_enable` is already the default when calling `.setup()` but I think this is more readable this way.

But IMO, it is better to have it enabled explicitly in the config instead, so I'm closing this for now.
If you still think we should turn on `automatic_enable` please open a [discussion](https://github.com/nvim-lua/kickstart.nvim/discussions).

PR #1953, opened 2026-03-19
feat: use ts-install.nvim to automatically install treesitter parsers

### LINKED ISSUE #1951: Automatically install treesitter parsers with ts-install.nvim
Hello, I recently synced my fork, and I noticed that I lost syntax highlighting in most of filetypes I use daily.

Investigation showed me that treesitter's `auto_install=true` was dropped in https://github.com/nvim-lua/kickstart.nvim/commit/7ea937dbc5933c7769b163e079b244f3a7a45063.

To get highlighting back, I started adding missing filetypes to the list of parsers manually one-by-one, but it's quite annoying once you were used to `auto_install`.

So I dug more. In treesitter commit which removed `auto_install` they suggest using a separate plugin to keep the old behavior: https://github.com/nvim-treesitter/nvim-treesitter/commit/ab230eadd4a96baec86fb17bd625893649f2612f

https://github.com/lewis6991/ts-install.nvim

I was able to make it work in my fork: https://github.com/tpwo/kickstart.nvim/commit/ab0222a51b188960b63e5c3ec19f70ea4674b2ce

Would you accept a PR doing something like that for the main repo? It's adding one more plugin, but having parsers installed automatically is quite convenient, so IMO it makes sense for the target audience of kickstart. Do you agree?

tpwo:
Resolves #1951.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -874,8 +874,6 @@ require('lazy').setup({
     branch = 'main',
     -- [[ Configure Treesitter ]] See `:help nvim-treesitter-intro`
     config = function()
-      local parsers = { 'bash', 'c', 'diff', 'html', 'lua', 'luadoc', 'markdown', 'markdown_inline', 'query', 'vim', 'vimdoc' }
-      require('nvim-treesitter').install(parsers)
       vim.api.nvim_create_autocmd('FileType', {
         callback = function(args)
           local buf, filetype = args.buf, args.match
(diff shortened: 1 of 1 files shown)


jgw6fg8qng-ai (commented):
I'm now using tammypesson@icloud.com which has my legal name my legal address is 1404 Carmel Dr. apt#23C Lafayette,Louisiana 70501 my cell phone number is (337) 376-3368 
My on SSI Disability Benefits and Medicaid of Louisiana Health Insurance is AmeriHealth

init.lua
```diff
@@ -899,6 +897,20 @@ require('lazy').setup({
...
+  { -- Automatically install Treesitter parsers for languages you encounter.
+
+    -- NOTE: This plugin is a convenience feature and is not strictly necessary for
+    -- regular use. If you prefer, you can remove it and install parsers manually
+    -- directly with treesitter using `require('nvim-treesitter').install()`.
+    'lewis6991/ts-install.nvim',
+    dependencies = { 'nvim-treesitter/nvim-treesitter' },
+    config = function()
+      require('ts-install').setup {
+        ensure_install = { 'bash', 'c', 'diff', 'html', 'lua', 'luadoc', 'markdown', 'markdown_inline', 'query', 'vim', 'vimdoc' },
```
Christoph-Raab: Could this be a merge of an array with those defaults and an empty array where one can configure custom parsers?

That would simplify configuring custom parsers with less possibility for merge conflicts.

Something like

```lua
local my_parsers = {}
local default_parsers = { 'bash', 'c', 'diff', 'html', 'lua', 'luadoc', 'markdown', 'markdown_inline', 'query', 'vim', 'vimdoc' }

require('ts-install').setup {
  ensure_install = vim.list_extend(my_parsers, default_parsers),
  auto_install = true,
}
```

tpwo: Good point, I noticed that similar approach is used for Mason's `ensure_installed`, and I replicated that approach: https://github.com/nvim-lua/kickstart.nvim/blob/0619d89884dec2fa911430fdc42796420421fd6b/init.lua#L653-L658

tpwo:
Closing this PR, as #1956 implements the same behavior without relying on a new plugin.

PR #1956, opened 2026-03-20
Auto install treesitter parsers

### LINKED ISSUE #1951: Automatically install treesitter parsers with ts-install.nvim

oriori1703:
Auto install treesitter parsers when opening a file.

Similar functionality to #1953, but takes a different approach by implementing the feature ourselves instead of using a plugin.
Based on my initial implementation in #1657 (almost a direct copy :)

As discussed in https://github.com/nvim-lua/kickstart.nvim/issues/1951#issuecomment-4100762912, I'm still not sure if this is too much for kickstart, so if you have feedback, I would love to hear it in that thread. 

closes #1951

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -903,7 +903,18 @@ require('lazy').setup({
           local language = vim.treesitter.language.get_lang(filetype)
           if not language then return end
 
-          treesitter_try_attach(buf, language)
+          local installed_parsers = require('nvim-treesitter').get_installed 'parsers'
+
+          if vim.tbl_contains(installed_parsers, language) then
+            -- enable the parser if it is installed
+            treesitter_try_attach(buf, language)
+          elseif vim.tbl_contains(available_parsers, language) then
+            -- if a parser is available in `nvim-treesitter` auto install it, and enable it after the installation is done
+            require('nvim-treesitter').install(language):await(function() treesitter_try_attach(buf, language) end)
+          else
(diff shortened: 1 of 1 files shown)


dotfrag:
I went back and forth on having this in my config a few times and still can't decide if I like it or not. The only feedback I have is that I prefer this instead of a separate plugin. Most of us already have a bunch of plugins already..

Issue #1971, opened 2026-03-30
Neovim 0.12 !!!!


eamonburns:
Neovim 0.12 just dropped, with [a bunch of features](https://neovim.io/doc/user/news-0.12/). We should consider what makes sense to add to kickstart.

Note: Replacing `lazy.nvim` with `vim.pack` is a bigger discussion than what is meant for this issue. This issue is meant for relatively small/uncontroversial changes. Also, there is already an issue open: #1630

Here are some initial things:
- The default mappings of `mini.ai` clash with the new [`v_an`](https://neovim.io/doc/user/treesitter/#v_an) and [`v_in`](https://neovim.io/doc/user/treesitter/#v_in) mappings. ~~This is probably just a bug with `mini.ai`~~ (This is a [known issue](https://nvim-mini.org/mini.nvim/doc/mini-ai.html#miniai.config-mappings)).
- The [`'autocomplete'`](https://neovim.io/doc/user/options/#'autocomplete') option may be a good option to enable by default. I am not sure how it would interact with blink.cmp though.
- Enabling [`ui2`](https://neovim.io/doc/user/lua/#ui2) might be a good thing to have commented out somewhere in the config. Or, maybe not 🤷‍♂️ something to think about.

SetsuikiHyoryu:
I’m really happy with how Neovim 0.12 has turned `nvim-lspconfig` (0.10) and `nvim-treesitter` into mere configuration and parser downloaders, while also allowing me to drop `lazy.nvim` entirely. I heard that native APIs for `blink.cmp`-like functionality existed before but weren't mature enough to replace it. I'm hoping 0.12 will finally allow native APIs to replace `blink.cmp` as well.

oriori1703:
> The default mappings of mini.ai clash with the new [v_an](https://neovim.io/doc/user/treesitter/#v_an) and [v_in](https://neovim.io/doc/user/treesitter/#v_in) mappings. This is probably just a bug with mini.ai.

This seems to be [known](https://nvim-mini.org/mini.nvim/doc/mini-ai.html#miniai.config-mappings) by mini.nvim and [not classified as a bug](https://github.com/nvim-mini/mini.nvim/discussions/1887#discussioncomment-13698326).
I'm still not sure what do with because I agree with echasnovski that those are good mappings for `mini.ai`.
Also, I would guess that most kickstart users are already used to them (including me). Though we may need to rip the band aid anyway, so it may as well be now.

[...]

Issue #1630, opened 2025-07-05
Use `vim.pack` as package manager when Nvim 0.12 is released


brianhuster:
Nvim just merge a PR that add a built-in plugin manager `vim.pack`. I think kickstart.nvim should switch to it when Nvim 0.12 is released, because it is built-in

guru245:
We don't have to use built-in features just because they're built-in -- we chooose what works best. For example, even though Neovim 0.11 introduced built-in auto-completion, we opted for `blink.cmp` because it offered a better experience. 

Thanks for the update about nvim 0.12 introducing `vim.pack`. However, I suspect `vim.pack` isn't yet as mature as `lazy.nvim`. If we are going to adopt `vim.pack`, it should be because it is clearly better -- not simply because it's built-in.

laurensdc:
`vim.pack` seems to do less magic than LazyVim. I think this is in line with the spirit of kickstart, where we want people to learn Neovim and have control and understanding of their config. 
This should also be taken into account in weighing this decision.

brianhuster:
Yes, also I don't think it's nice for beginners to copy this code to their config

```lua
local lazypath = vim.fn.stdpath 'data' .. '/lazy/lazy.nvim'
if not (vim.uv or vim.loop).fs_stat(lazypath) then
  local lazyrepo = 'https://github.com/folke/lazy.nvim.git'
  local out = vim.fn.system { 'git', 'clone', '--filter=blob:none', '--branch=stable', lazyrepo, lazypath }
  if vim.v.shell_error ~= 0 then
    error('Error cloning lazy.nvim:\n' .. out)
  end
end

---@type vim.Option
local rtp = vim.opt.rtp
rtp:prepend(lazypath)
```

sebastianstudniczek:
What about lazy loading? Won't using vim.pack slow nvim startup cause it will load all plugins?
AFAIK vim.pack in 0.12 is just bare minimum for a package manager

brianhuster:
You can do lazy loading by setting up autocmd manually.

If kickstart.nvim wants to be small, it can't be too featureful. What I see in the current state of kickstart.nvim is that it is too bloated, many plugins have too complicated configuration in kickstart. Not to say it unnecessarily overrides built-in features , like LSP mappings.

grous0:
I was using kickstart.nvim as scaffolding but now switched to vim.pack. I agree that kickstarter is too featureful to be a scaffolding.

[...]

PR #1982, opened 2026-04-06
fix: deprecated diagnostic jumping config

nathanzeng:
`opts.jump.float` is deprecated as of the 0.12 release. To confirm, use the kickstart config and run: 
`:checkhealth vim.deprecated`

- ⚠️ WARNING opts.jump.float is deprecated. Feature will be removed in Nvim 0.14
  - ADVICE:
    - use opts.jump.on_jump instead.

This also bumps the minimum version to 0.12

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -184,7 +184,15 @@ vim.diagnostic.config {
   virtual_lines = false, -- Text shows up underneath the line, with virtual lines
 
   -- Auto open the float, so you can easily read the errors when jumping with `[d` and `]d`
-  jump = { float = true },
+  jump = {
+    on_jump = function(_, bufnr)
+      vim.diagnostic.open_float {
+        bufnr = bufnr,
+        scope = 'cursor',
+        focus = false,
+      }
+    end,
+  },
 }
 
 vim.keymap.set('n', '<leader>q', vim.diagnostic.setloclist, { desc = 'Open diagnostic [Q]uickfix list' })
--- a/init.lua
+++ b/init.lua
@@ -192,7 +192,15 @@ do
     virtual_lines = false, -- Text shows up underneath the line, with virtual lines
 
     -- Auto open the float, so you can easily read the errors when jumping with `[d` and `]d`
-    jump = { float = true },
+    jump = {
+      on_jump = function(_, bufnr)
+        vim.diagnostic.open_float {
+          bufnr = bufnr,
+          scope = 'cursor',
+          focus = false,
+        }
+      end,
+    },
   }
 
   vim.keymap.set('n', '<leader>q', vim.diagnostic.setloclist, { desc = 'Open diagnostic [Q]uickfix list' })
--- a/lua/kickstart/health.lua
+++ b/lua/kickstart/health.lua
@@ -12,7 +12,7 @@ local check_version = function()
     return
   end
 
-  if vim.version.ge(vim.version(), '0.11') then
+  if vim.version.ge(vim.version(), '0.12') then
     vim.health.ok(string.format("Neovim version is: '%s'", verstr))
   else
     vim.health.error(string.format("Neovim out of date: '%s'. Upgrade to latest stable or nightly", verstr))


init.lua
```diff
@@ -184,7 +184,12 @@ vim.diagnostic.config {
   virtual_lines = false, -- Text shows up underneath the line, with virtual lines
 
   -- Auto open the float, so you can easily read the errors when jumping with `[d` and `]d`
-  jump = { float = true },
+  jump = {
+    on_jump = function(diagnostic, _)
+      if not diagnostic then return end
+      vim.diagnostic.open_float { focus = false }
+    end,
```
oriori1703: I checked neovim/neovim@dbe17da, which deprecated the option, and it seems like we should do the following if we want to exactly match the previous behaviour:
 
```suggestion
    on_jump = function(_, bufnr)
      vim.diagnostic.open_float {
          bufnr = bufnr,
          scope = 'cursor',
          focus = false,
        }
    end,
```

oriori1703 (changes requested):

oriori1703:
Because this uses a feature (`on_jump`) that was added in 0.12 you should also bump the version in `lua/kickstart/health.lua`.


On a related note, should we wait a bit before dropping support for 0.11?
I know a lot of people get Neovim from their distro’s package manager, and most distros haven’t updated to 0.12 yet.

Both Ubuntu and Fedora are releasing new versions this month, and I assume they’ll include 0.12. Should we wait until those releases?

(Mostly thinking out loud - though I would love to hear others' opinion)

init.lua
nathanzeng: I included `if not diagnostic then return end` because I saw it in the neovim help `:h diagnostic-on-jump-example` or see [here](https://github.com/neovim/neovim/blob/master/runtime/doc/diagnostic.txt#L198)

I didn't add bufnr or scope because I believe the defaults for those values are the same as what we would pass? Let me know if you want me to add it explicitly though and I'll force push

nathanzeng:
Thanks for the review, and good catch on bumping the health check. Waiting for Ubuntu and Fedora seems like a good idea to me.

init.lua
oriori1703: > I included if not diagnostic then return end because I saw it in the neovim help :h diagnostic-on-jump-example or see [here](https://github.com/neovim/neovim/blob/master/runtime/doc/diagnostic.txt#L198)

Because the diagnostic variable is not used in the callback (in contrast to the example), I think it would be better not to check for it.

> I didn't add bufnr or scope because I believe the defaults for those values are the same as what we would pass? Let me know if you want me to add it explicitly though and I'll force push

According to [`:help vim.diagnostic.Opts.Float`](https://neovim.io/doc/user/diagnostic/#vim.diagnostic.Opts.Float) the defaults are:
- `scope` = `line` (instead of `cursor`)
- `focus` = `true` (instead of `false`)
- `bufnr` = current buffer (which would probably be the correct one, but there may be an edge case where it is different)

considering all of the above, I do think my suggestion is the correct behaviour we want, so I would prefer if you make those changes.

nathanzeng: Ah yes that's correct, I made those changes. Reading is a lost art these days 😂

oriori1703:
LGTM :)
Will merge once we drop support for 0.11

oriori1703:
After checking fedora and ubuntu it seems both of them won't include 0.12 in their upcoming versions (at least at release).

https://packages.fedoraproject.org/pkgs/neovim/neovim/ (reason seems to be [this](https://pagure.io/fesco/issue/3588#comment-1010530))

https://packages.ubuntu.com/resolute/neovim (my guess is because [debian-testing](https://tracker.debian.org/pkg/neovim) didn't update either).

Maybe we should reconsider the timeline of dropping support.

nathanzeng:
I'm not sure if you mean dropping support earlier or later, but I think either is fine.

For what it's worth I think the readme for kickstart does a great job of explaining that only the latest version of neovim is supported, and pointing you to how to install it.

oriori1703:
Considering it seems like the big distros won't update to 0.12 anytime soon,  I think it fine to drop support earlier.

Though I think fixing a deprecation warning is not a big enough reason to do that.
A good time will probably be if (or when) we are going to use any of the new features suggested in https://github.com/nvim-lua/kickstart.nvim/issues/1971

agopalareddy:
I reproduced this locally on Neovim 0.12.1 with a fresh Kickstart-based setup.

` :checkhealth vim.deprecated ` reported the `opts.jump.float` deprecation warning for the current config, and switching to the `jump.on_jump` callback form cleared that warning in my verification.

I also ended up needing the explicit float options (`bufnr = bufnr`, `scope = 'cursor'`, `focus = false`) to match the previous behavior, so the current shape of this PR looks right to me from a compatibility/verification standpoint.

The only real open question I see is the repo's 0.11 support timeline, not the implementation itself.

oriori1703 (approved):

oriori1703:
Because #1630 was merged we now require 0.12 so I guess it is a good time to merge this.
Thanks for the help :)

PR #1991, opened 2026-04-13
Refactor/conform

### LINKED ISSUE #1819: please disable autoformat on save in default configuration
I wanted to give a try to neovim in lua repository. I wrote a code. Typed `wa`. Commited with amend. And got 2k changes in pr instead of 50. 

You know, it's quite painful. Please remove autoformat by default.

oriori1703:
1. Change `format_on_save` to a blacklist
2. Use stylua's lsp for formatting instead of its CLI

fixes #1819

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -686,14 +686,15 @@ require('lazy').setup({
     opts = {
       notify_on_error = false,
       format_on_save = function(bufnr)
-        -- Disable "format_on_save lsp_fallback" for languages that don't
-        -- have a well standardized coding style. You can add additional
-        -- languages here or re-enable it for the disabled ones.
-        local disable_filetypes = { c = true, cpp = true }
-        if disable_filetypes[vim.bo[bufnr].filetype] then
-          return nil
-        else
+        -- You can specify filetypes to autoformat on save here:
+        local enabled_filetypes = {
+          -- lua = true,
+          -- python = true,
+        }
+        if enabled_filetypes[vim.bo[bufnr].filetype] then
           return { timeout_ms = 500 }
+        else
+          return nil
         end
       end,
       default_format_opts = {
--- a/init.lua
+++ b/init.lua
@@ -616,6 +616,8 @@ require('lazy').setup({
         -- Special Lua Config, as recommended by neovim help docs
         lua_ls = {
           on_init = function(client)
+            client.server_capabilities.documentFormattingProvider = false -- Disable formatting (formatting is done by stylua)
+
             if client.workspace_folders then
               local path = client.workspace_folders[1].name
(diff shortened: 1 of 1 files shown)


init.lua
```diff
@@ -681,21 +686,23 @@ require('lazy').setup({
...
-          return {
-            timeout_ms = 500,
-            lsp_format = 'fallback',
-          }
+          return nil
         end
       end,
+      default_format_opts = {
+        lsp_format = 'fallback', -- Use LSP formatting if available, otherwise use external formatters. Set to `false` to disable LSP formatting.
+      },
```
nathanzeng, line 702: If the theme of this PR is to reduce unexpected formatting on save, I think these lines should be removed.

nathanzeng: Ah but then the lua formatting will not work. Hmm not sure.

nathanzeng: Anyway, that comment is not correct. Fallback means using LSP when no other formatters are available, not prioritizing it.

See conform documentation [here](https://github.com/stevearc/conform.nvim?tab=readme-ov-file#setupopts)

oriori1703: those lines were added  just so you don't have to repeat it in 2 places (previously it was in 693 and 674)

nathanzeng:
What is the reason for switching to stylua's LSP instead of its CLI? I think it was more clear when stylua was defined in the `formatters_by_ft` block.

SebastianBr11:
I tried using the Stylua LSP for formatting and removing it from `formatters_by_ft` in the conform settings, I also disabled formatting in the `lua_ls` settings, but then pressing `<leader>f` no longer formatted the file. I imagine, Conform tries `lua_ls`, but it doesn't do anything and stylua is ignored

oriori1703:
> I tried using the Stylua LSP for formatting and removing it from `formatters_by_ft` in the conform settings, I also disabled formatting in the `lua_ls` settings, but then pressing `<leader>f` no longer formatted the file. I imagine, Conform tries `lua_ls`, but it doesn't do anything and stylua is ignored

did you also disable the formatting server capability by adding the following line in `on_init`?
Without this I got the same issues:
```lua
           client.server_capabilities.documentFormattingProvider = false -- Disable formatting (formatting is done by stylua)
```

 It seems it is caused by a race condition betweeen stylua and lua_ls. They both run on the same input, and stylua's result arrives earlier, so it was getting overriden by lua_ls.
If you only disable formatting in lua_ls' settings it seems to still "try" to format, but without any changes, so it still causes the issue. My solution was disabling it's ability to format entirely, by removing the capability.
 
Another solution that worked for me was setting `async=false` in `conform.format`, which avoids the race condition by running them synchronously, but also a bit slower.

SebastianBr11:
Oh yeah, I didn't see that, my bad

oriori1703:
> Oh yeah, I didn't see that, my bad

Does it work now?

SebastianBr11:
Yeah, it works now, thanks

PR #1999, opened 2026-04-16
feat: built-in undotree plugin

nathanzeng:
The newly added built-in undotree plugin is pretty awesome. Let's make sure people have it.

This PR require neovim 0.12

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -129,6 +129,9 @@ do
 
   -- Enable undo/redo changes even after closing and reopening a file
   vim.o.undofile = true
+  -- To open a visual and interactive undo tree, type :Undotree
+  -- See `:help :Undotree`
+  vim.schedule(function() vim.cmd.packadd 'nvim.undotree' end)
 
   -- Case-insensitive searching UNLESS \C or one or more capital letters in the search term
   vim.o.ignorecase = true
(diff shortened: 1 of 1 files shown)


feketegy:
I would also add a keymap to open the undotree.

nathanzeng:
I don't think it's necessary. Kickstart is meant to jump start people's configs, not configure everything for them. Users can read the help and add a keymap for this command/action if they wish.

oriori1703:
> I don't think it's necessary. Kickstart is meant to jump start people's configs, not configure everything for them. Users can read the help and add a keymap for this command/action if they wish.

I agree with the argument, but it can also be used to argue that we don't need to enable this at all, so I'm a bit on the fence about this.

I think I'm going to leave this open for a bit and see if anyone has a strong opinion about this before deciding.

nathanzeng:
Sure I can definitely see that argument.

I wanted to include it because
1. Many of the major IDEs do not have this functionality, so I thought it could be cool to highlight this.
2. It's a neovim bundled plugin so it shouldn't add as much weight / dependency to the config

nathanzeng:
but yeah, after rebasing I do question whether it is a foundational setting. If it was added, maybe it goes in a different section?

Christoph-Raab:
> Many of the major IDEs do not have this functionality, so I thought it could be cool to highlight this.

How a bout adding it as commented out code like the repo does in other places? That way everyone (who actually reads the config :D ) is pointed to the function and can try it out/decided on their own, if they want it? 

Imo that would be a compromise and help people learn about neovim, without forcing a feature on them they don't want.

PR #2003, opened 2026-04-19
feat: add LSP document color style configuration

feketegy:
I think this is a cool feature that more people should be aware of when using Neovim.

It adds document color highlights if the LSP supports it and if Neovim is gratear than v0.12.

The highlights will look something like this, depending on the style preset used.

### Foreground

nvim-document-color-1

### Background

nvim-document-color-3

### Literal string

nvim-document-color-2

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -585,6 +585,13 @@ require('lazy').setup({
             })
           end
 
+          -- Set your preferred document color highlight style in the document if the LSP supports it and Neovim is >=0.12
+          --    See `:help lsp-document_color` for more information.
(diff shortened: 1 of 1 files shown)


oriori1703:
it seems to already be enabled by default, unless I'm missing something.
This is how I'm checking:
`lua print(vim.lsp.document_color.is_enabled())`

feketegy:
It is enabled by default, but the style preset is not specified.

oriori1703:
oops, I missed that.

I think it is too niche and the style too subjective for kickstart to pick.
Thanks anyway :)

PR #2012, opened 2026-04-22
Switch from nvim-web-devicons to mini.icons

oriori1703:
As far as I can tell, this functions exactly the same and has no downsides.
The upside is have one less dependency :)

Please let me know if I'm missing something / there is a good reason to stay with `nvim-web-devicons`


roland-5:
Is something blocking this? I don't think i saw any error or issue with this change.

oriori1703:
> Is something blocking this?

nope, I just forgot about it :smile:

Issue #2031, opened 2026-05-04
Current treesitter config causes errors and makes it not working


Jazzcort:
## Before Reporting an Issue
- I have read the kickstart.nvim README.md.
- I have read the appropriate plugin's documentation.
- I have searched that this issue has not been reported before.

- [x] **By checking this, I confirm that the above steps are completed. I understand leaving this unchecked will result in this report being closed immediately.**

## Describe the bug

I was syncing my kickstart fork this today and encountered this error after the merge.
<img width="744" height="162" alt="Image" src="https://github.com/user-attachments/assets/94f7677f-9631-43ba-82e0-2948a3dc9519" />
I went ahead and found `install` is actually `nil` which caused `require('nvim-treesitter').install(parsers)` to raise an error.
I was using nvim v0.11.0 when I first saw this error and then I updated to v0.12.2 and the error still exist, so I guess the error is from the plugin.

## To Reproduce

1. Simply use nvim  with the config of the current main and you will see the error pop up.

## Desktop

- OS: macOS Sequoia 15.7.5
- Terminal: wezterm

## Neovim Version

v0.12.2

oriori1703:
It seems like your are using an old nvim-treesiter version. You should try running `:Lazy update`

Jazzcort:
Thanks for pointing that out! Now it works well!

PR #2032, opened 2026-05-04
fix: change config for treesitter

Jazzcort:
Both `install()` and `get_available()` are nil which causes the error while spawning nvim and make the treesitter not working.

Replacing the custom `FileType` autocmd and manual parser attachment logic with the standard `nvim-treesitter.configs` setup function seems to work fine.

Address https://github.com/nvim-lua/kickstart.nvim/issues/2031


Jazzcort:
Fixed by updating nvim-treesitter version - https://github.com/nvim-lua/kickstart.nvim/issues/2031#issuecomment-4372774573

PR #2043, opened 2026-05-10
feat: markdown opt-in extra + fix local.lua load order

4rc0s:
## Summary

- Adds `lua/whipsmart/plugins/markdown.lua` as a new opt-in extra bundling **render-markdown.nvim** (always loaded) and **obsidian.nvim + blink.compat** (gated on `vim.g.obsidian_vaults` being set in `local.lua`)
- Fixes a load-order bug where `pcall(require, 'local')` ran before `vim.g.have_nerd_font` and all `vim.o.*` defaults, making `local.lua` unable to override them — the call is now at the end of Section 1
- Updates `UNIFIED.md`: markdown extra documented, roci migration and load-order fix checked off

## Enabling the markdown extra on a machine

In `lua/local.lua`:
```lua
vim.g.obsidian_vaults = { { name = 'personal', path = '~/Obsidian/Main' } }
```

In `lua/custom/plugins/markdown.lua`:
```lua
require 'whipsmart.plugins.markdown'
```

render-markdown loads even without `obsidian_vaults` set — only the obsidian/blink.compat block is gated.

## blink.cmp integration notes

`blink.compat` is added without a version pin (HEAD/main branch) — released tags historically lacked `cmp.get_config()` that obsidian.nvim calls internally.

The obsidian sources and `[[` wikilink trigger behaviour are applied via `require('blink.cmp.config').merge_with(...)` — the correct v1 post-setup extension API — called from `custom/plugins/` (Section 3) after `plugins.cmp` has already run `setup()`.

## Test plan

- [ ] Launch nvim with no `obsidian_vaults` set — no errors, render-markdown active in markdown buffers
- [ ] Launch nvim with `obsidian_vaults` set — obsidian completion sources available via `[[`
- [ ] Type a regular word in markdown — completion menu does NOT auto-open
- [ ] Ctrl-Space in markdown — menu opens with full source list
- [ ] Non-markdown buffer — completion behaviour unchanged
- [ ] Machine without Nerd Font sets `vim.g.have_nerd_font = false` in `local.lua` — override takes effect


4rc0s:
Opened against wrong repo by mistake — intended for 4rc0s/whipsmart.nvim

PR #2068, opened 2026-05-28
chore: updated readme

stephansama:
updated the readme to reflect new template direction

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,23 +61,26 @@ Neovim's configurations are located under the following paths, depending on your
 
 #### Recommended Step
 
-[Fork](https://docs.github.com/en/get-started/quickstart/fork-a-repo) this repo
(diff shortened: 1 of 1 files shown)


oriori1703:
This looks great.

I wonder if we should also mention that you could fork / add another remote if you want to be able to update (but say that the template is the recommended way for beginners).

WDYT?

stephansama:
im thinking maybe mentioning that and mentioning the discussion (that way people can understand the implication of both options) makes sense.
im going to update it with that.

oriori1703:
@stephansama are you still going to update this with the discussion?

stephansama:
> @stephansama are you still going to update this with the discussion?

🤦🏽‍♂️ i had it on my local just never pushed lol. sorry i just updated it

oriori1703:
looks great :)

PR #2092, opened 2026-06-11
Merge the 2 `gitsigns.nvim` configs

### LINKED ISSUE #1926: gitsigns recommended keymaps
Because the gitsigns recommended keymaps are not enabled by default, I think there is a case to be made for removing that documentation from which-key entirely, or to add those keymaps in by default.

I would love to hear some discussion around this, and for now I have made a PR  (#1925) that simply adds some clarifying comments on the which-key configuration

oriori1703:
I think it can be confusing for new users to have gitsigns configured in 2 places.

I'm still undecided if the merged config be in `init.lua` or `gitsigns.lua`.
In the latter, gitsigns would be disabled entirely by default, because everything outside of `init.lua` is optional and disabled by default.
I'm leaning towards the former because I think it is useful to for new users to know which lines they changed (especially if they didn't mean to because they are still new to neovim).

Another option is to enable the keymaps by default, which further implies that it should be in `init.lua`.

Anyone has opinions on this?

closes https://github.com/nvim-lua/kickstart.nvim/issues/1926


dam9000:
My vote is to enable the keymaps by default, but that's just me.
If they remain commented out I would suggest to use block instead of line comment, so:
```
--[[
...
--]]
```

nathanzeng:
Note: closes #1926

nathanzeng:
I vote enable by default. The other places in kickstart that there is commented out code, it feels much more optional or debatable. People almost certainly want these keymaps. 

They might want the keys to be different, which is why they can change it.

rez1-dev:
I second to enable by default.

SetsuikiHyoryu:
Enable by default is useful for me.

oriori1703:
seems everyone is in agreement that we should merge this :)

rez1-dev:
@oriori1703 fyi it is still commented out in this commit. Not enabled by default.

dam9000:
> seems everyone is in agreement that we should merge this :)

My reading of everyone's comment was enable by default not leaving it commented out, isn't that so?
There's 9 messages + thumbs up for enable by default.

dam9000:
I made a PR for enabling the gitsigns keymaps by default, please consider merging it:

https://github.com/nvim-lua/kickstart.nvim/pull/2158

oriori1703:
Whoops, I forgot to push the change before merging 😅

furai:
@oriori1703 correct me if I'm wrong - but this commit dropped the `bufnr` and now gitsigns behave differently than previously? Now bufnr seems unused? And shortcuts are loaded globally?

dam9000:
> @oriori1703 correct me if I'm wrong - but this commit dropped the `bufnr` and now gitsigns behave differently than previously? Now bufnr seems unused? And shortcuts are loaded globally?

Looks like you are right. The original code with bufnr was a verbatim copy from the gitsigns.nvim recommended keymaps:
https://github.com/lewis6991/gitsigns.nvim#-keymaps

And I would recommend that we keep it the same, there's no reason to diverge from that.
Although, not sure what negative consequences do the global keymaps have vs per-buffer.

oriori1703:
> @oriori1703 correct me if I'm wrong - but this commit dropped the `bufnr` and now gitsigns behave differently than previously? Now bufnr seems unused? And shortcuts are loaded globally?

I don't think it really matters if they are loaded globally vs per-buffer, but this was definitely unintentional, so I commit a fix.
I guess this is what happens when you merge a branch after 2 month without looking at it again :laughing:

oriori1703:
@dam9000
> Looks like you are right. The original code with bufnr was a verbatim copy from the gitsigns.nvim recommended keymaps: https://github.com/lewis6991/gitsigns.nvim#-keymaps

While verifying the fix I noticed the above was not entirely correct.
I seems that the `<leader>hD` keymap is a bit different than the one in gitsigns.nvim's docs.

This is what is needed to keep it the same:
```diff
diff --git a/init.lua b/init.lua
index 3624001..db1630d 100644
--- a/init.lua
+++ b/init.lua
@@ -391,7 +391,7 @@ do
       vim.keymap.set('n', '<leader>hi', gitsigns.preview_hunk_inline, { desc = 'git preview hunk [i]nline', buf = bufnr })
       vim.keymap.set('n', '<leader>hb', function() gitsigns.blame_line { full = true } end, { desc = 'git [b]lame line', buf = bufnr })
       vim.keymap.set('n', '<leader>hd', gitsigns.diffthis, { desc = 'git [d]iff against index', buf = bufnr })
-      vim.keymap.set('n', '<leader>hD', function() gitsigns.diffthis '@' end, { desc = 'git [D]iff against last commit', buf = bufnr })
+      vim.keymap.set('n', '<leader>hD', function() gitsigns.diffthis '~' end, { desc = 'git [D]iff against last commit', buf = bufnr })
       vim.keymap.set('n', '<leader>hQ', function() gitsigns.setqflist 'all' end, { desc = 'git hunk [Q]uickfix list (all files in repo)', buf = bufnr })
       vim.keymap.set('n', '<leader>hq', gitsigns.setqflist, { desc = 'git hunk [q]uickfix list (all changes in this file)', buf = bufnr })
       -- Toggles
```

The current one (`@`) is equivalent to `git diff HEAD`, which is similar to `gitsigns.diffthis()`, but also show staged changes.
The one from gitsigns.nvim's docs (`~`) is equivalent to `git diff HEAD~`, which show the diff between both the head commit, staged and unsaged changes.

It seems it was this ways since you introduced the gitsigns keymaps in f5c9fe8e15aafb6857706e3c05b5eee4ecb98a2b.
Do you remember if you had a reason, or was it just a mistake?

dam9000:
> It seems it was this ways since you introduced the gitsigns keymaps in [f5c9fe8](https://github.com/nvim-lua/kickstart.nvim/commit/f5c9fe8e15aafb6857706e3c05b5eee4ecb98a2b). Do you remember if you had a reason, or was it just a mistake?

I don't remember, I would suggest to use what gitsigns.nvim recommends.

PR #2093, opened 2026-06-12
chore(init.lua): clarify section comments

rez1-dev:
Now that there is section 2 for keymaps. It doesn't need to be in section 1.

Two more things:

- You have an autocommand under keymaps.
- Keeping comments like `-- [[ Basic Keymaps ]]` feels a bit redundant. I know you are keeping parity with kickstart-modular.nvim

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -86,7 +86,7 @@ P.S. You can delete this when you're done too. It's your config now! :)
 
 -- ============================================================
 -- SECTION 1: OPTIONS
--- Core Neovim settings, leaders, options, basic keymaps, basic autocmds
+-- Core Neovim settings, leaders, options
 -- ============================================================
 do
   -- Enable faster startup by caching compiled Lua modules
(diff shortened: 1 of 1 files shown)


oriori1703:
oops, you are totally right :)

about the other things:
> You have an autocommand under keymaps.

I agree this is not the best, but I couldn't think of a section that fits better.
Any suggestions?

> Keeping comments like -- [[ Basic Keymaps ]] feels a bit redundant.
I still find them useful even if they are repeating the section headers sometimes, because most sections contain multiple "sub sections" - which those comments help separate.

rez1-dev:
Keymaps and autocommands aren't closely related. For a starter guide like Kickstart, users should be able to differentiate between what they trigger and what Neovim triggers automatically.

I prefer `autocmd` to be a new chapter.

```lua
-- SECTION 2: KEYMAPS
-- SECTION 3: AUTOCMDS
```
or you could keep it like before.
```lua
-- SECTION 2: KEYMAPS & AUTOCMDS
```

oriori1703:
For now I will update the section header to your second suggestion, because having a whole section for just one autocmd seems weird to me.

PR #2095, opened 2026-06-16
fix: deprecated `on_yank` command

f1100k:
**vim.hl.on_yank** is deprecated and the feature will be removed in Nvim 0.14.

It's possible to check through the command: `:checkhealth vim.deprecated` 
```
- ⚠️ WARNING vim.hl.on_yank is deprecated. Feature will be removed in Nvim 0.14
  - ADVICE:
    - use vim.hl.hl_op instead.
```

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -245,11 +245,11 @@ do
 
   -- Highlight when yanking (copying) text
   --  Try it with `yap` in normal mode
-  --  See `:help vim.hl.on_yank()`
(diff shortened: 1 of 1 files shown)


rez1-dev:
would be better if this pr gets merged after Nvim 0.13 stable release.

https://neovim.io/doc/user/deprecated/#deprecated-0.13

PR #2132, opened 2026-07-27
feat(readme): clarify _windows_ install recipe

### LINKED ISSUE #2096: `cl.exe` Plugin Dependency On Windows
## Before Reporting an Issue

- I have ran `:checkhealth` and so no obvious issue.

- [x] **By checking this, I confirm that the above steps are completed. I understand leaving this unchecked will result in this report being closed immediately.**

## Describe the bug

`cl.exe` is required for one of plugins. I could see adding the dependencies below to the `README.md`.

#### Error
```shell
[nvim-treesitter/install/query] error: Error during "tree-sitter build": ^[[31mError:^[[0m Failed to compile parser

Caused by:
    Failed to execute the C compiler with the following command:
    "cl.exe" "-nologo" "-MD" "-O2" "-Brepro" "-std:c11" "-I" "C:\\Users\\matt\\AppData\\Local\\Temp\\nvim\\tree-sitter-query\\src" "-W4" "-LD" "
-utf-8" "/Fo\\\\?\\C:\\Users\\matt\\AppData\\Local\\Temp\\nvim\\tree-sitter-query\\tmp_5076_ThreadId(1)\\" "C:\\Users\\matt\\AppData\\Local\\Tem
p\\nvim\\tree-sitter-query\\src\\parser.c" "-link" "-out:\\\\?\\C:\\Users\\matt\\AppData\\Local\\Temp\\nvim\\tree-sitter-query\\parser.so" "/IMP
LIB:\\\\?\\C:\\Users\\matt\\AppData\\Local\\Temp\\nvim\\tree-sitter-query\\tmp_5076_ThreadId(1)\\temp.lib"
    Error: program not found
```

## To Reproduce

1. Be on windows
2. `git clone`
3. Open `nvim`
4. Run through the `vim.pack` install
5. Notice errors with `cl.exe` missing.

### Workaround Using `pwsh.exe`
1. `choco install visualstudio2026buildtools -y`
2. `choco install visualstudio2026-workload-vctools -y`
3. `& 'C:\Program Files (x86)\Microsoft Visual Studio\18\BuildTools\Common7\Tools\Launch-VsDevShell.ps1'`

Some notes on the above. Need to be running with admin to perform the install. I am on _7.6.2_ of _PowerShell_. Item 3 is needed to fix-up the `$env:Path` so processes can find `cl.exe`. On my system, when I run `Get-Command cl` after running the launcher `ps1`, I get `C:\Program Files (x86)\Microsoft Visual Studio\18\BuildTools\VC\Tools\MSVC\14.51.36231\bin\HostX86\x86\cl.exe`. Presumably, one could update the `$env:Path` manually too, though there are four different copies of `cl.exe`.

There is probably a better world where one could use `scoop install vswhere` or `choco install vswhere` to ensure _VS_ 2017 and later can be accounted for and we don't rely on the default location of the binary in the _Visual Studio Installer_ location.

Below is how one can find the instance id to call into the cmdlet directly that is called by the `ps1` above. It does get messy if folks don't have the exact install path as the above.
```pwsh
& 'C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe' -products Microsoft.VisualStudio.Product.BuildTools -latest -format json
```
And then with the whole archival of _nvim-treesitter_. Is it all worth it? Is there going to be a push to go "native" and "core" features with _neovim_? So one would rip out the `nvim-treesitter` dependency?

## Desktop

- OS: Windows
- Terminal: Wezterm

## Neovim Version

```
NVIM v0.12.3
Build type: Release
LuaJIT 2.1.1774638290
```

mattcargile:
Closes #2096 
Rework the ordering and create a single command for all deps. Add additional detail on compiler, minimal installs, etc.

I tried to go minimal and maximal in my approach with the collapsed regions. I wanted to try to appease different audiences. The change could very well be to add two pieces of software ( e.g. `visualstudio2026buildtools visualstudio2026-workload-vctools` to item 2 under the main `choco install -y neovim...` command.

It is unfortunate a little there is a year in the software name and it would grow stale. Maybe there could be a blurb on how to install it? It is a pretty hard dependency for `tree-sitter` on _Windows_ from what I can tell. Other options like `MSYS2` and _Git Bash_ are substantially more complicated and the users in those environments are probably fairly aware of the tool chain as it is. Having the single `choco install` command to get all the software in one shot is important for adoption, I think?

Thanks for letting me contribute. Open to any suggestions. :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,55 @@ After installing all the dependencies continue with the [Install Kickstart](#ins
 
 #### Windows Installation
 
-<details><summary>Windows with Microsoft C++ Build Tools and CMake</summary>
+<details><summary>Windows</summary>
+<details><summary>Chocolatey Install</summary>
+
+Install [chocolatey](https://chocolatey.org/install)
+Either follow the instructions on the page or use `winget`,
+Run in cmd as **admin**:
+```
+winget install --accept-source-agreements chocolatey.chocolatey
+```
+</details>
+
+With _Chocolatey_ already installed, use the below. See additional detail for minimal installs or non standard installs.
+```
+choco install -y neovim git ripgrep wget fd unzip gzip tree-sitter visualstudio2026buildtools visualstudio2026-workload-vctools mingw make
+```
+<details><summary>Windows Detail</summary>
+<details><summary>General Dependencies</summary>
+
+```
+choco install -y neovim git ripgrep wget fd unzip gzip tree-sitter
+```
+</details>
+<details><summary>Windows Microsoft C++ Build Tools</summary>
+
+The tools are required by the `tree-sitter` command line tool which is used by the `nvim-treesitter` plugin. This is the recommended way for a typical Windows install outside of _MSYS2_ , _Git Bash_, etc.
+
+The below installs the base build tools and the [_Microsoft_](https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=visualstudio) recommended _Visual C_ workload and components. Installing it this way is recommended by the _Chocolatey_ package maintainer and allows for easier upgrades, etc. The dependency requires the compiler ( e.g. `cl.exe` ) and header files like `stdlib.h` in the _Windows_ SDK.
+```
+choco install -y visualstudio2026buildtools visualstudio2026-workload-vctools
+```
+<details><summary>Check Previous Install</summary>
+
+If you already are a _Visual Studio_ user with _C_ workloads, you may not require the build tools. You can check your installation with the below to make sure the applicable workload and components are available. It should output something like _Visual Studio Community_ or _Microsoft Visual Studio Build Tools_. If you don't receive any output and you aren't using preview builds, etc, you may need to install the tools.
+```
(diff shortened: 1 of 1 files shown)


oriori1703:
Overall looks good, though we might be able to simplify this a bit.

I'm pretty sure that if gcc is available (through mingw) it can be used by nvim-treesitter instead of msvc.
So maybe we can just suggest users install mingw unconditionally, and drop the vs-build-tools entirely.
WDYT?


BTW, sorry for the delay

nonepork:
> Overall looks good, though we might be able to simplify this a bit.
> 
> I'm pretty sure that if gcc is available (through mingw) it can be used by nvim-treesitter instead of msvc. So maybe we can just suggest users install mingw unconditionally, and drop the vs-build-tools entirely. WDYT?
> 
> BTW, sorry for the delay

Not OP but, the main branch of nvim-treesitter doesn't seem to be able to configure compilers' orders, as it relies on `tree-sitter build` (which uses cc-rs), and the only way to change it is to modify environment variable CC, as far as I'm aware.

mattcargile:
Right. The `gcc` , _mingw_, and _msys2_ scenarios are more complex with environment variables and such; not to mention they have separate shells, etc. With _Visual Studio_ build tools with `cl.exe`, _treesitter_ can find it automatically.

I'll see if I can step through all the scenarios to see what we like the best.

PR #2135, opened 2026-07-29
refactor(conform): notify on error

nathanzeng:
In my opinion, we should notify on error by default. This is also the default configuration for conform.

In my usage with conform, I haven't felt that the error noise is an issue. Sometimes it's nice for me when I'm blazing through stuff, I miss the LSP diagnostics, but then when I write I get an error from conform formatter because my syntax tree is broken.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -784,7 +784,6 @@ do
   -- [[ Formatting ]]
   vim.pack.add { gh 'stevearc/conform.nvim' }
   require('conform').setup {
-    notify_on_error = false,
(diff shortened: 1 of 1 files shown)


oriori1703:
I think it is kind of an hack to use the formatter output to know if you have a syntax issue. This is what your lsp and language tools are for.

Also most of the time I format out of muscle memory, even if I know there are syntax errors, so I find it kind of annoying.

Closing this for now, but if you disagree feel free to open a discussion. If other agree, I will reconsider :)

PR #2151, opened 2026-08-25
docs: clarify custom plugin load ordering

### LINKED ISSUE #2038: Document explicit custom plugin requires when order matters
the new `vim.pack` config has the convenience loader:

```lua
-- require 'custom.plugins'
```

I think there should also be something about explicit ordering for custom plugin modules where setup order matters:

```lua
require 'custom.plugins.colorscheme'
require 'custom.plugins.ui'
require 'custom.plugins.git'
```

## Current code

`lua/custom/plugins/init.lua` currently auto-loads every Lua file in `lua/custom/plugins/`:

```lua
local plugins_dir = vim.fs.joinpath(vim.fn.stdpath 'config', 'lua', 'custom', 'plugins')
for file_name, type in vim.fs.dir(plugins_dir) do
  if type == 'file' and file_name:match '%.lua$' and file_name ~= 'init.lua' then
    local module = file_name:gsub('%.lua$', '')
    require('custom.plugins.' .. module)
  end
end
```

for independent modules - great. however, since `vim.pack` plugin files are imperative Lua and may call `vim.pack.add`, define commands/keymaps, or immediately run `setup()`, load order can matter.

## Suggested change

Update the `init.lua` to add short blurb about load ordering in vim.pack world:

```lua
  -- NOTE: You can add your own plugins, configuration, etc from `lua/custom/plugins/*.lua`
  --
  -- For independent modules, uncomment this convenience loader:
  -- require 'custom.plugins'
  --
  -- If you have a custom module that depends on another plugin or module being set up first,
  -- require modules explicitly in the order they should run:
  --
  -- require 'custom.plugins.colorscheme'
  -- require 'custom.plugins.ui'
  -- require 'custom.plugins.git'
```

TyceHerrman:
## Summary

- clarify that the custom-plugin convenience loader's file order is unspecified
- recommend keeping dependent plugins in the same file
- document explicit ordered `require`s for intentionally separate modules
- place the same warning beside the loader implementation

The loader itself remains unchanged.

Closes #2038

## Testing

- `stylua --check .`
- `git diff --check`

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -976,10 +976,19 @@ do
   -- require 'kickstart.plugins.neo-tree'
   -- require 'kickstart.plugins.gitsigns' -- adds gitsigns recommended keymaps
 
-  -- NOTE: You can add your own plugins, configuration, etc from `lua/custom/plugins/*.lua`
+  -- NOTE: You can add your own plugins, configuration, etc. in `lua/custom/plugins/*.lua`.
   --
-  --  Uncomment the following line and add your plugins to `lua/custom/plugins/*.lua` to get going.
+  -- For independent modules, uncomment the convenience loader:
   -- require 'custom.plugins'
+  --
+  -- `custom.plugins` automatically loads files from that directory, but their
+  -- order is unspecified. If plugins depend on each other, keep them in the same
+  -- file and put their `vim.pack.add()` and `setup()` calls in the required order.
+  --
+  -- If separate modules need a specific order, require them explicitly instead:
+  -- require 'custom.plugins.colorscheme'
+  -- require 'custom.plugins.ui'
+  -- require 'custom.plugins.git'
 end
 
 -- The line beneath this is called `modeline`. See `:help modeline`
--- a/lua/custom/plugins/init.lua
+++ b/lua/custom/plugins/init.lua
@@ -3,7 +3,8 @@
 --
 -- See the kickstart.nvim README for more information
 
--- Iterate over all Lua files in the plugins directory and load them
+-- Iterate over all Lua files in the plugins directory and load them.
+-- `vim.fs.dir()` iteration order is unspecified and must not be relied upon.
 local plugins_dir = vim.fs.joinpath(vim.fn.stdpath 'config', 'lua', 'custom', 'plugins')
 for file_name, type in vim.fs.dir(plugins_dir, { follow = true }) do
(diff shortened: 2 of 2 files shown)


init.lua
```diff
@@ -976,10 +976,19 @@ do
...
   -- require 'kickstart.plugins.gitsigns' -- adds gitsigns recommended keymaps
 
-  -- NOTE: You can add your own plugins, configuration, etc from `lua/custom/plugins/*.lua`
+  -- NOTE: You can add your own plugins, configuration, etc. in `lua/custom/plugins/*.lua`.
   --
-  --  Uncomment the following line and add your plugins to `lua/custom/plugins/*.lua` to get going.
+  -- For independent modules, uncomment the convenience loader:
   -- require 'custom.plugins'
+  --
+  -- `custom.plugins` automatically loads files from that directory, but their
```
Christoph-Raab, line 984: Maybe out of scope for the PR, but why doesn't the repo recommand using the official `after/plugin` way? That would be more nvim native and would save a bit of complexity with the custom init.lua in `lua/custom/plugins`.

TyceHerrman: that could be a better approach, but it's a bigger change. files under after/plugin/ run in a later startup phase. During init.lua, vim.pack.add() defaults to load = false, and after init.lua it defaults to load = true. Moving the current vim.pack.add() and setup() calls into after/plugin/ would change when plugin scripts are loaded. This PR documents how the existing loader works, but Kickstart adopting after/plugin/ as its recommended extension layout would supersede it. so could be closed if that's the direction to go now, but if that needs further deliberation, could merge this PR as is imho

oriori1703: > Maybe out of scope for the PR, but why doesn't the repo recommand using the official after/plugin way?

The main reason I didn't do it when I migrated to `vim.pack` was because I wanted to keep the migration easier for existing users.
Could you open a disscution about this topic?

> files under after/plugin/ run in a later startup phase. During init.lua, vim.pack.add() defaults to load = false, and after init.lua it defaults to load = true. Moving the current vim.pack.add() and setup() calls into after/plugin/ would change when plugin scripts are loaded.

If we want to keep parity, instead of `/after/plugin/` we could also use `/plugin/`, which keeps the same vim.pack.add behavior.

oriori1703 (approved):
Looks great :)

PR #2158, opened 2026-09-05
Enable gitsigns.nvim recommended keymaps by default

dam9000:
Please consider merging this.
I see no drawback in having the gitsigns.nvim recommended keymaps be enabled by default
The initial reason these were outside of init.lua was so that the line count of init.lua was kept relatively low and the reason it was disabled by default was the wish that an init.lua without any extra files (plugins) would work just as well - in case someone just copied the init.lua file and not the rest of the accompanying plugins.
Now with the keymaps integrated into init.lua, there is no need to have them commented out, there is nothing left to protect against, it just adds friction to what is a sensible default.

related: https://github.com/nvim-lua/kickstart.nvim/pull/2092

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -360,47 +360,46 @@ do
       topdelete = { text = '‾' }, ---@diagnostic disable-line: missing-fields
       changedelete = { text = '~' }, ---@diagnostic disable-line: missing-fields
     },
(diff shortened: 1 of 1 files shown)


dam9000:
@oriori1703 Thank you! :)

PR #2161, opened 2026-09-07
gitsigns: use `buf` keymap option on Neovim 0.12.1+, fallback to buffer on older versions

dam9000:
The recent move from keymap.set buffer to buf broke neovim on ubuntu 25.10 with:
```
vim/keymap.lua:0: invalid key: buf
```
nvim on ubuntu 25.10 is: v0.12.0-dev
So recent enough but not quite 0.12.0 release - it's an earlier pre-release snapshot, so it breaks.
I think we could make an effort to keep the support of such a version, hence this PR.
There are 2 commits:
```
    gitsigns re-introduce back the map helper so that buf/buffer can be runtime selected
```
and:
```
    gitsigns: use `buf` keymap option on Neovim 0.12.1+
    
    Follows the upstream rename of `buffer` to `buf` (neovim/neovim#38360,
    first shipped in v0.12.0); `buffer` is soft-deprecated, removal
    planned in v0.15. Fall back to `buffer` on older Neovim.
    
    The 0.12.1 floor (not 0.12.0) avoids distro snapshot builds like
    Ubuntu 25.10 `0.12.0~ubuntu1` (git snapshot d62bbe24cb), which reports
    `0.12.0-dev` but predates the `buf` support and rejects it with
    `invalid key: buf`.
```

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -364,7 +364,12 @@ do
     on_attach = function(bufnr)
       local function map(mode, l, r, opts)
         opts = opts or {}
-        opts.buffer = bufnr
+        -- opts.buf requires Neovim 0.12.1+; older versions use opts.buffer
+        if vim.version.ge and vim.version.ge(vim.version(), '0.12.1') then
+          opts.buf = bufnr
+        else
+          opts.buffer = bufnr
+        end
         vim.keymap.set(mode, l, r, opts)
       end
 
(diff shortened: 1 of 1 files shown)


oriori1703:
Sorry, but considering that we say that we only the support the **latest** stable version, and the fact that ubuntu 25.10 is already EOL since July 9th (because interim release are only supported for 9 month), I think it's best to just use `buf` and avoid this complexity.

I'm closing the PR for now, but if you strongly disagree I can reevaluate.

